PR #30, opened 2025-08-26
Feat/local network support

Nathan1258:
## Summary

Adds a toggle in the server configuration to expose the Osaurus server to the local network (0.0.0.0). This allows other devices and services (like Open WebUI) on the same network to connect to it, mirroring functionality found in tools like Ollama. When disabled, the server remains safely bound to localhost. As requested in this issue: https://github.com/dinoki-ai/osaurus/issues/25

## Changes

- [x] Behavior change
- [ ] Refactor / chore
- [ ] Tests
- [ ] Docs

## Test Plan

1. Launch the app with the server stopped
2. Click the Gear Icon to open the Server Configuration panel
3. Enable the "Expose to the network" toggle and click save
4. Click the Start button. The UI should now display your computer's local network IP address instead of the loopback address
5. From another device on the same network, open a terminal and run `curl http://<YOUR_LOCAL_IP>:<YOUR_PORT>/health`.
6. Verify that you receive a successful JSON response 
7. Stop the server and go back into the configuration and disable the "Expose to the network" toggle. Click save.
8. Verify the UI now shows the loopback address `curl http://127.0.0.1:8080`.
9. Verify that the curl command from your other device now fails (times out etc.,)

## Screenshots

CleanShot 2025-08-26 at 17 41 08

## Checklist


osaurus/Views/ContentView.swift
```diff
@@ -8,6 +8,7 @@
 import SwiftUI
 import AppKit
 import Sparkle
+import Darwin
```
tpae: Do we need this here?

osaurus.xcodeproj/project.pbxproj
```diff
@@ -452,8 +452,19 @@
 				COMBINE_HIDPI_IMAGES = YES;
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_TEAM = 4W8QF9VR2F;
+				ENABLE_APP_SANDBOX = YES;
```
tpae: let's disable this

osaurus/Views/ContentView.swift
Nathan1258: no! it was left over after I decided it was best to move the IPv4 logic into the Server Configuration - it's removed

tpae (approved):
❤️

osaurus.xcodeproj/project.pbxproj
Nathan1258: removed

PR #286, opened 2025-12-26
Fix sidebar missing glass blur effect

hkay-dev:
Note: This fix was developed with assistance from Claude Code. I've reviewed the changes and kept them minimal.

---

The chat history sidebar had semi-transparency but no blur effect. It was using a solid color with opacity instead of the glass material.

1-before-sidebar-fix
2-after-sidebar-fix

Fix: Use `ThemedGlassSurface` for the sidebar background when `theme.glassEnabled` is true, with `cornerRadius: 0` since it sits inside the window frame.


auroraflux:
Closing in favor of #287 which includes this fix along with additional theme color handling improvements.

Issue #310, opened 2026-01-01
LFM2-2.6B-Exp tool-calling output doesn’t plug cleanly into Osaurus tool system


EphraimElgrabli:
### Problem statement

I’m trying to use LFM2-2.6B-Exp (mlx-community) inside Osaurus, and the model itself is honestly wild. For a 2.6B model, the instruction following and agent-style reasoning are on another level. It genuinely feels like one of the best small “agent brains” out there, and the IFBench results vs models hundreds of times larger totally match real usage.
The problem I’m hitting is purely around tool calling. LFM2 emits tool calls as custom textual tokens inside the assistant output (e.g. <|tool_call_start|> … <|tool_call_end|>). Because Osaurus exposes tools via OpenAI-compatible APIs and MCP, these tool calls aren’t automatically picked up as executable actions, which creates friction when running the model inside the Osaurus ecosystem.
This isn’t a complaint about the model at all — it’s excellent. I’m opening this because LFM2-2.6B-Exp feels like a perfect fit for Osaurus (local, fast, agentic), and this output mismatch is currently the only thing stopping it from working smoothly with Osaurus’ tool + MCP system.

tpae:
Thanks for the suggestion, I will take a look at this

tpae:
Hey @EphraimElgrabli , i was able to look deeper into this issue. it seems like LFM2 tool calling isn't fully supported yet by mlx-swift-lm library. i've filed an issue here: https://github.com/ml-explore/mlx-swift-lm/issues/47

tpae:
Hey @EphraimElgrabli , i pointed the branch to `main` of my fork. it should contain the fix for LFM2 tool calling. Let me know if you can test it out. Thanks again for being a user!

PR #403, opened 2026-02-01
fixed lfm2 models not calling tools properly

tpae:
## Summary

You can read more details here: https://github.com/ml-explore/mlx-swift-lm/pull/91

## Checklist


tpae:
This should also fix #310

Issue #632, opened 2026-03-14
Fatal crash when a high res image is selected in MLX vision models


RaajeevChandran:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest `main` or release

### Describe the bug

When a high-resolution image is selected in MLX vision models, the app crashes due to exceeding the maximum allowed buffer size

### Steps to reproduce

1. Select an MLX vision model
2. Select a high resolution image
3. Tap send

### Osaurus version / commit

v0.14.2

### macOS version

26.2

### Apple Silicon chip

M4 Pro 24 GB

### Xcode version

26.3 (17C519)

### Logs

```shell
MLX/ErrorHandler.swift:343: Fatal error: [metal:: malloc]
Attempting to allocate 73359820800 bytes which is greater than the maximum allowed buffer size of 14302248960 bytes.
```

### Screenshots

<img width="1896" height="720" alt="Image" src="https://github.com/user-attachments/assets/ea5e040d-14e5-485e-86ab-e3aa20379867" />

PR #636, opened 2026-03-14
added image resize config for MLX vision input

RaajeevChandran:
## Summary

enforce a fixed image resize config for MLX vision models to avoid fatal crash due to maximum allowed buffer size. fixes [#634](https://github.com/osaurus-ai/osaurus/issues/632)

## Test Plan
1. Select an MLX vision model
2. Select a high resolution image and tap send
3. The app must run without crashing

## Checklist


tpae:
Looks great! Thank you for your contribution!

PR #641, opened 2026-03-15
fix: tapping active mode segment incorrectly switches to opposite mode

RaajeevChandran:
## Summary

The Chat/Work mode toggle was firing a single toggle action on any tap, which always switched to the opposite mode regardless of which segment was tapped. Each segment now reports its own `Mode` back through the action closure and the caller guards against switching when the tapped mode is already active

## Changes

- [x] Behavior change
- [x] Refactor / chore
- [x] Tests
- [x] Docs

## Test Plan
1. Tap the **Chat** segment (already selected) - confirm nothing happens and the mode stays as Chat
2. Tap the **Work** segment - confirm the mode switches to Work
3. Tap the **Work** segment again (already selected) - confirm nothing happens and the mode stays as Work
4. Tap the **Chat** segment - confirm the mode switches back to Chat

## Screenshots

Before

https://github.com/user-attachments/assets/9622879e-ffaf-409f-ba44-2e0f2514dee0

After

https://github.com/user-attachments/assets/c2aa64c1-8393-4c48-99c9-0506df522375

## Checklist

### DIFF
--- a/Packages/OsaurusCore/Views/Components/SharedHeaderComponents.swift
+++ b/Packages/OsaurusCore/Views/Components/SharedHeaderComponents.swift
@@ -54,16 +54,12 @@ struct ModeToggleButton: View {
     @Namespace private var animation
 
     var body: some View {
-        Button(action: {}) {
-            HStack(spacing: 0) {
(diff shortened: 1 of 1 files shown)


tpae:
This is awesome! Thank you for your contribution!

PR #643, opened 2026-03-15
UI/UX enhancements

RaajeevChandran:
## Summary

This PR addresses a few UI/UX enhancements.

# a. Added an "Add Model" button in ModelPickerView
When I first started using Osaurus, I had a difficult time figuring out how I could download new models besides the ones that I downloaded during onboarding. It took me a few minutes to understand that the Settings button at the top right is visible only when starting a new chat and disappears after the first prompt in the conversation. Since new models can only be downloaded from the settings screen, I think its reasonable to add another way to get to the settings window without having to create a new conversation. So I added an "Add Model" button which takes us directly to the Models section of the settings window.

image

https://github.com/user-attachments/assets/d0a25b91-5cb0-4a14-abc2-156b69bfc498

# b. Show scroll bar in chat input text view only when necessary
This is a minor UI issue that bugged me every time I opened the app

**Before**
Pasted Graphic 1

**After**
> hides the scroll bar when its not necessary
image

> shows scroll bar based on the content instead of showing it all the time

image

# c. Center `ChatToolbarModeToggleView` based on macOS HIG
Currently the app has a "left-heavy" layout where the toolbar toggle is too close to the sidebar button and traffic lights. This change adheres to macOS HIG by establishing the primary state switching control as a symmetrical visual anchor and clearly distinguishing global navigation from localized utilities.

**Before**
image

**After**
image

# d. Created a filter menu in `ModelDownloadView` for better model browsing
The new filter menu will let users filter the model list by several parameters like model type, model download size, model parameter size and model family.

https://github.com/user-attachments/assets/3b7d29ba-eda7-4416-96d4-f97f52e7bddb

# d. Improved scroll stagger animation
The model cell's animation is currently out of sync with the scroll speed. When scrolling faster, the cells in the list seem to suffer a lot of delays before they appear. So I've implemented a better stagger animation strategy (max 0.04s) and a snappier spring curve that maintains a float up effect while still ensuring that the content appears almost instantly as it enters the viewport.

**Before**

https://github.com/user-attachments/assets/dec2d4e9-46b4-4c8d-a364-2fea58ca2068

**After**

https://github.com/user-attachments/assets/2001051b-2eac-4584-83af-7d51b0327933

# e. Constrain images to bounding box in chat list view
Currently the images are rendering at their native resolution which makes them appear massive in the chat list view. I've constrained the images to a fixed 280x200 bounding box to make them look compact.

**Before**
Pasted Graphic 5

**After**
Pasted Graphic 6

# f. Right-align user messages
This is a continuation of previous enhancement. We now have unnecessary trailing space inside the user message bubble which we should get rid of. 

**Before**
Pasted Graphic 7

So I've made the contents of the user message bubble to align towards the right and wrap only the text content inside the bubble instead of all the views. The user bubble now dynamically grows based on the prompt text content instead of being a fixed width container.

**After**
image

# g. Consistent padding for bottom bar buttons in chat input view
Addresses [#644](https://github.com/osaurus-ai/osaurus/issues/644)

**Before**
image

**After**
image

## Changes
- [ ] Behavior change
- [ ] Refactor / chore
- [ ] Tests
- [ ] Docs

## Checklist

### DIFF
--- a/Packages/OsaurusCore/Views/Components/FloatingInputCard.swift
+++ b/Packages/OsaurusCore/Views/Components/FloatingInputCard.swift
@@ -1114,7 +1114,7 @@ struct FloatingInputCard: View {
             // Bottom button bar with all action buttons
             buttonBar
                 .padding(.horizontal, 12)
-                .padding(.vertical, 6)
+                .padding(.vertical, 12)
         }
         .fixedSize(horizontal: false, vertical: true)
         .background(cardBackground)


tpae (approved):
LGTM!

Issue #689, opened 2026-03-20
Transcription mode is very unreliable


thirdnull:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest `main` or release

## Problem

Transcription mode is currently unreliable and difficult to use. The root cause appears to be the "type-as-you-speak" implementation. Streaming transcription output into the active text field in real time is an extremely hard problem to get right.

## Suggested Fix

Replace the live-typing approach with a simpler push-to-talk model:

1. User presses a hotkey to start recording
2. User presses it again to stop
3. The transcribed text is copied to the clipboard and pasted into the active field (restoring the previous clipboard contents afterward)

## Prior Art / Reference

SuperWhisper is a macOS app whose *entire product* is voice transcription, and it uses this exact pattern while deliberately avoiding live typing. Notably, it ships with the same Parakeet model Osaurus uses. If a dedicated transcription app considers live typing out of scope, it's a strong signal that the complexity isn't worth it for a secondary feature here.

## Why This Matters

The clipboard approach is simpler to implement, significantly more reliable, and is already a proven UX pattern. The current implementation risks making transcription mode feel broken, which reflects poorly on the product overall. Scoping down to the clipboard model would ship something that actually works.

### Provider type

Local model (MLX)

### Osaurus version / commit

Latest

### macOS version

26.1

### Logs

```shell

```

PR #699, opened 2026-03-22
batch process voice input instead of real time

RaajeevChandran:
## Summary

Addresses #689.

- Voice input will be transcribed incrementally and shown after the user has stopped talking instead of showing real-time transcription which is currently unreliable. 

- This behavior can be turned on/off from the Voice section of the settings page using this toggle.
image

- Fixed a bug preventing the voice input button from appearing initially in the chat input

- Previously, the voice input button was hidden after sending every prompt. From now on, the voice input button will be disabled when the response is streaming instead of hiding it for better UX. 

## Changes

- [x] Behavior change
- [ ] Refactor / chore
- [ ] Tests
- [ ] Docs

## Checklist


tpae (approved):
LGTM!

PR #712, opened 2026-03-24
Fix ChatEngine remote services isolation, bookmark fallback, and config trimming

vernonstinebaker:
## Summary

Four independent fixes discovered during local MLX inference work:

1. **ChatEngine**: remote services were fetched via `MainActor.run` inside the actor on every request — now snapshotted at construction time, eliminating an unnecessary hop per request.
2. **DirectoryPickerService**: security-scoped bookmarks become stale on non-sandboxed builds after a volume remount; adds a fallback resolution without the scope flag.
3. **ConfigurationView**: `genMaxKVSize` / `genPrefillStepSize` text fields could produce `0` on empty input; now trims whitespace and returns `nil` for empty strings.
4. **DebugLog**: moves `debugLog()` utility to `Utils/DebugLog.swift` (per architecture guidelines) — previously it was defined inline in `ChatView.swift`.

Independent of the other MLX PRs.

## Changes

- [x] Behavior change
- [x] Refactor / chore
- [ ] Tests
- [ ] Docs

## Test Plan

1. Build and launch the app
2. In Settings → Model, clear the "Max KV Size" field and save — verify no `0` is stored
3. Connect a remote provider; send a chat request — verify routing still works
4. (Non-sandboxed only) Unmount/remount the volume containing a watched directory — verify the directory picker still resolves

## Screenshots

N/A — no visible UI change.

## Checklist

### DIFF
--- a/Packages/OsaurusCore/Services/Chat/ChatEngine.swift
+++ b/Packages/OsaurusCore/Services/Chat/ChatEngine.swift
@@ -9,6 +9,7 @@ import Foundation
 
 actor ChatEngine: Sendable, ChatEngineProtocol {
     private let services: [ModelService]
+    private let remoteServicesSnapshot: [ModelService]
     private let installedModelsProvider: @Sendable () -> [String]
 
     /// Source of the inference (for logging purposes)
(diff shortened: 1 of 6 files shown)


tpae (approved):
LGTM!

PR #735, opened 2026-03-27
feat: add Disable Tools toggle to chat configuration

vernonstinebaker:
## Summary

Adds a **Disable Tools** toggle to the Chat section of Settings. When enabled, no tool specs, preflight RAG results, or capability context snippets are injected into the request — the raw message goes directly to the model.

**Why:** When osaurus is acting as a plain LLM backend for an external agent (e.g. Claude via the OpenAI-compatible API), the caller manages its own tool routing. Injecting osaurus's tool list into every request:
- Changes the system-prompt token sequence, preventing KV-cache reuse across turns.
- Adds tokens that the external caller does not want.

With Disable Tools on, the prompt is stable across all turns in a session, maximising KV-cache hit rate.

Also fixes the `maxToolAttempts` clamp in `ConfigurationView` from `min(10, v)` to `min(50, v)`, matching the UI's 1–50 allowed range and the default value of 15.

## Changes

- [x] Behavior change

## Test Plan

1. Open Settings → Chat.
2. Enable "Disable tools".
3. Start a conversation with a local model.
4. Verify no tool specs appear in the request payload (check app logs or use the Server Explorer to inspect the enriched inference object).
5. Disable the toggle; confirm tools re-appear on the next message.

```
swift test --package-path Packages/OsaurusCore
```

## Screenshots

_UI change: new "Tools" subsection below "Capability Search" in the Chat settings panel with a "Disable tools" toggle and descriptive help text. The Capability Search mode picker is greyed out when Disable Tools is on._

## Checklist

### DIFF
--- a/Packages/OsaurusCore/Models/Chat/ChatConfiguration.swift
+++ b/Packages/OsaurusCore/Models/Chat/ChatConfiguration.swift
@@ -57,6 +57,12 @@ public struct ChatConfiguration: Codable, Equatable, Sendable {
     /// Controls how aggressively pre-flight capability search loads context (nil defaults to .balanced)
     public var preflightSearchMode: PreflightSearchMode?
 
+    // MARK: - Tool Settings
+    /// When true, no tools or preflight context are passed to the model. The raw message is sent
+    /// directly, keeping the prompt stable across turns for maximum KV-cache reuse. Recommended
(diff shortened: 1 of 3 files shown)


tpae:
Built in tools
Sandbox tools
Work mode tools
Folder context tools
Plugin tools

tpae (approved):

PR #736, opened 2026-03-27
perf: tiered KV cache store, O(1) streaming decode, eval fix

vernonstinebaker:
## Summary

Three coordinated improvements to inference latency and correctness:

**1. Tiered KV cache store (RAM + SSD)**

`KVCacheStore` gains a two-tier architecture. Hot sessions are kept in RAM up to a model-size-derived budget (≈ ¼ of model weights, capped at 8 GB). Sessions evicted from RAM are serialised to SSD in a background task and rehydrated on the next turn. An O(1) doubly-linked LRU list drives eviction so no O(n) scans occur on every put. `os_signpost` intervals bracket SSD load/save for easy Instruments profiling.

**2. O(1) streaming decode in StreamAccumulator**

The previous decode path re-decoded the entire token history on every new token, giving O(n²) total work for a response of n tokens. The new path decodes only the net-new suffix per step (O(n) total). `os_signpost` intervals wrap each `accumulate()` call.

**3. MLXArray materialisation fix (correctness)**

`TokenIterator` uses `asyncEval()` to keep the GPU pipeline full, leaving graph nodes inside the KV cache arrays. When the cache is stored and later read for the next turn, MLX re-walks those graphs from scratch — effectively triggering a full re-prefill even though `offset` and token counts are correct.

Fix: call `eval()` on all cache arrays immediately after prefill in both `storeSessionCache` and `buildPrefixCache`. Subsequent reads are then O(1) buffer accesses.

**4. Pre-generation snapshot (correctness)**

The previous approach stored the cache post-generation, keyed by `promptTokens + generatedIds`. On the next turn, the common prefix diverges by exactly `len(generatedIds)` tokens, requiring trim. `MambaCache` is not trimmable, so this caused a full re-prefill on every turn for hybrid SSM models.

Fix: snapshot the cache immediately after prefill (before any tokens are generated) and store it keyed by `promptTokens` only. The post-generation `wrapWithCacheStore` wrapper is removed.

Also adds `KVCacheStore.deepCopyCache` (used by the pre-generation snapshot and by PR #perf/two-phase-prefill) with test coverage for all three cache types (`KVCacheSimple`, `RotatingKVCache`, `MambaCache`), including a fix for a latent crash when deep-copying an empty `RotatingKVCache`.

## Changes

- [x] Behavior change
- [x] Refactor / chore
- [x] Tests

## Test Plan

```
swift test --package-path Packages/OsaurusCore
```

New tests: `KVCacheStoreTests` — `deepCopyCache_kvCacheSimple`, `deepCopyCache_rotatingKVCache`, `deepCopyCache_mambaCache`, `deepCopyCache_mixedTypes`, `saveToDisk_noopForMissingSession`.

Integration: run a 3-turn conversation with any local model and observe:
- T2/T3 TTFT is shorter than T1 (cache reused).
- SSD eviction log lines appear after the RAM budget is exceeded.
- Instruments → os_signpost shows `ssdLoad`/`ssdSave`/`accumulate` intervals.

Benchmark (M4 Pro 64GB, non-hybrid models, 3 turns):
- Phi-4-mini-reasoning T2 TTFT: 513 ms
- GLM-4.7-Flash T2 TTFT: 843 ms

## Checklist

### DIFF
--- a/Packages/OsaurusCore/Services/ModelRuntime.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime.swift
@@ -228,6 +228,23 @@ actor ModelRuntime {
     }
 
     private func storeSessionCache(sessionId: String, cache: [any KVCache], promptTokens: [Int], modelName: String) {
+        // Materialize all lazy MLXArrays before storing.
+        //
+        // During generation, TokenIterator uses asyncEval() to keep the GPU pipeline full.
+        // Each in-place key/value write (`keys[..., prev..<offset, ...] = newKeys`) adds a
+        // graph node.  After generateLoopTask completes and Stream().synchronize() returns,
+        // the GPU is finished — but the MLXArray objects in the cache may still hold
+        // unevaluated computation-graph references rather than concrete GPU buffers.  On the
+        // next turn, MLX walks those graphs from scratch, triggering a full re-prefill even
+        // though offset and token counts are correct.
+        //
+        // eval() forces the arrays to concrete buffers right now, so subsequent reads are
+        // O(1) buffer accesses instead of O(n) graph replays.
+        let arraysToEval = cache.flatMap { $0.state }
+        if !arraysToEval.isEmpty {
+            eval(arraysToEval)
+        }
+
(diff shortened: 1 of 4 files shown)


tpae (approved):
LGTM

PR #748, opened 2026-03-29
batch embedding for context index rebuilds

RaajeevChandran:
## Summary

rebuild and tool registry sync no longer call addDocument once per row. they use addDocuments so the embedder runs in fewer, larger batches. startup ordering for DB open and Vectura init is unchanged.

## Changes

- `SkillSearchService`, `MethodSearchService`, `ToolSearchService` rebuildIndex collects texts/IDs and calls `addDocuments(texts:ids:)` instead of looping addDocument
- `ToolIndexService.syncFromRegistry` upserts SQLite (and prunes stale rows) then calls `ToolSearchService.rebuildIndex() `once instead of per tool `indexEntry`
- `AppDelegate` comment clarifies ordered init vs batched rebuild embedding

- [x] Behavior change
- [ ] Refactor / chore
- [ ] Tests
- [ ] Docs

## Checklist

### DIFF
--- a/Packages/OsaurusCore/AppDelegate.swift
+++ b/Packages/OsaurusCore/AppDelegate.swift
@@ -159,7 +159,9 @@ public final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelega
             }
         }
 
-        // Initialize context management (sequential to avoid concurrent CoreML model loads).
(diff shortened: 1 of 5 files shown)


tpae (approved):
LGTM!

PR #767, opened 2026-04-02
fix: bonjour agent calls now run tools on the server

ritave:
## Summary

Previously the Bonjour agents would run tools on the client instead of server. This fixes that.

## Changes

- [x] Behavior change
- [ ] Refactor / chore
- [ ] Tests
- [ ] Docs

## Test Plan

Set up bonjour agent, connect to it, try to run one of the tools

## Screenshots

image

## Checklist

### DIFF
--- a/Packages/OsaurusCore/Networking/HTTPHandler.swift
+++ b/Packages/OsaurusCore/Networking/HTTPHandler.swift
@@ -1587,7 +1587,8 @@ final class HTTPHandler: ChannelInboundHandler, Sendable {
 
         // Extract agent ID: /agents/{id}
         let components = path.split(separator: "/")
-        guard components.count == 2, components[0] == "agents", let agentId = UUID(uuidString: String(components[1])) else {
+        guard components.count == 2, components[0] == "agents", let agentId = UUID(uuidString: String(components[1]))
+        else {
             hop {
                 var headers = [("Content-Type", "application/json; charset=utf-8")]
                 headers.append(contentsOf: cors)


Packages/OsaurusCore/Services/Provider/RemoteProviderService.swift
```diff
@@ -612,6 +612,15 @@ public actor RemoteProviderService: ToolCapableService {
         toolChoice: ToolChoiceOption?,
         requestedModel: String?
     ) async throws -> String {
+        // Native Osaurus agents run tools server-side — route through generateOneShot
+        if provider.providerType == .osaurus {
```
tpae, line 633: we should create `OsaurusProviderService.swift`, convention over configuration

Packages/OsaurusCore/Views/Chat/ChatView.swift
```diff
@@ -1010,8 +1010,34 @@ final class ChatSession: ObservableObject {
...
+                                toolTurn.toolCallId = done.callId
+                                let newAssistantTurn = ChatTurn(role: .assistant, content: "")
+                                turns.append(contentsOf: [toolTurn, newAssistantTurn])
+                                assistantTurn = newAssistantTurn
+                                processor = StreamingDeltaProcessor(
+                                    turn: newAssistantTurn,
+                                    modelId: selectedModel ?? "default",
+                                    modelOptions: activeModelOptions
+                                ) { [weak self] in self?.rebuildVisibleBlocks() }
+                                rebuildVisibleBlocks()
```
tpae, line 1067: @RaajeevChandran can we check the performance hit on this one (if there's any)

Packages/OsaurusCore/Services/Provider/RemoteProviderService.swift
ritave: I've extracted OsaurusProvider, but it feels messier than what's done currently.

There are still switch cases in RemoteProviderService which needs to be exhaustive so there's still logic around Osaurus in there

Packages/OsaurusCore/Networking/HTTPHandler.swift
```diff
@@ -299,6 +299,22 @@ final class HTTPHandler: ChannelInboundHandler, Sendable {
                 handleMemoryIngest(head: head, context: context, startTime: startTime, userAgent: userAgent)
             } else if head.method == .GET, path == "/agents" {
                 handleListAgents(head: head, context: context, startTime: startTime, userAgent: userAgent)
+            } else if head.method == .GET, path.hasPrefix("/agents/") {
+                handleGetAgentEndpoint(
```
RaajeevChandran, line 303: this and the `else if head.method == .POST, path.hasPrefix("/agents/"), path.hasSuffix("/run")` could be single else if block since both these cases call the same method with no change in arguments

Packages/OsaurusCore/Views/Chat/ChatView.swift
RaajeevChandran: Server side tool completions are infrequent (one per tool call) and not per streaming delta. so these rebuilds are cheap and shouldn't be hurting performance

PR #796, opened 2026-04-06
added metal gate to prevent crashes

tpae:
## Summary

added metal gate

## Checklist

### DIFF
--- a/Packages/OsaurusCore/Services/ModelRuntime.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime.swift
@@ -470,12 +470,6 @@ actor ModelRuntime {
 
         genLog.info("generateEventStream: start model=\(modelName, privacy: .public)")
 
-        await MetalGate.shared.enterGeneration()
-        if Task.isCancelled {
-            await MetalGate.shared.exitGeneration()
-            throw CancellationError()
-        }
-
         let effectiveStopSequences = stopSequences
         let cfg = await getConfig()
         let holder = try await loadContainer(id: modelId, name: modelName)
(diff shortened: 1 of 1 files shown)


jjang-ai:
## Review: MetalGate architecture

Overall this is a solid improvement — replacing the one-shot startup gate with always-on mutual exclusion is the right call. The actor-based concurrency model is correct, the `while` + `withCheckedContinuation` double-check handles re-entrancy properly, and the `gatedGenTask` wrapper with `withTaskCancellationHandler` is clean.

The test coverage is good and the removal of 6 scattered `awaitStartupInit()` call sites in favor of automatic coordination through `MetalSafeEmbedder` is a nice simplification.

### Potential issue: gate leak on `loadContainer` failure

In `generateEventStream`, after `enterGeneration()` is called, there are throwing calls **outside** the `do { ... } catch` blocks that have `exitGeneration()`:

```swift
await MetalGate.shared.enterGeneration()      // gate acquired
if Task.isCancelled { ... exitGeneration() }   // ✅ handled

let cfg = await getConfig()                     // probably doesn't throw
let holder = try await loadContainer(...)       // ⚠️ CAN THROW — no exitGeneration on this path

// ... ~55 lines of setup ...

do {                                            // exitGeneration() lives in catch blocks down here
```

If `loadContainer` throws (model not found, disk error, etc.), `exitGeneration()` is never called. The gate stays locked permanently — all future embeddings block forever.

Suggested fix: wrap the throwing setup in a do/catch that releases the gate on failure, or move `enterGeneration()` to just before the inner `do` block (after all throwing setup completes).

PR #819, opened 2026-04-09
fix: hide placeholder during IME composition to prevent text overlap

SliverKeigo:
## Summary
- Fix placeholder text overlapping with IME composing text (marked text) in the chat input field
- Affects CJK input methods (Chinese pinyin, Japanese, Korean, etc.) where characters are composed before being committed
- Track `setMarkedText`/`unmarkText` state in `CustomNSTextView` and suppress the placeholder while the IME is actively composing

## Test plan
- [ ] Type in the chat input using a CJK input method (e.g. Chinese pinyin) — placeholder should disappear as soon as composing begins
- [ ] Confirm placeholder reappears after committing or cancelling the composition
- [ ] Verify English/direct input still works as before (placeholder hides on first character)
- [ ] Verify placeholder still shows when input is empty and no attachments are present


tpae (approved):
thank you for your contribution! this looks good

SteveShi:
The problem is that the Chinese input method isn't entering the first character properly; it disappears automatically. I'm using the built-in Chinese input method in macOS.

PR #836, opened 2026-04-12
build: pin MLX dependencies and fix current FluidAudio compatibility

mimeding:
## Scope
- fix speech model initialization to match the current FluidAudio API
- pin MLX-related dependencies to exact revisions
- sync the Xcode/workspace lockfiles to the pinned dependency set
- add an optional `LOCAL_MLX_SWIFT_LM_PATH` override for coordinated local MLX development

## Resulting Improvements
- Build resolution becomes reproducible instead of depending on mutable MLX-related refs.
- The current speech-service path works again against the FluidAudio API this branch targets.
- Xcode and SwiftPM dependency state stay aligned more reliably.
- Cross-repo MLX development becomes easier to validate locally without repeated manifest edits.

## Detailed Improvement Explanation
This PR makes the repository behave more like a controlled product build instead of an ad hoc local setup. Pinning the MLX-related graph removes a class of failures where the same checkout resolves differently over time or between machines. Syncing the Xcode lockfiles closes the gap between command-line and IDE package state. The FluidAudio fix restores the currently expected speech initialization path, so the branch is not only more reproducible but also compatible with the dependency versions it declares.

## Why This PR Exists
This branch addresses two concrete problems:
1. `SpeechService` was calling an outdated FluidAudio initialization path.
2. MLX-related dependencies were being resolved from mutable refs, which made local and CI builds harder to reason about.

## Validation
- validated a Release app build against a local MLX checkout
- confirmed the branch stays scoped to build compatibility and dependency reproducibility
- cleaned and audited the published diff for local-path and temp-file leakage

## Merge Notes
- This is the build/reproducibility base for the local MLX and Work Mode follow-up work.
- `.mov` / broader media forwarding stays intentionally out of this PR.
- MLX-specific hardening and VLM changes are proposed separately in `osaurus-ai/mlx-swift-lm`.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,13 +1,13 @@
 {
-  "originHash" : "6d825c4d68c47e18f5d4607729be796d583f1ad7ace433a900815fdb0bf43c67",
+  "originHash" : "d22ff3a751a5cfed0fc3949c0b920f3b8b245231d1fcbb5bb8084cb8c2d8ff45",
   "pins" : [
     {
       "identity" : "async-http-client",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/swift-server/async-http-client.git",
       "state" : {
-        "revision" : "2fc4652fb4689eb24af10e55cabaa61d8ba774fd",
-        "version" : "1.32.0"
+        "revision" : "3a5b74a58782c3b4c1f0bc75e9b67b10c2494e8f",
+        "version" : "1.33.1"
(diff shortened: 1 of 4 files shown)


mimeding:
**Resulting Improvements**

- Builds become reproducible because the MLX-related package graph is pinned instead of floating on mutable branch heads.
- Current speech initialization works again against the FluidAudio API surface this branch targets.
- Xcode and SwiftPM resolution stay aligned, which reduces "works in one build path but not the other" drift.
- Local cross-repo MLX development becomes practical through the path override instead of repeated manifest surgery.

mimeding:
Closing this because it is an omnibus integration branch and no longer meets the PR hygiene bar. The relevant fixes are being carried forward in focused, validated PRs instead of this mixed branch.

PR #838, opened 2026-04-12
fix: harden MLX cache reuse under long-context pressure

mimeding:
## Summary
- harden local MLX generation against long-context cache corruption and Metal allocation failures
- stop reusing wrapped rotating caches as if they were valid prefix caches
- make pre-generation snapshotting adaptive so oversized cache copies are skipped instead of crashing
- add focused regression tests for adaptive KV retry logic and rotating-cache guards

## What changed
- `ModelRuntime` now learns lower `maxKV` / `prefillStep` caps from Metal allocation failures and retries with tighter settings
- `MLXGenerationEngine` now:
  - logs effective KV/runtime settings
  - skips prefix reuse when a `RotatingKVCache` has already wrapped
  - snapshots caches before generation starts mutating them
  - skips snapshot copies when cache size exceeds a machine-scaled byte limit
- `KVCacheStore` now uses detached MLX array copies for transformer-style cache snapshots to avoid aliasing cached Metal buffers into live generation
- `RuntimeConfig` now supports capped runtime variants used by adaptive retries

## Why this matters
This isolates the local-runtime failure class behind recent Gemma 4 regressions:
- crashes during large prefills
- garbage output after rotating-cache wrap
- snapshot-related pressure spikes during long sessions

## Validation
Passing deterministic coverage in the compatibility-fixed local environment:
- `swift test --package-path /Users/mmeding/Documents/Claude/Projects/osaurus/.local-worktrees/workmode-all-local/Packages/OsaurusCore --filter ModelRuntimeAdaptiveKVTests`
- `swift test --package-path /Users/mmeding/Documents/Claude/Projects/osaurus/.local-worktrees/workmode-all-local/Packages/OsaurusCore --filter 'hasWrappedRotatingCache|snapshotCopyByteLimit|shouldSkipSnapshotCopy'`

Additional note:
- Clean-branch validation against `upstream/main` is currently blocked by an unrelated repo-head compile break in `Packages/OsaurusCore/Managers/SpeechService.swift` (`AsrManager.initialize(models:)` no longer exists in the current FluidAudio API).
- That compatibility fix is intentionally not folded into this PR.

## Out of scope
- prompt hygiene / stale-history filtering
- attached-document retrieval changes
- UI or settings changes

### DIFF
--- a/Packages/OsaurusCore/Services/ModelRuntime.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime.swift
@@ -57,6 +57,8 @@ actor ModelRuntime {
     private var currentModelName: String?
     private var kvCacheStore = KVCacheStore()
     private var cachedConfig: RuntimeConfig?
+    private var learnedMaxKVCaps: [String: Int] = [:]
+    private var learnedPrefillStepCaps: [String: Int] = [:]
     private var activeGenerationTask: Task<Void, Never>?
     // modelName:taskHash -> Task
     private var prefixCacheTasks: [String: Task<Void, Never>] = [:]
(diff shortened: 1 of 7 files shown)


mimeding:
Closing this because it is no longer a truthful draft against current `main`. I attempted a clean re-port directly onto current `upstream/main` and hit structural conflicts: the branch edits the old local KV/snapshot runtime path (`ModelRuntime.swift`, `RuntimeConfig.swift`, `MLXGenerationEngine.swift`, `KVCacheStore.swift`, related tests), while current `main` has moved cache ownership into `vmlx-swift-lm`'s `CacheCoordinator` path.

That means this PR is not in a state where it can be made credible with a rebase-only update. If we still need runtime hardening here, it should come back as a new narrow PR that targets the current `CacheCoordinator` architecture and is validated there with real long-context reproduction and tests.

PR #839, opened 2026-04-12
fix: restore SpeechService compatibility with current FluidAudio API

mimeding:
## Summary
- update `SpeechService` to call the current `AsrManager.loadModels(_:)` API
- restore clean compilation on `upstream/main` after the FluidAudio API change
- keep the fix isolated so follow-on runtime work can be reviewed separately

## Validation
- `swift build --package-path /Users/mmeding/Documents/Claude/Projects/osaurus/.local-worktrees/pr0-fluidaudio-asrmanager-api-compat/Packages/OsaurusCore`

## Context
This is a narrow compatibility unblocker for [#838](https://github.com/osaurus-ai/osaurus/pull/838), whose clean-branch validation was previously failing on `SpeechService` before runtime tests could run.

### DIFF
--- a/Packages/OsaurusCore/Managers/SpeechService.swift
+++ b/Packages/OsaurusCore/Managers/SpeechService.swift
@@ -661,7 +661,7 @@ public final class SpeechService: ObservableObject {
             }.value
 
             let manager = AsrManager(config: .default)
-            try await manager.initialize(models: models)
+            try await manager.loadModels(models)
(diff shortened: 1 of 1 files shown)


mimeding:
Closing this because current upstream/main already contains the correct SpeechService loadModels integration for the latest FluidAudio API. Keeping this PR open would now reintroduce the wrong API call instead of fixing anything.

PR #853, opened 2026-04-13
sandbox fixes and other updates

tpae:
## Summary

replace deprecated container logic

## Checklist

### DIFF
--- a/Packages/OsaurusCore/Services/Context/PreflightCapabilitySearch.swift
+++ b/Packages/OsaurusCore/Services/Context/PreflightCapabilitySearch.swift
@@ -107,9 +107,8 @@ enum CapabilitySearch {
         )
     }
 
-    static func canCreatePlugins() async -> Bool {
+    static func canCreatePlugins(agentId: UUID) async -> Bool {
         await MainActor.run {
-            let agentId = AgentManager.shared.activeAgent.id
             guard let config = AgentManager.shared.effectiveAutonomousExec(for: agentId) else { return false }
             return config.enabled && config.pluginCreate
         }
@@ -124,7 +123,7 @@ enum PreflightCapabilitySearch {
 
     // MARK: Search
 
-    static func search(query: String, mode: PreflightSearchMode = .balanced) async -> PreflightResult {
+    static func search(query: String, mode: PreflightSearchMode = .balanced, agentId: UUID) async -> PreflightResult {
         guard mode != .off,
             !query.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
         else { return .empty }
@@ -140,7 +139,7 @@ enum PreflightCapabilitySearch {
             return (sorted, groupMap)
         }
 
-        guard !catalog.isEmpty else { return .empty }
+        guard !catalog.isEmpty else { return await sandboxPluginCreatorFallback(agentId: agentId) }
 
         InferenceProgressManager.shared.preflightWillStartAsync()
         defer { InferenceProgressManager.shared.preflightDidFinishAsync() }
@@ -153,7 +152,7 @@ enum PreflightCapabilitySearch {
         )
 
         if selectedNames.isEmpty {
-            return .empty
+            return await sandboxPluginCreatorFallback(agentId: agentId)
         }
 
         let (toolSpecs, items) = await MainActor.run {
@@ -277,8 +276,8 @@ enum PreflightCapabilitySearch {
 
     // MARK: Fallback
 
-    private static func sandboxPluginCreatorFallback() async -> PreflightResult {
-        guard await CapabilitySearch.canCreatePlugins() else { return .empty }
+    private static func sandboxPluginCreatorFallback(agentId: UUID) async -> PreflightResult {
+        guard await CapabilitySearch.canCreatePlugins(agentId: agentId) else { return .empty }
         let skill = await MainActor.run {
             SkillManager.shared.skill(named: "Sandbox Plugin Creator")
         }
--- a/Packages/OsaurusCore/Services/Context/PreflightCapabilitySearch.swift
+++ b/Packages/OsaurusCore/Services/Context/PreflightCapabilitySearch.swift
@@ -142,6 +142,9 @@ enum PreflightCapabilitySearch {
 
         guard !catalog.isEmpty else { return .empty }
 
+        InferenceProgressManager.shared.preflightWillStartAsync()
+        defer { InferenceProgressManager.shared.preflightDidFinishAsync() }
+
         let selectedNames = await selectTools(
             query: query,
             catalog: catalog,
--- a/Packages/OsaurusCore/Services/Sandbox/SandboxManager.swift
+++ b/Packages/OsaurusCore/Services/Sandbox/SandboxManager.swift
@@ -84,7 +84,7 @@
             return _status
         }
 
-        private var staleContainerDir: URL {
+        var staleContainerDir: URL {
             OsaurusPaths.container().appendingPathComponent("containers/\(Self.containerID)")
         }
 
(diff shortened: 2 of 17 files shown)


Packages/OsaurusCore/Services/Sandbox/SandboxManager.swift
```diff
@@ -879,6 +893,21 @@
             }
         }
 
+        private static func friendlyError(from error: Error) -> Error {
```
RaajeevChandran, line 896: this converts the error to a string via String(describing: error) and then checks for substrings like "GRPC" and "vmnet". If the underlying framework changes its error description text, this match will break silently and the user would get a raw unreadable error instead of the friendly one. we instead need to match on concrete error types (like, error as? SomeGRPCError) rather than display strings.

Packages/OsaurusCore/Services/Sandbox/SandboxManager.swift
```diff
@@ -797,6 +797,20 @@
             )
         }
 
+        /// Polls until the guest can reach the Alpine CDN, so plugins that
+        /// run `apk add` right after provisioning don't hit DNS failures.
+        private func waitForNetwork() async {
+            for attempt in 1 ... 5 {
+                let result = try? await exec(
```
RaajeevChandran, line 804: If all 5 wget attempts fail, this method returns normally and provisioning continues as if the network is available. Downstream operations that depend on network (like apk add in plugins) will then fail with confusing connection errors that are harder to diagnose than a clear "network not ready" message. 


we need to either 
(a) throwing after exhausting retries so the provisioning flow can surface a meaningful error or 

(b) at minimum logging a prominent warning like NSLog([SandboxManager] WARNING: Network connectivity check failed after 5 attempts. Plugins requiring network may fail") so there's a breadcrumb when things break later.

Packages/OsaurusCore/Tools/CapabilityTools.swift
```diff
@@ -66,7 +72,14 @@ final class CapabilitiesSearchTool: OsaurusTool, @unchecked Sendable {
         )
 
         if hits.isEmpty {
-            if await CapabilitySearch.canCreatePlugins() {
+            let id: UUID
```
RaajeevChandran, line 75: WorkExecutionContext.currentAgentId is a thread local value that represents the agent at the time execution started. If the user switches agents while a tool call is in flight or if a work session runs a tool on behalf of a different agent than the one currently active in the UI, this could resolve to the wrong agent and either incorrectly allow or deny plugin creation. Worth adding a comment documenting the expected lifecycle and which value is authoritative or pinning the agent ID at the call site rather than esolving it lazily inside the tool.

Packages/OsaurusCore/Services/Sandbox/SandboxManager.swift
```diff
@@ -776,6 +768,15 @@
 
         // MARK: - Private Helpers
 
+        private func cleanupAfterFailure() async {
```
RaajeevChandran: this is currently running 5 cleanup steps, each wrapped in try?: stopping the container, deleting it from the manager, nil-ing out references, removing the container directory and stopping the bridge server. 

if any step fails silently (like if container.stop() hangs or throws), the subsequent steps still run. but there's no verification that the system is actually in a clean state afterward. if removeItem(at: staleContainerDir) fails due to permissions, the next provision attempt may find a stale directory and behave unexpectedly. 

since this is error path code that's hard to test manually, we should probably add a test that simulates a failed provision and verifies the cleanup leaves the system in a provisionable state

Packages/OsaurusCore/Views/Chat/NativeBlockViews.swift
```diff
@@ -125,18 +125,27 @@ final class NativeTypingIndicatorView: NSView {
...
         cancellables.append(sink)
     }
 
-    private func setLoadingModelState(_ loading: Bool) {
-        guard loading != isShowingLoadingLabel else { return }
-        isShowingLoadingLabel = loading
-        loadingLabel.isHidden = !loading
-        dotStack.isHidden = loading
-        memoryStack?.isHidden = loading
+    private func setLoadingModelState(loadingModel: Bool, preflighting: Bool) {
```
RaajeevChandran, line 136: If both preflighting and the model loading fire at the same time (like if the model is still loading when preflight kicks off or they overlap briefly), the user sees "Searching capabilities..." which is misleading because the model isn't ready yet. 

We need to decide whether model loading should take priority since it's a prerequisite for inference or whether the states are truly mutually exclusive in practice. If they can overlap, a priority order like loadingModel > preflighting would be more accurate for UX purposes.

Packages/OsaurusCore/Services/Context/PreflightCapabilitySearch.swift
```diff
@@ -140,7 +139,10 @@ enum PreflightCapabilitySearch {
             return (sorted, groupMap)
         }
 
-        guard !catalog.isEmpty else { return .empty }
+        guard !catalog.isEmpty else { return await sandboxPluginCreatorFallback(agentId: agentId) }
```
RaajeevChandran, line 142: Previously, when the tool catalog was empty or when selectTools returned no matches (line 155), the method returned .empty meaning no tools, no context. Now both paths call sandboxPluginCreatorFallback(agentId:) which checks if plugin creation is enabled and, if so, injects the "Sandbox Plugin Creator" skill into the preflight result. 

This means agents with autonomousExec.pluginCreate enabled will now be offered the plugin creator skill in situations where they previously got nothing. If it's a side effect of passing agentId through, the old .empty returns should be preserved for these paths.

RaajeevChandran (changes requested):

Packages/OsaurusCore/Services/Context/PreflightCapabilitySearch.swift
tpae: what do you mean? this is how it needs to work. if there's no tools matched, and if user has plugin creation enabled, it should inject the plugin creator skill. that's the intention.

Packages/OsaurusCore/Views/Chat/NativeBlockViews.swift
tpae: i don't think so. the logic doesn't work that way. models get loaded on first inference call, preflight happens before inference gets called. having it backwards will introduce more logic which is not ideal trade-off.

Packages/OsaurusCore/Tools/CapabilityTools.swift
tpae: no. currentAgentId is shaky implementation that doesn't work with background agents. Osaurus is a local AI server that can work across networks (local area network discovery), and background executions will need to be considered.

Packages/OsaurusCore/Services/Sandbox/SandboxManager.swift
tpae: agree this is a bad implementation but generally it's to prevent race conditions and usually never hits the 2nd (sometimes it does).

Packages/OsaurusCore/Services/Sandbox/SandboxManager.swift
tpae: the errors are very specific and it doesn't display any rich context. not worth optimizing imo

Packages/OsaurusCore/Views/Chat/NativeBlockViews.swift
RaajeevChandran: makes sense

Packages/OsaurusCore/Services/Sandbox/SandboxManager.swift
RaajeevChandran: okay

RaajeevChandran (approved):

jjang-ai (approved):
Approved — sandbox cleanup + new effectiveToolsDisabled/effectiveMemoryDisabled resolvers + preflight TTFT fix all look clean.

PR #861, opened 2026-04-15
Improve work prompt tool guidance foundation

mimeding:
## Problem
The harness currently exposes tools through their raw function schemas and generic descriptions. That leaves the model to infer when a tool should be used, when it should be avoided, and which tools are terminal versus preparatory. In practice this increases malformed tool use, avoidable retries, and wasted inference on work-mode tasks.

## Technical reason
Reliable tool use needs a second layer beyond JSON schema: concise operational guidance that fits inside the system prompt budget. This change adds internal metadata for high-value work tools and renders that metadata into budgeted tool cards. The cards are intentionally short, sanitized, and ordered by prompt priority so the prompt adds decision support without turning into an unbounded glossary. The behavior is gated behind a dedicated rollout flag so it can be enabled deliberately and expanded phase by phase.

## Why this matters in real use
When the model can distinguish between discovery tools, mutation tools, coordination tools, and completion tools, it spends less time thrashing between the wrong actions. That translates into fewer failed first attempts, faster completion on file-heavy work, and clearer end-user outcomes because completion and artifact-sharing expectations are stated up front.

## Implementation summary
- Added `ToolMetadata` and `ToolPromptCard` support for concise, priority-ordered tool guidance.
- Added a metadata catalog for the core work tools and sandbox tools used in work mode.
- Extended `OsaurusTool` with internal prompt metadata and taught `ToolRegistry` to render a budgeted work-mode tool guide from the active tool set.
- Appended the tool guide to the structured work prompt in `SystemPromptComposer`.
- Added `HarnessReliabilityFlags.toolPromptFoundation` so the new prompt section is disabled by default and can be enabled explicitly.
- Added tests for disabled-by-default behavior, enabled rendering, prompt budget limits, and sanitization against prompt-shaped markup.

## Failure modes and safeguards
- Prompt bloat: the tool guide is budget-capped and ordered by prompt priority.
- Prompt injection through tool copy: rendered cards collapse whitespace, strip backticks, and truncate long text.
- Premature rollout: the new section is behind `OSAURUS_HARNESS_TOOL_PROMPT_FOUNDATION` / `HarnessReliability.ToolPromptFoundationEnabled` and remains off by default.
- Regression in existing work prompts: existing workflow guidance tests were retained and updated for the new actor boundary.

## Build and test evidence
- `swift-format lint --strict --recursive Packages App`
- `swiftlint lint` completed with the repository's existing warning baseline and no new serious violations.
- `xcodebuild -workspace osaurus.xcworkspace -scheme osaurus -configuration Debug -destination 'platform=macOS' -derivedDataPath build/DerivedData CODE_SIGN_IDENTITY='-' CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO build`
- `xcodebuild test -workspace osaurus.xcworkspace -scheme OsaurusCoreTests -destination 'platform=macOS' -derivedDataPath build/DerivedData -resultBundlePath build/results/phase1-osauruscore-full-postflag.xcresult -skip-testing:OsaurusCoreTests/KVCacheStoreTests -skip-testing:OsaurusCoreTests/MLXGenerationEngineTests`
- `xcodebuild test -workspace osaurus.xcworkspace -scheme OsaurusCLITests -destination 'platform=macOS' -derivedDataPath build/DerivedData -resultBundlePath build/results/phase1-osauruscli.xcresult`
- Focused regression runs also passed for `WorkPromptToolGuideTests` and `WorkExecutionEngineTests`.
- Local launch smoke check: opened the built macOS app bundle successfully and confirmed the `osaurus` process started.

## Security evidence
- Added negative coverage proving prompt-card rendering sanitizes prompt-shaped markup and fenced code snippets before insertion into the system prompt.
- The new metadata is internal-only and does not widen tool permissions or execution surface area.
- Rollout is fail-closed: disabled by default unless the flag is explicitly enabled.

## Rollout and fallback
- Flag: `HarnessReliability.ToolPromptFoundationEnabled`
- Environment override for local testing: `OSAURUS_HARNESS_TOOL_PROMPT_FOUNDATION=1`
- Default state: off
- Fallback: disable the flag to return to the prior work prompt shape without reverting code.

## Out of scope
- No change to tool execution policy or permission handling.
- No new analytics or runtime telemetry.
- No parallel execution, denial normalization, or compaction recovery changes yet.

### DIFF
--- a/Packages/OsaurusCore/Tests/Work/WorkPromptToolGuideTests.swift
+++ b/Packages/OsaurusCore/Tests/Work/WorkPromptToolGuideTests.swift
@@ -1,0 +1,97 @@
+import Foundation
+import Testing
+
+@testable import OsaurusCore
+
+@Suite(.serialized)
+@MainActor
+struct WorkPromptToolGuideTests {
+    @Test
+    func toolGuideIsDisabledByDefault() {
+        let context = WorkFolderContext(
+            rootPath: FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString),
+            projectType: .swift,
+            tree: "./\nPackage.swift\nSources/\n",
+            manifest: "Package.swift",
+            gitStatus: "## main\n M Sources/App.swift",
+            isGitRepo: true
+        )
+
+        let (prompt, _) = SystemPromptComposer.composeWorkPrompt(
+            base: "Base prompt",
+            executionMode: .hostFolder(context)
+        )
+
+        #expect(!prompt.contains("## Tool Guide"))
+    }
+
+    @Test
+    func hostFolderPromptIncludesBudgetedToolGuide() throws {
+        let context = WorkFolderContext(
+            rootPath: FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString),
+            projectType: .swift,
+            tree: "./\nPackage.swift\nSources/\n",
+            manifest: "Package.swift",
+            gitStatus: "## main\n M Sources/App.swift",
+            isGitRepo: true
+        )
+
+        let prompt = HarnessReliabilityFlags.withOverride(true, for: .toolPromptFoundation) {
+            let (prompt, _) = SystemPromptComposer.composeWorkPrompt(
+                base: "Base prompt",
+                executionMode: .hostFolder(context)
+            )
+            return prompt
+        }
+
+        #expect(prompt.contains("## Tool Guide"))
+        #expect(prompt.contains("`file_read`"))
+        #expect(prompt.contains("`file_edit`"))
+        #expect(prompt.contains("`shell_run`"))
+        #expect(prompt.contains("`complete_task`"))
+        #expect(prompt.contains("`share_artifact`"))
+    }
+
+    @Test
+    func compactToolGuideStaysWithinBudget() {
+        let context = WorkFolderContext(
+            rootPath: FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString),
+            projectType: .swift,
+            tree: "./\nPackage.swift\nSources/\n",
+            manifest: "Package.swift",
+            gitStatus: "## main\n M Sources/App.swift",
+            isGitRepo: true
+        )
+
+        let guide = HarnessReliabilityFlags.withOverride(true, for: .toolPromptFoundation) {
+            ToolRegistry.shared.workPromptGuide(mode: .hostFolder(context), compact: true)
+        }
+
+        #expect(guide.contains("## Tool Guide"))
+        #expect(guide.count <= ToolMetadataCatalog.promptGuideBudget(compact: true))
+    }
+}
+
+struct ToolPromptCardTests {
+    @Test
+    func sanitizesPromptInjectedMarkup() {
+        let metadata = ToolMetadata(
+            purpose: "Read the file",
+            whenToUse: "checking the implementation",
+            avoidWhen: "editing unread files",
+            example: "```json\n{\"path\":\"App.swift\"}\n```",
+            promptPriority: 50
+        )
+
+        let card = ToolPromptCard(
+            name: "file_read\n## injected",
+            description: "Read a file\n## injected",
+            metadata: metadata
+        )
+        let rendered = card.render(compact: false)
+
+        #expect(!rendered.contains("```"))
+        #expect(!rendered.contains("\n## injected"))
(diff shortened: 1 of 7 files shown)


Packages/OsaurusCore/Tools/ToolPromptSupport.swift
```diff
@@ -0,0 +1,434 @@
...
+    let example: String?
+    let promptPriority: Int
+
+    init(name: String, description: String, metadata: ToolMetadata) {
+        self.name = Self.sanitize(name, maxLength: 80)
+        let fallbackSummary = description.isEmpty ? metadata.purpose : description
+        self.summary = Self.sanitize(metadata.purpose.isEmpty ? fallbackSummary : metadata.purpose, maxLength: 160)
+        self.whenToUse = Self.optionalSanitize(metadata.whenToUse, maxLength: 160)
+        self.avoidWhen = Self.optionalSanitize(metadata.avoidWhen, maxLength: 160)
+        self.example = Self.optionalSanitize(metadata.example, maxLength: 160)
```
RaajeevChandran: Examples are only rendered in non-compact mode (line 81) but truncation to 160 chars happens unconditionally at construction time. With a 2400-char non-compact budget the JSON examples in the catalog could afford more room

Packages/OsaurusCore/Tools/ToolPromptSupport.swift
```diff
@@ -0,0 +1,434 @@
...
+
+    static func promptGuideBudget(compact: Bool) -> Int {
+        compact ? 1100 : 2400
+    }
+
+    private static func promptCards(
+        toolDescriptions: [String: String],
+        executionMode: WorkExecutionMode
+    ) -> [ToolPromptCard] {
+        let orderedNames = orderedPromptToolNames(for: executionMode)
```
RaajeevChandran: the declaration order in orderedPromptToolNames (line 391) only affects deduplication via seen, since promptCards re-sorts everything by promptPriority anyway. the ordering in orderedPromptToolNames is giving a false impression of control over the final card sequence

Packages/OsaurusCore/Tests/Work/WorkPromptToolGuideTests.swift
RaajeevChandran, line 1: can you add a test covering the `.sandbox` execution mode?

`hostFolderPromptIncludesBudgetedToolGuide` covers host folder but sandbox tool names (like sandbox_read_file) have no equivalent assertion

RaajeevChandran (changes requested):

mimeding:
Rebased this branch onto current `main` and force-pushed the refreshed head `24abedda` from a clean Phase 0 worktree.

Addressed follow-up items on the branch head:
- kept the sandbox tool-guide coverage on the refreshed branch
- preserved the prompt/tool metadata updates after the rebase
- reran targeted local validation with `swift test --package-path Packages/OsaurusCore --filter 'WorkPromptToolGuideTests|WorkExecutionEngineTests'` and got 17 passing tests\n\nAt this point the remaining GitHub action is review-thread cleanup rather than additional code work.

Packages/OsaurusCore/Tools/ToolPromptSupport.swift
mimeding: Expanded the non-compact example budget and still keep examples out of compact mode, so this is covered on the current head.

Packages/OsaurusCore/Tools/ToolPromptSupport.swift
mimeding: Clarified this in code: candidate ordering only affects inclusion and dedup, while the final rendered order comes from `promptPriority`.

Packages/OsaurusCore/Tests/Work/WorkPromptToolGuideTests.swift
mimeding: Added sandbox coverage with `sandboxPromptIncludesBudgetedToolGuide()`, so the `.sandbox` path is now exercised too.

mimeding:
Maintainer follow-up on April 16, 2026: the previously open review items on this branch have been addressed on the current head (`24abedda`), including the sandbox prompt-guide regression coverage and the tool-guide ordering/budget clarifications. I’ve replied on the affected threads and resolved them so the remaining action here is reviewer re-check rather than more implementation work.

PR #862, opened 2026-04-15
fix: tighten prompt hygiene for local chat

mimeding:
## Summary
- tighten local-chat tool selection so local auto mode only loads on-demand capability tools when the query actually needs them
- treat slash-style installed local model ids as local models for compact prompt selection
- detect and strip obviously corrupted assistant output so repeated loops and leaked channel markers do not poison later turns

## Why
This packages the prompt-hygiene slice from the local inference recovery work as a separate change from the MLX runtime and attached-document work. The main user-visible effect is cleaner prompt assembly for local chat and safer handling of malformed streamed output.

## Validation
- `swift test --package-path Packages/OsaurusCore --filter ChatViewSandboxTests`

### DIFF
--- a/Packages/OsaurusCore/Services/Chat/SystemPromptComposer.swift
+++ b/Packages/OsaurusCore/Services/Chat/SystemPromptComposer.swift
@@ -183,6 +183,8 @@ public struct SystemPromptComposer: Sendable {
         let toolMode = AgentManager.shared.effectiveToolSelectionMode(for: agentId)
         let isManual = toolMode == .manual
 
+        // Work mode keeps the full tool set available. On-demand narrowing is chat-only so
+        // autonomous task execution never silently loses a required tool.
         if preferOnDemandTools && !isManual, case .none = executionMode {
             return resolveOnDemandLocalChatTools(query: query, preflight: preflight)
         }
@@ -239,6 +241,8 @@ public struct SystemPromptComposer: Sendable {
 
     private static func queryLikelyNeedsCapabilityDiscovery(_ query: String) -> Bool {
         let normalized = query.lowercased()
+        // This is intentionally a loose first-pass heuristic. Prefer a few false positives in
+        // local auto chat over hiding capability tools when the user is asking what Osaurus can do.
         let keywords = [
             "tool", "tools", "skill", "skills", "plugin", "plugins", "method", "methods", "capability",
             "capabilities", "workflow", "workflows",
--- a/Packages/OsaurusCore/Services/Chat/SystemPromptComposer.swift
+++ b/Packages/OsaurusCore/Services/Chat/SystemPromptComposer.swift
@@ -83,6 +83,7 @@ public struct SystemPromptComposer: Sendable {
             composer: composer,
             agentId: agentId,
             executionMode: executionMode,
+            preferOnDemandTools: model.map(SystemPromptTemplates.isLocalModel) ?? false,
             query: query,
             toolsDisabled: toolsDisabled,
             model: model,
@@ -98,6 +99,7 @@ public struct SystemPromptComposer: Sendable {
         composer: SystemPromptComposer,
         agentId: UUID,
         executionMode: WorkExecutionMode,
+        preferOnDemandTools: Bool = false,
         query: String,
         toolsDisabled: Bool,
         model: String? = nil,
@@ -137,13 +139,18 @@ public struct SystemPromptComposer: Sendable {
             agentId: agentId,
             executionMode: executionMode,
             toolsDisabled: effectiveToolsOff,
-            preflight: preflight
+            preflight: preflight,
+            query: query,
+            preferOnDemandTools: preferOnDemandTools
         )
         trace?.mark("resolve_tools_done")
 
         let manifest = comp.manifest()
         let toolNames = tools.map { $0.function.name }
         debugLog("[Context] \(manifest.debugDescription)")
+        debugLog(
+            "[Context] resolvedTools count=\(toolNames.count) preferOnDemand=\(preferOnDemandTools) names=\(toolNames)"
+        )
 
         let rendered = comp.render()
         trace?.set("systemPromptChars", rendered.count)
@@ -167,12 +174,18 @@ public struct SystemPromptComposer: Sendable {
         agentId: UUID,
         executionMode: WorkExecutionMode,
         toolsDisabled: Bool = false,
-        preflight: PreflightResult = .empty
+        preflight: PreflightResult = .empty,
+        query: String = "",
+        preferOnDemandTools: Bool = false
     ) -> [Tool] {
         guard !toolsDisabled else { return [] }
 
         let toolMode = AgentManager.shared.effectiveToolSelectionMode(for: agentId)
         let isManual = toolMode == .manual
+
+        if preferOnDemandTools && !isManual, case .none = executionMode {
+            return resolveOnDemandLocalChatTools(query: query, preflight: preflight)
+        }
 
         var tools = ToolRegistry.shared.alwaysLoadedSpecs(
             mode: executionMode,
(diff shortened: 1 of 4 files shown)


Packages/OsaurusCore/Services/Chat/SystemPromptComposer.swift
```diff
@@ -167,13 +174,19 @@ public struct SystemPromptComposer: Sendable {
...
+        query: String = "",
+        preferOnDemandTools: Bool = false
     ) -> [Tool] {
         guard !toolsDisabled else { return [] }
 
         let toolMode = AgentManager.shared.effectiveToolSelectionMode(for: agentId)
         let isManual = toolMode == .manual
 
+        if preferOnDemandTools && !isManual, case .none = executionMode {
+            return resolveOnDemandLocalChatTools(query: query, preflight: preflight)
```
RaajeevChandran, line 189: `resolveOnDemandLocalChatTools` is MainActor but resolveTools (line 173) has no actor    
annotation. The call at line 187 requires the caller to already be on the main actor, but there's no compile time enforcement. In strict concurrency this could surface as a warning or error. So let's either add MainActor to resolveTools for consistency or make the on-demand path actor agnostic

Packages/OsaurusCore/Services/Chat/SystemPromptComposer.swift
```diff
@@ -197,6 +210,42 @@ public struct SystemPromptComposer: Sendable {
...
+
+        for spec in preflight.toolSpecs
+        where seen.insert(spec.function.name).inserted {
+            tools.append(spec)
+        }
+
+        return tools
+    }
+
+    private static func queryLikelyNeedsCapabilityDiscovery(_ query: String) -> Bool {
```
RaajeevChandran, line 242: Substring matching on words like "tool", "method", "plugin" will false positive on benign queries (like "What's the best method for making pasta?") and false-negative on capability queries that don't use these exact words (like "What can you help me with?"). This is probably fine as a first pass heuristic but a comment explaining it's intentionally loose would help future maintainers calibrate changes

Packages/OsaurusCore/Services/Chat/SystemPromptComposer.swift
```diff
@@ -249,6 +298,7 @@ public struct SystemPromptComposer: Sendable {
             composer: composer,
             agentId: agentId,
             executionMode: executionMode,
+            preferOnDemandTools: false,
```
RaajeevChandran, line 322: please add a comment like "Work mode always loads the full tool set would" to prevent someone from wondering whether this should be wired up later

Packages/OsaurusCore/Views/Chat/ChatView.swift
```diff
@@ -347,6 +347,98 @@ final class ChatSession: ObservableObject {
...
+
+        let lowercased = sanitized.lowercased()
+        let markerCandidates = [
+            "<channel>|<channel>thought",
+            "<channel><|channel|>thought",
+            "<channel><channel>thought",
+        ]
+
+        for marker in markerCandidates {
+            if let range = lowercased.range(of: marker) {
```
RaajeevChandran: Using lowercased.range(of: marker) and then manually converting the offset back to sanitized via distance/offsetBy is unsafe for strings with multi byte characters like emojis. The character offsets in lowercased may not map 1:1 to sanitized which could produce an incorrect cut or a runtime crash.

a safer alternative would be (this avoids the manual index arithmetic entirely):
```
        if let range = sanitized.range(of: marker, options: [.caseInsensitive]) {
            sanitized = String(sanitized[..<range.lowerBound])
            break
        }
```

Packages/OsaurusCore/Views/Chat/ChatView.swift
```diff
@@ -347,6 +347,98 @@ final class ChatSession: ObservableObject {
...
+                    if normalizedWord == lastWord {
+                        repeatedTailCount += 1
+                    } else {
+                        break
+                    }
+                }
+
+                if repeatedTailCount >= 16 {
+                    let escapedWord = NSRegularExpression.escapedPattern(for: lastWord)
+                    let pattern = "(?i)(?:\\b\(escapedWord)\\b(?:\\s+\\b\(escapedWord)\\b){15,})\\s*$"
```
RaajeevChandran: line 427 uses repeatedTailCount >= 16 but the regex pattern at line 429 uses {15,} (15 or more). These are slightly inconsistent as {15,} will match one fewer repetition than the guard requires. Please align both to the same value

Packages/OsaurusCore/Views/Chat/ChatView.swift
```diff
@@ -1075,11 +1184,27 @@ final class ChatSession: ObservableObject {
                                 }
                                 uiDeltaCount += 1
                                 processor.receiveDelta(delta)
+                                if assistantTurn.contentLength >= 256, uiDeltaCount % 8 == 0 {
+                                    let currentContent = assistantTurn.content
+                                    if Self.containsLeakedChannelThoughtMarkers(currentContent)
+                                        || Self.isLikelyCorruptedAssistantOutput(currentContent)
+                                    {
+                                        abortedForCorruption = true
```
RaajeevChandran, line 1188: after setting abortedForCorruption to true, sanitizedCorruptedAssistantOutput is still applied unconditionally at line 1204. in the abort case, the content at that point is partially corrupted and the sanitized result may still be poor. I think it's better to fully suppress the turn (set content to nil or an empty string) rather than displaying a potentially garbled partial response

Packages/OsaurusCore/Views/Chat/ChatView.swift
```diff
@@ -347,6 +347,98 @@ final class ChatSession: ObservableObject {
         return parts.joined(separator: "\n\n")
     }
 
+    static func isLikelyCorruptedAssistantOutput(_ content: String) -> Bool {
```
RaajeevChandran, line 350: this method is running an O(n²  * 8) character array scan every 8 deltas. for a slowly degrading model output that produces thousands of tokens before triggering the threshold, this scan runs many times on ever growing content. the streaming performance could take a hit on long outputs

Packages/OsaurusCore/Tests/Chat/ChatViewSandboxTests.swift
RaajeevChandran, line 1: There's no test covering strings with emoji or multi byte characters before a channel marker or repeated tail. 
Given the index mapping bug that I already mentioned in one of the comments, a test like this would catch regressions:
let corrupted = "Hello 🎉 world.\n<channel>|<channel>thought"
let sanitized = ChatSession.sanitizedCorruptedAssistantOutput(corrupted)
#expect(sanitized == "Hello 🎉 world.")

RaajeevChandran (changes requested):

mimeding:
Rebased this branch onto current `main` and force-pushed the refreshed head `68c69b7e` from a clean Phase 0 worktree.

The current head keeps the requested fixes for actor isolation, local-model detection, corrupted-output suppression, bounded suffix scanning, and regression coverage. I reran local validation with `swift test --package-path Packages/OsaurusCore --filter ChatViewSandboxTests` and got 18 passing tests.

The remaining GitHub follow-up looks like review-thread resolution and reviewer confirmation rather than another code change.

Packages/OsaurusCore/Services/Chat/SystemPromptComposer.swift
mimeding: The current head now marks `resolveTools` as `@MainActor`, so the on-demand local-chat path no longer relies on implicit caller actor context.

Packages/OsaurusCore/Services/Chat/SystemPromptComposer.swift
mimeding: Added the explanatory heuristic comment and tightened the detection to word-boundary keywords plus explicit capability phrases, with a benign-query regression check.

Packages/OsaurusCore/Services/Chat/SystemPromptComposer.swift
mimeding: Added the explicit comment here that Work Mode always keeps the full tool set loaded, and that on-demand narrowing stays chat-only.

Packages/OsaurusCore/Views/Chat/ChatView.swift
mimeding: Swapped the manual offset conversion for a direct `range(of:options:)` search, which keeps the marker trimming safe for emoji and other multibyte text.

Packages/OsaurusCore/Views/Chat/ChatView.swift
mimeding: Aligned the repeat-count guard and regex threshold so the repeated-tail detection now uses a single minimum count consistently.

Packages/OsaurusCore/Views/Chat/ChatView.swift
mimeding: Updated the abort path so a corruption-triggered stop clears the assistant turn instead of showing a partially sanitized fragment.

Packages/OsaurusCore/Views/Chat/ChatView.swift
mimeding: Bounded the corruption scan to a trailing window, which removes the repeated full-history rescans from this path.

Packages/OsaurusCore/Tests/Chat/ChatViewSandboxTests.swift
mimeding: Added the Unicode regression coverage here so the sanitizer is exercised with emoji and other multibyte characters before the marker.

mimeding:
Maintainer follow-up on April 16, 2026: the previously open review items on this branch have been addressed on the current head (`68c69b7e`), including the actor annotation, heuristic/comment clarifications, bounded corruption scanning, corruption-abort handling, and the Unicode regression coverage. I’ve replied on the affected threads and resolved them so the remaining action here is reviewer re-check rather than more implementation work.

PR #918, opened 2026-04-22
fix(chat): skip embedding-type models when auto-selecting default

### LINKED ISSUE #884: Chat tab by default picks 1st model and doesn't check if it's an LLM
### Checks

### Describe the bug

If I add a custom server as a provider, Osaurus will pick the first model in the list of /v1/models and it doesn't check if the model is actually an LLM. In my case embeddings model was first so I was getting 500 error without any details. However, after I switched to work tab and picked the correct model it started working just fine.

### Steps to reproduce

_No response_

### Provider type

Local model (MLX)

### Model name

_No response_

### Osaurus version / commit

0.16.15

### macOS version

15.6

### Apple Silicon chip

M4 Max 128 Gb

### Xcode version

26.3

### Logs

```
Error: Request failed: HTTP 500: {"error":{"message":"Internal server error","type":"server_error","param":null,"code":null}}
```

### Screenshots

_No response_

mimeding:
## Business rationale

When a user adds a custom OpenAI-compatible server as a remote provider, Osaurus fetches the model list from `/v1/models` and auto-selects the first entry as the Chat tab's default model. The response format is a flat list of IDs with no capability metadata, so an embedding or reranker model is indistinguishable from a chat model by type. When such a model happens to be first, every subsequent chat turn fails with an opaque HTTP 500 and the user has no actionable detail — per the reporter, the workaround (switching to another tab to pick a model manually) was found only by chance.

This is a first-use trap: the worst time to hit an opaque error is right after adding a new provider, before the user trusts Osaurus's plumbing. Silently skipping obvious embedding/reranker IDs during default-selection turns a confusing failure into a working out-of-the-box experience.

### Root cause

Remote picker items are built in `ModelPickerItemCache.computeItems` from `RemoteProviderManager.cachedAvailableModels()`, which comes from `RemoteProviderService.fetchModels(from:)` — and that method returns `[String]`. There is no kind/capability field in the pipeline today.

Across `ChatView.swift`, eight call sites fall back to `pickerItems.first?.id` (or the equivalent on `newOptionIds` / `providerItems`) when no agent-preferred, saved, or otherwise-valid model exists. Any of them can land on an embedding model.

### Approach

Name-based heuristic on the model ID, applied only at default-selection time. The fuller typed-capability solution (per-provider capability probing) is a larger follow-up; this PR takes the smallest change that removes the first-use failure.

- `ModelPickerItem.isLikelyChatCapable` — `true` for `.foundation` and `.local` items (curated catalog is chat-only); for `.remote`, delegates to…
- `ModelPickerItem.isLikelyEmbeddingOrRerankerID(_:)` — whole-token match on the ID (after stripping any `provider/` prefix) for `embed`, `embedding`, `embeddings`, `reranker`, `rerank`, `colbert`; plus `bge-` / `nomic-embed-` family prefixes.
- `[ModelPickerItem].firstChatCapable` — `first(where: { $0.isLikelyChatCapable }) ?? first`, so if every item trips the heuristic (e.g. a provider that exposes only reranker models) the picker is never left unset.

Call sites in `ChatView.swift` that previously used `.first` for default selection now use `.firstChatCapable`. Specific-model lookups (`pickerItems.first(where: { \$0.id == model })`) are unchanged — those are not default-selection.

### Known limits (deliberate)

- Heuristic is name-only. A chat model whose ID happens to contain a flagged token would be wrongly skipped — the `?? first` fallback protects the common case where every item in the list trips the heuristic, but a mixed list with an unusually-named chat model hidden after a correctly-named embedding model would still end up on the embedding. Acceptable trade: the common shape is "embedding-first, chat-second" and this PR fixes that. A non-trivial fraction of the remaining tail requires typed metadata.
- The `e5-` family prefix is intentionally NOT in the family-prefix list, even though `e5-small` / `e5-base` / `e5-large` are embeddings — because `e5-mistral-7b-instruct` is a chat model and the token `e5` alone is too generic to gate without a false-positive regression.
- The correct long-term fix is typed capability metadata on remote models (per-provider probing; OpenAI's `/v1/models/{id}` exposes no `kind`, some Ollama tags imply it, Gemini's `supportedGenerationMethods` carries it). Out of scope for this PR.

## What changed and why

- `Packages/OsaurusCore/Models/Configuration/ModelPickerItem.swift` — add `isLikelyChatCapable`, the static `isLikelyEmbeddingOrRerankerID(_:)` classifier, and `[ModelPickerItem].firstChatCapable`. New code, no existing behavior changed.
- `Packages/OsaurusCore/Views/Chat/ChatView.swift` — 8 default-selection call sites switch from `.first` to `.firstChatCapable` (lines 297, 322, 576, 674, 2050, 2135, 2213, 2230).
- `Packages/OsaurusCore/Tests/Model/ModelPickerItemChatCapabilityTests.swift` — new file; 15 tests covering known embedding/reranker IDs (OpenAI, Nomic, Jina, BGE, Colbert), known chat IDs (GPT, Claude, Qwen, Llama, Mistral, Gemini), provider-prefix stripping, substring-vs-token matching, `.foundation` / `.local` bypass, and `firstChatCapable` fallback behavior including the empty-array case.

## Validation

\`\`\`
# Build complete!

swift test --package-path Packages/OsaurusCore --filter ModelPickerItemChatCapabilityTests
# Test run with 15 tests in 1 suite passed
\`\`\`

Manual repro of #884: added a custom remote provider whose `/v1/models` returns an embedding model first; before this change the Chat tab auto-selected it and chat turns failed with HTTP 500. With this change, the picker skips to the first chat-capable entry and the chat works.

Closes #884.

### DIFF
--- a/Packages/OsaurusCore/Models/Configuration/ModelPickerItem.swift
+++ b/Packages/OsaurusCore/Models/Configuration/ModelPickerItem.swift
@@ -141,9 +141,82 @@ extension ModelPickerItem {
     }
 }
 
+// MARK: - Default-selection capability heuristic
+
+extension ModelPickerItem {
+    /// Heuristic used only for default-selection: is this item plausibly a
+    /// chat-capable model?
+    ///
+    /// Remote providers expose `/v1/models` as a flat list of IDs with no
+    /// capability metadata, so an embedding or reranker model is
+    /// indistinguishable by type from a chat model. When such a model happens
+    /// to be first in the list, the Chat tab previously auto-selected it and
+    /// every message failed with an opaque HTTP 500. This check lets the
+    /// default-selection step skip obvious non-chat IDs while remaining
+    /// conservative: if a chat model has an unusual name that trips the
+    /// heuristic, the array helper below falls back to the first item so the
+    /// picker is never left empty when models exist.
+    var isLikelyChatCapable: Bool {
+        switch source {
+        case .foundation, .local:
+            // Foundation is Apple's on-device chat model; `.local` items come
+            // from the curated MLX catalog, which is chat-only.
+            return true
+        case .remote:
+            return !Self.isLikelyEmbeddingOrRerankerID(id)
+        }
+    }
+
+    /// Token- and prefix-based classifier that returns `true` when the model
+    /// ID almost certainly belongs to an embedding or reranker family.
+    ///
+    /// Matching is word-boundary so "embedded" in a chat model's description
+    /// would not trigger (though only the ID is inspected). A provider prefix
+    /// like `"provider-name/model-id"` is stripped before matching.
+    static func isLikelyEmbeddingOrRerankerID(_ id: String) -> Bool {
+        // Strip any `"provider/"` prefix added by `fromRemoteModel`.
+        let tail = id.split(separator: "/").last.map(String.init) ?? id
+        let lower = tail.lowercased()
+
+        // Whole-token match on non-alphanumerics so we catch, e.g.,
+        // `text-embedding-ada-002`, `nomic-embed-text`, `bge-reranker-v2-m3`
+        // without misfiring on substrings like `embedded` or `rerankable`.
+        let tokens = lower.split(whereSeparator: { !$0.isLetter && !$0.isNumber }).map(String.init)
+        for token in tokens {
+            switch token {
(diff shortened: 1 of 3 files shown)


tpae (approved):
This looks good to me. Thank you for your contributions!

PR #921, opened 2026-04-23
fix(icon): pad AppIcon to match macOS template grid

### LINKED ISSUE #676: Dock icon is too large
### Checks

### Describe the bug

Sorry, not a bug as such, but the dock icon is very large compared to other icons. Maybe intentional coz this is a large dinosaur? 🦕 Any way to reduce the size? 

Here's what it looks like in my dock. This is from version 0.14.22. 

Image

### Steps to reproduce

_No response_

### Provider type

Not applicable

### Model name

_No response_

### Osaurus version / commit

0.14.22

### macOS version

15.7.5 (Sequoia)

### Apple Silicon chip

_No response_

### Xcode version

_No response_

### Logs

```shell

```

### Screenshots

_No response_

mimeding:
## Business rationale

The Dock icon visibly renders larger than neighboring system apps. Issue reporter noted this against 0.14.22 and it persists on 0.17.2 — every macOS user who keeps Osaurus in the Dock sees the size mismatch. It's a quick visual fix that removes a first-impression-of-polish issue, especially relevant now that 0.17.x is consolidating the app into a single Chat/Agent surface.

### Root cause

Every PNG in `AppIcon.appiconset` rendered the brand squircle edge-to-edge on its canvas (e.g. the 1024×1024 PNG is 1024×1024 of filled squircle, no padding). macOS scales the PNG to fit the dock slot verbatim, so Osaurus's squircle ends up filling the slot entirely.

Apple's macOS app-icon template reserves about 12% of the canvas as transparent padding around a ~88%-scale rounded-square. System apps (Finder, Safari, Messages, Notes) all follow this grid, which is why they render at a consistent visual size. An app that ships a full-bleed squircle looks "bigger" even when the underlying canvas dimensions are identical.

### Approach

Regenerate each mac slot from the `icon_1024.png` master:
- Scale the master by `0.88` using Lanczos resampling.
- Center the scaled result on a transparent canvas at the slot's exact physical pixel size.
- Write to the existing filename (no `Contents.json` change required — slot names and dimensions are unchanged).

The script used to regenerate the icons is kept out of the repo (it's a one-shot tooling step, not a checked-in asset), but for reviewers who want to reproduce or adjust the scale:

```python
from PIL import Image
SCALE = 0.88
master = Image.open("icon_1024.png").convert("RGBA")
for name, size in [
    ("icon_16.png", 16), ("icon_32 1.png", 32), ("icon_32.png", 32),
    ("icon_64.png", 64), ("icon_128.png", 128), ("icon_256 1.png", 256),
    ("icon_256.png", 256), ("icon_512 1.png", 512), ("icon_512.png", 512),
    ("icon_1024.png", 1024),
]:
    content = int(round(size * SCALE))
    scaled = master.resize((content, content), Image.Resampling.LANCZOS)
    canvas = Image.new("RGBA", (size, size), (0, 0, 0, 0))
    canvas.paste(scaled, ((size - content) // 2, (size - content) // 2), scaled)
    canvas.save(name, format="PNG", optimize=True)
```

## What changed and why

- `App/osaurus/Assets.xcassets/AppIcon.appiconset/icon_{16,32,64,128,256,512,1024}.png` (plus the Xcode-duplicated `" 1"` variants) — regenerated so each slot's content squircle is sized to 88% of its physical pixel dimensions, centered with transparent padding.
- `Contents.json` — unchanged.
- Non-mac (iOS) slots — unchanged (they don't have filenames attached in `Contents.json` anyway).

## Validation

- Visual verification that the generated PNGs carry transparent padding around the squircle (confirmed on the 1024 master; the downscaled slots follow by construction).
- A maintainer with a local build should verify the in-dock appearance against another system app (Finder, Notes, Safari) to confirm the size now matches. A single `xcodebuild build` + launch is sufficient; no Swift code is touched so no tests to run.

Closes #676.


mimeding:
The `test-core` failure here was the cold SwiftPM cache (`CNIOLLHTTP` / `CNIOExtrasZlib` module resolution), not the icon change. Since #925 just landed a green `test-core` run (6m1s) the DerivedData cache is now warm — a maintainer rerun of this job should warm-start and pass. I don't have rerun permission on the repo itself. Thanks.

mimeding:
@tpae small cosmetic fix ready whenever you have time — pads `AppIcon` to match the macOS template grid so the dock icon has its normal safe-area margins instead of bleeding to the corners. Single-asset change, low risk. test-core is green on the current revision.

tpae:
@mimeding thank you for this PR, but we have new icons coming shortly (early next week) I would rather do a new icon release. I've shared your specs with our designer, he will consider it to have new ones added.

PR #923, opened 2026-04-23
fix(models): refuse model download when disk lacks space to finish

### LINKED ISSUE #580: Bug Report: Storage Exhaustion Leading to UI/CLI Desync
### Checks

### Describe the bug

Issue: When downloading two models simultaneously, my storage ran out of space, causing both downloads to fail. Currently, the models do not appear in the UI, but they are visible via the CLI. Even after manually deleting the model files, re-downloading the same models results in them appearing in the terminal but remaining invisible in the UI.

Observed Result: The models are recognized by the terminal/CLI but do not populate or sync with the User Interface.

Expected Result: The UI should correctly reflect the status of the models, and a fresh download should fully register the models in both the CLI and UI environments.

### Steps to reproduce

1. Initiate the download of two models at the same time.
2. Trigger a "Storage Full" error during the process.
3. Observe that the UI shows no models, while the CLI lists them.
4. Manually delete the model files via CLI.
5. Re-download the same models.

### Osaurus version / commit

Version 0.13.4 (0.13.4)

### macOS version

Tahoe 26.3 (25D125)

### Apple Silicon chip

M1

### Xcode version

_No response_

### Logs

```shell

```

### Screenshots

_No response_

mimeding:
## Business rationale

Issue #580 describes a cascade: two simultaneous model downloads fail with an out-of-space error, the UI then shows no models (while the CLI still lists them), and the inconsistency persists across manual delete-and-redownload. Users who hit this path once land in a "the app is broken" state with no actionable diagnostic.

The primary trigger is the failed download itself — nothing today stops a download from starting when the target volume cannot possibly hold the result. Catching this at the preflight turns a silent-data-corruption-adjacent failure into a clear, actionable error before any bytes hit disk, which is the cheapest place to fix the class of bug.

The issue is labeled `good first issue`, suggesting a scoped contributor fix is expected; this PR is that scope.

### Root cause (for the preflight portion)

`ModelDownloadService.download(_:)` resolves the Hugging Face file list, sums its byte count into `totalBytes`, and starts writing immediately. There is no disk-space check. When the volume fills mid-download, URLSession's atomic-move fails and the service sets `.failed`, but by then partial files are on disk and tracked state can drift from what the UI shows.

### Approach

Refuse to start the download if the target volume can't hold what's left to fetch:

1. **Compute remaining bytes** as `totalBytes - completedFileBytes` — i.e. the actual new bytes we'd write, accounting for files already fully present on disk from an earlier partial attempt.
2. **Query the correct volume.** The models directory can be redirected to an external drive via `DirectoryPickerService`, so capacity is queried on `model.localDirectory` rather than assuming the boot volume. Preferred key is `URLResourceKey.volumeAvailableCapacityForImportantUsageKey` (Int64, excludes system reserve), with a fallback to `FileManager.attributesOfFileSystem(forPath:)` + `.systemFreeSize` when the key isn't available.
3. **Fail open on query failure.** If both queries fail, the preflight returns `nil` and the download proceeds — a stale or missing estimate is worse than none; the ordinary write-path error handling still catches the failure.
4. **Safety margin.** `256 MB` is subtracted from the free-space estimate before the comparison, because Hugging Face LFS pointers can slightly under-report file size and the OS needs some headroom for the atomic-rename at the tail of each file. Refusing at the edge avoids "ran out of space on the last megabyte" failures.

The comparison itself is extracted into a pure `storageRefusalMessage(neededBytes:freeBytes:)` helper so the threshold arithmetic can be unit-tested without filesystem mocking.

### Known limits (deliberate, noted for reviewers)

- **Partial-file cleanup is NOT part of this PR.** When a download fails mid-flight for any other reason (network drop, user cancel that races the writer), orphaned partial files stay on disk. A follow-up PR should add cleanup in `DirectDownloader`'s error path; that's a separate concern with its own testing surface.
- **CLI/UI source-of-truth alignment is NOT part of this PR.** Issue #580 also reports that the CLI sees models the UI does not. The CLI queries the running server's `/models` endpoint (`Packages/OsaurusCLI/.../List.swift`), while the UI scans the models directory via `ModelManager.discoverLocalModels()`. Reconciling the two warrants its own design — they can legitimately diverge briefly during downloads — and is out of scope here.

## What changed and why

- `Packages/OsaurusCore/Services/ModelDownloadService.swift`
  - Added preflight block immediately after `totalBytes` is computed and `filesToDownload` is determined, before the first byte is written. Refuses with `.failed(error: …)` and `clearDownloadTracking(for:)` when the helper returns a message.
  - Added three new static helpers: `storageSafetyMarginBytes` (256 MB constant), `storageRefusalMessage(neededBytes:freeBytes:)` (pure comparison), and `freeBytesOnVolume(containing:)` (the URL-resource-value query with fallback).
- `Packages/OsaurusCore/Tests/Service/ModelDownloadServiceStorageTests.swift` (new)
  - Six tests covering the refusal-threshold arithmetic: refuses when needed exceeds free, allows when free comfortably exceeds needed, enforces the safety margin exactly at the boundary, zero needed always proceeds, refusal message is user-facing (ByteCountFormatter-formatted, not raw digits), and a sanity check that `freeBytesOnVolume` returns a positive value on `NSTemporaryDirectory()`.

## Validation

\`\`\`
# Build complete!

swift test --package-path Packages/OsaurusCore --filter ModelDownloadServiceStorageTests
# Test run with 6 tests in 1 suite passed
\`\`\`

Closes #580.

### DIFF
--- a/Packages/OsaurusCore/Services/ModelDownloadService.swift
+++ b/Packages/OsaurusCore/Services/ModelDownloadService.swift
@@ -165,6 +165,29 @@ final class ModelDownloadService: ObservableObject {
                     }
                 }
 
+                // Preflight disk-space check. Runs on the filesystem that hosts
+                // `model.localDirectory` — which may be an external drive when
+                // the user has pointed `DirectoryPickerService` at one — so we
+                // can't assume boot-volume capacity. If the query itself fails
+                // we proceed with the download; a stale estimate is worse than
+                // none, and the ordinary write-path error handling still fires.
+                let bytesToDownload = totalBytes - completedFileBytes
+                if bytesToDownload > 0,
+                    let freeBytes = Self.freeBytesOnVolume(containing: model.localDirectory),
+                    let refusal = Self.storageRefusalMessage(
+                        neededBytes: bytesToDownload,
+                        freeBytes: freeBytes
+                    )
+                {
+                    await MainActor.run {
+                        if self.downloadTokens[model.id] == token {
+                            self.downloadStates[model.id] = .failed(error: refusal)
+                            self.clearDownloadTracking(for: model.id)
+                        }
+                    }
+                    return
+                }
+
                 await MainActor.run {
                     guard self.downloadTokens[model.id] == token else { return }
                     let fraction = totalBytes > 0 ? Double(completedFileBytes) / Double(totalBytes) : 0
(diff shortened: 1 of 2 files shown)


tpae (approved):
Thank you for your contribution! This looks good.

PR #890, opened 2026-04-18
P0B: Harden import/export trust boundaries for attachments and shared artifacts

mimeding:
## Business Rationale

This PR closes the highest-confidence import/export trust-boundary defects before broader document workflow work expands the attack surface. It reduces prompt corruption risk, artifact path escape risk, and oversized-file intake risk in the current product surface.

## Engineering Rationale

The current architecture still relies on prompt-injected attachment wrappers and model/tool-provided artifact metadata. Hardening those boundaries now is cheaper and safer than retrofitting them after registry and format work land. This PR keeps the fix narrow and compatible with the parallel registry track.

## Exact Scope

- escape and basename-normalize chat attachment wrapper filenames and content
- canonicalize and contain artifact source and destination paths
- reject oversized document files before full parse
- add targeted regression tests for wrapper escaping, traversal rejection, and oversized-file handling

## Non-Scope

- no new format support
- no registry ownership changes
- no export/validate runtime wiring changes
- no broader runtime or feature refactors

## Validation

- `swift test --package-path Packages/OsaurusCore --filter ChatAttachmentSecurityTests`
- `swift test --package-path Packages/OsaurusCore --filter SharedArtifactSecurityTests`
- `swift test --package-path Packages/OsaurusCore --filter DocumentParserSecurityTests`
- `swift build --package-path Packages/OsaurusCore`

## Security / Injection Checks

- attachment filenames with quotes, slashes, or closing-tag fragments are basename-normalized and escaped
- attachment bodies containing `</attached_document>` or tool-like markup remain escaped inside the wrapper
- host-folder traversal attempts like `../outside.txt` are rejected
- sandbox traversal attempts like `/workspace/agents/<agent>/../../../outside.txt` are rejected
- destination traversal like `../exports/../../quarterly.md` is reduced to a contained basename
- files over 10 MB are rejected before parse

## Residual Risks

- decompression or package-bomb defenses remain future work
- sandbox artifact lookup still uses heuristic basename fallback, though containment is enforced
- escaped wrapper content is safer but not byte-identical inside the prompt envelope

## Planning Artifact Publication

- local files under `results/osaurus-plan/` remain unpublished planning artifacts and are not part of this PR
- PR rationale is paraphrased from local planning notes rather than publishing those notes directly

PR #925, opened 2026-04-23
fix(security): harden import/export trust boundaries for attachments and shared artifacts

mimeding:
## Business rationale

Chat attachments and `share_artifact` tool results are the two paths where agent-controlled bytes cross into either the user's prompt envelope or the host filesystem. Today both paths forward those bytes with minimal validation, so a single hostile document attached to a chat, or a single malformed `share_artifact` call, can forge a closing `</attached_document>` wrapper, inject bracketed pseudo-tool markers into the prompt, or write an artifact outside its per-context directory. Closing these gaps now — before the broader document-workflow expansion lands — is cheaper and safer than retrofitting them afterwards.

## Coding rationale

This PR replaces #890, which targeted the pre-#893 `Models/Work/WorkModels.swift` layout. After Work Mode was collapsed into the single Chat/Agent system, the hardening belongs on three surfaces that still exist on `main`:

- `ChatSession.buildUserMessageText` in `Packages/OsaurusCore/Views/Chat/ChatView.swift` — previously interpolated the attachment filename and body directly into the wrapper. The fix reduces the filename to its basename and XML-entity-escapes both the name and the body.
- `SharedArtifact.processToolResult` in `Packages/OsaurusCore/Models/Chat/SharedArtifact.swift` — previously built `destPath = contextDir.appendingPathComponent(parsed.filename)` without sanitization or containment. The fix sanitizes `parsed.filename` to a contained basename (re-written into the metadata so downstream consumers also see the safe name) and resolves the destination against a canonicalized `contextDir`, rejecting anything that escapes it.
- `SharedArtifact.resolveSourcePath` — previously trusted the agent-provided path inside each execution mode. The fix runs every branch through a canonicalize-and-contain helper and drops the old raw `appendingPathComponent` usage. The basename-fallback heuristic is preserved but confined to `agentDir`.
- `DocumentParser.parseAll` in `Packages/OsaurusCore/Utils/DocumentParser.swift` — the existing `maxParsedTextLength` trim only bounds the *decoded* text, so a crafted ~hundreds-of-MB latin-1 or rich-document input can still OOM the decode pass. A 10 MiB pre-read cap mirrors the image cap in `FloatingInputCard` and aborts before allocation.

The fix is deliberately narrow — no new format support, no registry changes, no rename churn. Compatible with the in-flight document-workflow track.

## What changed and why

- `Packages/OsaurusCore/Views/Chat/ChatView.swift`: add `escapeAttachmentName` and `xmlEscape` private helpers on `ChatSession`; route both the wrapper name and body through them in `buildUserMessageText` — a hostile document can no longer forge wrapper tags or smuggle pseudo-tool markers into the prompt.
- `Packages/OsaurusCore/Models/Chat/SharedArtifact.swift`: make `ParsedMarkers.filename` mutable; sanitize the filename at intake and re-publish it into metadata; replace `destPath` construction with `resolveDestinationPath`; rebuild `resolveSourcePath` around `resolveContainedPath` for the sandbox, workspace, host-folder, and basename-fallback branches — artifacts can no longer be written or read outside their trusted root, regardless of execution mode.
- `Packages/OsaurusCore/Utils/DocumentParser.swift`: add `maxFileSize = 10 * 1024 * 1024` and short-circuit `parseAll` before the read when the file exceeds it — documents can no longer OOM the decode pass.
- `Packages/OsaurusCore/Tests/Chat/ChatAttachmentSecurityTests.swift`: pins wrapper escaping, empty-name fallback, and the no-attachment passthrough.
- `Packages/OsaurusCore/Tests/Chat/SharedArtifactSecurityTests.swift`: pins destination containment, `../../../etc/passwd`-style filename sanitization, the `..`-only fallback name, and host-folder source traversal rejection. Restores `OsaurusPaths.overrideRoot` in teardown per the PR template checklist.
- `Packages/OsaurusCore/Tests/Service/DocumentParserSecurityTests.swift`: pins the oversize rejection and the at-cap acceptance.

## Validation

- `swift test --filter 'ChatAttachmentSecurity|SharedArtifactSecurity|DocumentParserSecurity'` — 8 of 8 passing.
- `swift test` (full `OsaurusCore` package) — 891 of 891 passing.
- `xcrun swift-format lint --strict` on every touched file — clean.
- `swiftlint` on every touched file — no new violations introduced (pre-existing warnings untouched).

## Security / injection checks

- `<attached_document>` wrapper payloads containing `</attached_document>`, `<tool …>`, or `<system>…</system>` are entity-escaped and can no longer reach the model as control text.
- Attachment filenames with quotes, angle brackets, or path separators are basename-normalized and escaped before interpolation.
- Share-artifact filenames like `../../../etc/passwd` are rewritten to `passwd` and land inside the context dir; filenames that collapse to `.` or `..` fall back to `artifact`.
- `hostFolder` source paths like `../outside.txt` are refused rather than silently escaping the user-picked folder.
- `/workspace/` and `containerHome`-prefixed sandbox paths are canonicalized and confined to their trusted root; the basename-fallback is confined to `agentDir/<output|out|build|dist>`.
- Documents larger than 10 MiB are rejected before the rich-document decode pass.

## Residual risks

- Decompression / package-bomb defenses remain future work.
- Sandbox artifact lookup still relies on a heuristic basename fallback inside `agentDir`; containment is now enforced but semantic correctness for ambiguous basenames isn't.
- The wrapper body is XML-escaped rather than byte-identical — bytes still round-trip conceptually but the on-the-wire prompt representation is not the raw document. Consistent with how PDFs, DOCX, etc. are already summarized into the envelope.

Replaces #890.

### DIFF
--- a/Packages/OsaurusCore/Models/Chat/SharedArtifact.swift
+++ b/Packages/OsaurusCore/Models/Chat/SharedArtifact.swift
@@ -160,7 +160,7 @@ extension SharedArtifact {
     /// Raw parsed content extracted from the marker-delimited region.
     struct ParsedMarkers {
         var metadata: [String: Any]
-        let filename: String
+        var filename: String
         let contentLines: [String]
         let startRange: Range<String.Index>
         let endRange: Range<String.Index>
@@ -218,9 +218,25 @@ extension SharedArtifact {
         let hasContent = parsed.metadata["has_content"] as? Bool ?? false
         let path = parsed.metadata["path"] as? String
 
+        // Strip any path segments the agent may have smuggled into the filename
+        // (e.g. `../quarterly.md`) before we resolve it against the context dir.
+        let sanitizedFilename = sanitizeArtifactFilename(parsed.filename)
+        if sanitizedFilename != parsed.filename {
+            NSLog(
+                "[SharedArtifact] Sanitized artifact filename '%@' → '%@'",
+                parsed.filename,
+                sanitizedFilename
+            )
+        }
+        parsed.filename = sanitizedFilename
+        parsed.metadata["filename"] = sanitizedFilename
+
         let contextDir = OsaurusPaths.contextArtifactsDir(contextId: contextId)
         OsaurusPaths.ensureExistsSilent(contextDir)
-        let destPath = contextDir.appendingPathComponent(parsed.filename)
+        guard let destPath = resolveDestinationPath(filename: parsed.filename, contextDir: contextDir) else {
+            NSLog("[SharedArtifact] Refused destination path for filename '%@'", parsed.filename)
+            return nil
+        }
 
         let artifact: SharedArtifact
         let contentLines: [String]
@@ -386,6 +402,9 @@ extension SharedArtifact {
 
     /// Maps an agent-provided path to the host-side URL, normalizing absolute
     /// in-container paths, `./` prefixes, and falling back to a basename search.
+    /// Every returned URL is canonicalized and verified to live inside the
+    /// caller's trusted root — a crafted `../` path cannot escape the sandbox
+    /// agent dir, the container workspace, or the user-picked host folder.
     private static func resolveSourcePath(
         _ path: String,
         executionMode: ExecutionMode,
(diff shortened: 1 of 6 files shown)


mimeding:
@tpae ready for review when you have time. This re-works the hardening I had in the earlier #890 against current main (Work Mode removal in #893 moved the target files). Three narrow surfaces:

- Attachment wrapper in `ChatSession.buildUserMessageText` — XML-entity-escape name + body so a hostile document can't forge `</attached_document>` or inject pseudo-tool markers.
- `SharedArtifact.processToolResult` — sanitize filename to a contained basename, canonicalize + contain the destination + every `resolveSourcePath` branch.
- `DocumentParser.parseAll` — 10 MiB pre-read cap so a crafted rich document can't OOM the decode pass.

8 security regression tests, full suite 891/891 green. Low risk for the attached-document path since the only behaviour change is escaping + containment; the LLM-visible text in the wrapper stays byte-for-byte the same for benign input.

tpae (approved):

PR #926, opened 2026-04-23
feat(documents): foundations for document format adapters

mimeding:
## Business rationale

Osaurus today handles attached files through a flat ingestion path that turns rich content into plain text. That is enough for some quick answers, but it does not give Osaurus a clean boundary between minimal Core parsing and richer document workflows implemented by plugins.

This PR adds only the Core substrate for that boundary: document adapter contracts, a typed parse result with text fallback, common size/error limits, and a process-wide registry. It does not add high-fidelity format implementations, editing capability, generated output, or domain-specific behavior.

The review direction is that Core should stay small: document parsing for LLM quick answers belongs in Core; editing, generation, and high-fidelity format or domain workflows belong in plugins such as XLSX, PPTX, vision, finance, engineering, and similar specialty packages. This PR supports that split by giving plugins stable registration points instead of forcing every format into Osaurus Core.

## Core / plugin boundary

Core owns only the minimal functionality needed for safe quick-answer ingestion:

- Stable parser/adapter contracts.
- Typed parse result and text fallback.
- Shared size limits and error types.
- Registry/routing for Core and plugin-provided parsers.
- Compatibility with the existing attachment read path.

Plugins own everything beyond that minimal ingestion layer:

- Editing and generated output.
- XLSX/PPTX workflows.
- OCR, vision, and layout-heavy extraction.
- High-fidelity spreadsheet, PDF, CAD, GIS, geoscience, finance, marketing, statistics, and other domain parsing.
- Domain-specific agent tools, dashboards, reports, and writeback.
- Large optional dependencies and format-specific UX.

The `DocumentFormatEmitter` protocol is included only as a neutral boundary contract so plugin-provided emitters can have a typed interface later. It is not a Core commitment to ship editing or emission implementations. If maintainers prefer, emitter support can remain unused until the plugin bridge PRs land.

## Coding rationale

Three things had to be decided up front:

- **Protocols are split by responsibility.** `DocumentFormatAdapter` reads, `DocumentFormatEmitter` writes, `DocumentFormatStreamer` streams. Splitting them keeps Core ingestion small while allowing plugins to opt into richer behavior without forcing stub writers or streamers onto every format.
- **The registry is lock-based, not actor-isolated.** The agent tool surface can run off the main actor and should not pay an `await` hop per file lookup. An `NSLock` on an `@unchecked Sendable` final class keeps lookup cheap and lets tests run concurrent registrations without plumbing Task isolation into the fixtures.
[...]

PR #927, opened 2026-04-23
feat(documents): wrap PlainText/PDF/DOCX as adapters and route DocumentParser through the registry

mimeding:
Rebased onto `origin/main` after #1015. The diff is now the adapter wrappers plus the `DocumentParser` registry shim.

## Business rationale

Osaurus today ingests every attached file through a single `DocumentParser` switch. To expand file-format fidelity (XLSX cells with formulas, DOCX tables, PDF tables, MT940/OFX bank exports), that switch has to move onto a pluggable adapter surface — and it has to do so without a user-visible behaviour change, because chat attachments are a core workflow. This PR is the bridge: it migrates the three file types already handled (plain text, PDF, DOCX-family) onto the adapter surface from #926 and proves the `DocumentParser` shim routes correctly. It deliberately adds zero fidelity; it's the groundwork every later adapter plugs into.

## Coding rationale

Three design decisions worth calling out:

- **`DocumentParser.parseAll` stays synchronous.** The adapter protocol is async (`parse(url:sizeLimit:) async throws -> StructuredDocument`) because streaming adapters land in the next slice. Both existing call sites (`FloatingInputCard`, `ClipboardService`) already wrap `parseAll` in `Task.detached`, so the sync-to-async bridge inside the shim (a dispatch-semaphore + detached inner Task) blocks a cooperative thread rather than the main thread. Making `parseAll` itself async would touch every ingress call site and is out of scope for this PR.
- **The PDF image-fallback stays in the legacy switch.** `PDFAdapter` throws `DocumentAdapterError.emptyContent` when a PDF has no text layer; the shim catches that specific error and falls through to `parsePDFWithFallback`, which still renders each page as PNG. Moving the image path onto the adapter surface needs a typed `PDFDocument` representation that isn't in scope until a later fidelity slice.
- **One `RichDocumentAdapter` covers DOCX, DOC, RTF, RTFD, HTML.** They share the same `NSAttributedString(url:options:documentAttributes:)` call today, so a single adapter keeps the migration minimal. The high-fidelity DOCX reader can split this along format lines later; at that point this adapter becomes the RTF/HTML-only path.

## What changed and why

- `Packages/OsaurusCore/Services/Documents/PlainTextAdapter.swift`: wraps the existing UTF-8-first / ISO-Latin-1-retry decoding path from `DocumentParser.parsePlainText`, including the 500K-character truncation marker, so the legacy behaviour stays byte-identical.
- `Packages/OsaurusCore/Services/Documents/PDFAdapter.swift`: wraps `PDFDocument.page.string` extraction; throws `.emptyContent` for text-layerless PDFs so the shim can fall through.
- `Packages/OsaurusCore/Services/Documents/RichDocumentAdapter.swift`: wraps `NSAttributedString(url:options:documentAttributes:)` across the rich-document formats; one adapter today, splits later.
- `Packages/OsaurusCore/Managers/Documents/DocumentAdaptersBootstrap.swift`: idempotent registration entry point.
- `Packages/OsaurusCore/Models/Documents/PlainTextRepresentation.swift`: neutral representation for adapters that don't yet publish a format-native shape.
- `Packages/OsaurusCore/Utils/DocumentParser.swift`: `parseAll` now tries the registry first, falls through to the existing switch on `.emptyContent` / `.unsupportedFormat`, and translates other `DocumentAdapterError` cases into the legacy `ParseError` surface. The translation is conservative — `sizeLimitExceeded` becomes `fileTooLarge`, `readFailed` preserves the underlying reason, anything else becomes a generic read error.
- `Packages/OsaurusCore/AppDelegate.swift`: calls `DocumentAdaptersBootstrap.registerBuiltIns()` near the start of `applicationDidFinishLaunching` so the shim sees adapters on the first file ingress.
- `Packages/OsaurusCore/Tests/Documents/*`: unit tests for each adapter (text decoding, size-limit refusal, empty-content handling, PDF text extraction, PDF without text layer, RTF/HTML extraction) plus integration tests for the shim (routing through a registered fixture adapter, fall-through on empty content, legacy path still works with no adapter registered, bootstrap registers all three built-ins on an isolated registry).

## Validation

- `swift build --package-path Packages/OsaurusCore` — passed.
- `swift build --package-path Packages/OsaurusCore -c release` — passed.
- `swift test --package-path Packages/OsaurusCore` — passed: 1,454 tests in 196 suites. Sandbox integration skipped by its documented env gate.
- `xcrun swift-format lint --strict` on every touched Swift file — passed.
- `swiftlint lint --strict` on every touched Swift file — passed.
- `git diff --check` — passed.
- `Packages/OsaurusCLI` gate skipped; this PR does not touch CLI code.

## Non-scope

- No new format support. CSV is already plain-text today; it stays that way until the streaming CSV/TSV adapter slice. XLSX is the first real fidelity win.
- No changes to `Attachment.Kind`. Chat attachments still decode as `.document(filename, content, fileSize)`; the new `.structuredDocument` case lands after several real adapters prove the model.
- No plugin host changes. The plugin host API that lets external plugins register their own adapters is a later slice.
- No size-limit UI. The per-format ceilings in `DocumentLimits` are internal defaults; a user-facing "Maximum attached document size" preference is deferred.

## Residual risks

- The sync-to-async bridge in `routeThroughRegistry` relies on `parseAll` always being called from a task context, not directly from the main actor. Both existing call sites satisfy this. A comment documents the invariant.
- `DocumentAdaptersBootstrap.didRegisterShared` uses the same `nonisolated(unsafe) + NSLock` pattern as `OsaurusPaths.overrideRoot`; a future migration to an actor or `OSAllocatedUnfairLock` would be cleaner but isn't in scope.
- The `runBlocking` helper allocates a detached `Task` per `parseAll` call. That's one task allocation per file ingress, which is negligible compared to the PDFKit / `NSAttributedString` work it wraps.

### DIFF
--- a/Packages/OsaurusCore/AppDelegate.swift
+++ b/Packages/OsaurusCore/AppDelegate.swift
@@ -179,7 +179,7 @@ public final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelega
         #endif
 
         // Initialize directory access early so security-scoped bookmark is active
-        let _ = DirectoryPickerService.shared
+        _ = DirectoryPickerService.shared
 
         if LaunchGuard.isSafeMode {
             NotificationService.shared.postSafeModeActive()
@@ -880,7 +880,7 @@ extension AppDelegate {
     }
 
     @objc private func handleServeCommand(_ note: Notification) {
-        var desiredPort: Int? = nil
+        var desiredPort: Int?
         var exposeFlag: Bool = false
         if let ui = note.userInfo {
             if let p = ui["port"] as? Int {
--- a/Packages/OsaurusCore/AppDelegate.swift
+++ b/Packages/OsaurusCore/AppDelegate.swift
@@ -32,6 +32,10 @@ public final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelega
         // the specific crash class this prevents.
         MLXErrorRecovery.installGlobalHandler()
 
+        // Register in-tree document format adapters before any file-ingress
+        // path can run. Idempotent; safe if a future migration moves this.
+        DocumentAdaptersBootstrap.registerBuiltIns()
+
         // Detect repeated startup crashes and enter safe mode if needed
         LaunchGuard.checkOnLaunch()
 
--- a/Packages/OsaurusCore/Managers/Documents/DocumentAdaptersBootstrap.swift
+++ b/Packages/OsaurusCore/Managers/Documents/DocumentAdaptersBootstrap.swift
@@ -1,0 +1,34 @@
+//
+//  DocumentAdaptersBootstrap.swift
+//  osaurus
+//
+//  Registers the in-tree document adapters with `DocumentFormatRegistry.shared`
+//  exactly once, at app launch. Kept separate from `AppDelegate` so tests can
+//  opt into the same registration (or opt out of it entirely) without dragging
+//  in `NSApplication`.
+//
+
+import Foundation
+
+public enum DocumentAdaptersBootstrap {
+    private static let lock = NSLock()
+    // Guarded by `lock`; the `nonisolated(unsafe)` matches the project pattern
+    // for lock-protected process-global state (see `OsaurusPaths.overrideRoot`).
+    nonisolated(unsafe) private static var didRegisterShared = false
+
+    /// Idempotent against the shared registry: safe to call from multiple
+    /// launch paths without producing duplicate adapter registrations.
+    /// Non-shared registries (tests, isolated instances) are re-registered on
+    /// every call so each test gets a clean baseline.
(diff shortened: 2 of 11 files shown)


mimeding:
Recovery-plan note: this branch is still the base of the document stack. The current `test-core` failure maps to `slowToolReturnsTimeoutEnvelopeBeforeBudgetExpires`, and I opened #976 as the narrow fix for the timeout race in `ToolRegistry.runToolBody`.

Recommended order: land #975 first for the shared CI cache failure, then land #976, then rerun/rebase #927 before continuing the document stack (#929 -> #936 -> #937 -> #939 -> #940 -> #941 -> #942).

tpae (approved):

PR #928, opened 2026-04-23
fix(provider): send max_tokens to non-OpenAI-reasoning endpoints

### LINKED ISSUE #556: mistral api returns 422
### Checks

### Describe the bug

when using the mistral api, it returns error code 422 since mistral doesnt accept extra inputs.

it's the same issue which i reported here: [hermes-agent](https://github.com/NousResearch/hermes-agent/issues/134)

### Osaurus version / commit

0.12.8

mimeding:
## Business rationale

Osaurus users on 0.17.4 who point a remote provider at Mistral (or other strict OpenAI-compatible endpoints such as Azure AI Foundry's Mistral passthrough) can't complete a chat turn. The first request returns HTTP 422 with `{\"type\":\"extra_forbidden\",\"loc\":[\"body\",\"max_completion_tokens\"],\"msg\":\"Extra inputs are not permitted\"}` and the session never gets further. Issue #556 has been open since the 0.12.8 era and a user has just re-confirmed it still happens on the current release. The fix unblocks Mistral end-to-end; a handful of other providers that validate their schemas strictly (DeepSeek, Groq, OpenRouter when routed to Mistral, Azure AI Foundry) benefit from the same change.

## Coding rationale

`RemoteChatRequest.CodingKeys` emitted `max_completion_tokens` unconditionally for every openaiLegacy outbound body. That's OpenAI's newer parameter name, introduced alongside the `o1` family. Three things are simultaneously true:

- OpenAI's **reasoning models** (`o1*`, `o3*`, `o4*`, `gpt-5*`) **require** `max_completion_tokens` and **reject** `max_tokens`.
- OpenAI's **non-reasoning models** accept **both** names.
- Most third-party OpenAI-compatible schemas (Mistral, DeepSeek, Groq, most of OpenRouter) only accept **`max_tokens`** and 422 on `max_completion_tokens`.

So the widely-accepted default is `max_tokens` — the only case where we absolutely cannot use it is an OpenAI reasoning model. The project already has `OpenAIReasoningProfile.matches(modelId:)` (`Models/Configuration/ModelOptions.swift:108`) that identifies those by prefix, and `buildChatRequest` already uses it to gate `temperature` / `top_p`. This PR reuses the same helper to pick the token-cap key at encode time.

I chose a custom `Encodable.encode(to:)` over introducing a second stored property because the canonical in-process field stays `max_completion_tokens` (every converter — Anthropic, Gemini, OpenResponses — reads from it), and the choice is purely a wire-format concern. Adding a `max_tokens` entry to `CodingKeys` and selecting at encode time keeps every existing caller unchanged.

I also had to lift `RemoteChatRequest`, `ReasoningConfig`, and `VeniceParameters` from fileprivate to internal so the new test file can construct requests without spinning up a full provider harness. They stay internal to the module.

## What changed and why

- `Packages/OsaurusCore/Services/Provider/RemoteProviderService.swift` — removed `private` on `RemoteChatRequest`, `ReasoningConfig`, `VeniceParameters`; added `max_tokens` to `CodingKeys`; replaced synthesized `Encodable` with an explicit `encode(to:)` that switches on `OpenAIReasoningProfile.matches(modelId:)` and emits exactly one of the two key names per request. Every other call-site invariant is preserved — nil stays nil, the Anthropic / Gemini / OpenResponses converters continue to read the same canonical field.
- `Packages/OsaurusCore/Tests/Provider/RemoteChatRequestEncodingTests.swift` — 5 contract tests pinning the key-selection:
  - Mistral-family model (`mistral-large-latest`) → `max_tokens` present, `max_completion_tokens` absent.
  - OpenAI non-reasoning model (`gpt-4o-mini`) → `max_tokens`.
  - OpenAI reasoning model (`o1-mini`) → `max_completion_tokens`, `max_tokens` absent.
  - GPT-5 family (`gpt-5-nano`) → `max_completion_tokens`, `max_tokens` absent.
  - Nil token cap → neither key present.

## Validation

- `swift test --filter 'RemoteChatRequestEncoding'` — 5 of 5 passing.
- `swift test` (full `OsaurusCore` package) — 888 of 888 passing.
- `xcrun swift-format lint --strict` on every touched file — clean.
- `swiftlint` on every touched file — no new violations introduced (pre-existing warnings in unrelated parts of `RemoteProviderService.swift` untouched).
- Manual: encoded a sample request via `JSONEncoder` for `mistral-large-latest` and confirmed the JSON body contains `\"max_tokens\":1024` and no `max_completion_tokens` key. Confirmed the same request body for `o1-mini` contains `\"max_completion_tokens\":1024` and no `max_tokens`.

## Non-scope

- No change to temperature / top_p / stop gating — the reasoning-model guardrails added earlier stay exactly as they were.
- No change to Anthropic / Gemini / OpenResponses conversion paths. They continue to read the canonical `max_completion_tokens` field and map it to each API's native name (`max_tokens`, `maxOutputTokens`, `max_output_tokens`).
- No new provider-selection UI. The fix is a pure wire-format correction; it doesn't need a user-facing toggle.

## Residual risks

- The heuristic is model-name-based, not provider-host-based. If someone configures a custom OpenAI-compatible provider and names a non-reasoning model with an `o1*` / `gpt-5*` prefix, the encoder will pick `max_completion_tokens`. In practice this only collides when the upstream actually forwards to OpenAI (OpenRouter's OpenAI route, Azure OpenAI), where OpenAI expects `max_completion_tokens` anyway — so the collision is benign.
- The reverse is also true: a Mistral reasoning variant (e.g., Magistral family) would get `max_tokens` because its name doesn't match the OpenAI prefix set. That's the correct behaviour — Mistral's schema accepts `max_tokens` regardless of model.

Closes #556

### DIFF
--- a/Packages/OsaurusCore/Services/Provider/RemoteProviderService.swift
+++ b/Packages/OsaurusCore/Services/Provider/RemoteProviderService.swift
@@ -1935,24 +1935,27 @@ private struct OpenResponsesSSEEvent: Decodable {
 // MARK: - Request/Response Models for Remote Provider
 
 /// Reasoning configuration for OpenAI reasoning models (o-series, gpt-5+).
-private struct ReasoningConfig: Encodable {
+struct ReasoningConfig: Encodable {
     let effort: String
 }
 
 /// Venice-specific parameters injected into the request body for Venice AI providers.
 /// See https://docs.venice.ai/api-reference/api-spec
-private struct VeniceParameters: Encodable {
+struct VeniceParameters: Encodable {
     var enable_web_search: String?
     var disable_thinking: Bool?
     var include_venice_system_prompt: Bool?
 }
 
 /// Chat request structure for remote providers (matches OpenAI format)
-private struct RemoteChatRequest: Encodable {
+struct RemoteChatRequest: Encodable {
     let model: String
     let messages: [ChatMessage]
     let temperature: Float?
-    let max_completion_tokens: Int?  // OpenAI's newer parameter name
+    /// Canonical token-cap field. Named after OpenAI's newer parameter; the
+    /// on-the-wire key is chosen in `encode(to:)` based on the model — see
+    /// the block below for the Mistral / OpenAI-compat rationale.
+    let max_completion_tokens: Int?
     let stream: Bool
     let top_p: Float?
     let frequency_penalty: Float?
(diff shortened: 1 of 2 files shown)


mimeding:
CI `test-core` note for maintainers — the failure is upstream, not this PR.

Every `test-core` run on this PR has failed at 2–5 min with the identical error:
```
error: Unable to find module dependency: 'CAsyncHTTPClient' (in target 'EventSource' from project 'EventSource')
```

The same error appears on #921 and #929. All three hit the EXACT same stack from `EventSource` / `CNIOLLHTTP` / `CNIOExtrasZlib` / `CNIOPosix` / `_NumericsShims`. It started sometime between 19:14 UTC today (#927's `test-core` was green) and 19:27 UTC (#928's first run failed). No contributor commit landed between those timestamps, so the flip is environmental — either a runner image change or a transitive version auto-resolving to something new.

**Likely root cause:** the MCP Swift SDK (`modelcontextprotocol/swift-sdk`) transitively pulls in `mattt/eventsource`, which introduced a SwiftPM *trait* system in v1.1+ that gates `AsyncHTTPClient` behind an opt-in trait. The trait is NOT declared as default and osaurus doesn't enable it anywhere, yet Xcode's SPM integration is pulling `CAsyncHTTPClient` as if it were required. That looks like an Xcode ↔ SwiftPM-traits interop bug.

**Things I verified:**
- `swift test` on the full `OsaurusCore` package passes 888/888 locally for this branch.
- The code diff is narrow and contains no Package.swift change, so nothing in this PR can be the cause.
- A force-push / amend re-run failed identically at 2m51s.

**Things I would try but cannot from a contributor branch:**
- Pin `swift-nio` to the last-known-good (2.98.0) in case 2.99.0's release on 2026-04-23T12:37:33Z is involved.
- Commit a `Package.resolved` so CI uses a frozen graph instead of re-resolving.
- Fork `swift-sdk` to pin EventSource <1.1.0.

Happy to help diagnose further if it's useful — the fix isn't in this PR's scope though.

mimeding:
@tpae when you have a minute — this closes #556 (the Mistral 422 on current release). Narrow-scope wire-format fix: emit `max_tokens` for all OpenAI-compatible endpoints and only switch to `max_completion_tokens` for OpenAI reasoning-model families (`o1*`/`o3*`/`o4*`/`gpt-5*`). 5 encoding contract tests pin the key selection per model family, full `OsaurusCore` suite 888/888 green. Two users have confirmed the 422 on 0.17.4 so the sooner this lands the less user churn on Mistral.

tpae (approved):

PR #929, opened 2026-04-23
feat(documents): XLSX read via CoreXLSX into a typed Workbook

mimeding:
## Applicability review (2026-06-06)

Status: superseded; do not merge as-is.

XLSX read-adapter work is now covered by current-main document/workbook infrastructure and the workbook workflow PR #1368. Keep this branch as reference material only.

> **Stacked on #927 while #927 awaits merge.** This branch has been rebased onto current `origin/main` (`fd2ffece`, #1015). Because #927 is not merged yet, this PR still carries the adapter-wrapper commit; reviewers should focus on the XLSX read adapter commit and the small rebase cleanup commit.

## Business rationale

This is the first spreadsheet-format slice that delivers file fidelity instead of reducing a workbook to unsupported bytes or a hand-copied table. Business users attach XLSX files constantly; preserving sheet names, formulas, booleans, merged ranges, and shared strings lets osaurus answer questions about real workbooks while keeping the typed data available for later agent tools. It advances the harness through file fidelity: the local agent can reason over the user's actual business document without sending it elsewhere or losing the structure that makes the file trustworthy.

## Coding rationale

CoreXLSX 0.14.2 is used only at the adapter boundary because it already exposes the read-side OOXML structures we need and resolves cleanly on current SwiftPM. The osaurus-owned `Workbook` model is the public intermediate instead of leaking CoreXLSX types, which keeps future emitter, tool, or vendor-fork work insulated from library churn. Styles are intentionally not parsed in this slice because CoreXLSX crashes on common style defaults; value and formula fidelity can ship independently, while number formats and explicit date typing stay out until a tolerant styles parser exists. The text fallback remains deliberately lossy and readable, while high-fidelity consumers use the structured representation.

## What changed

- Added CoreXLSX to `Packages/OsaurusCore/Package.swift` and declared the XLSX fixture resource.
- Added `Workbook`, `Sheet`, `Row`, `Cell`, `CellValue`, and `CellRange` as Sendable document models.
- Added `XLSXAdapter` with format detection, workbook parsing, formula preservation, merged-range capture, and readable text fallback.
- Registered the XLSX adapter through `DocumentAdaptersBootstrap` so `DocumentParser.parseAll` routes `.xlsx` through the registry.
- Added the small XLSX fixture and adapter/bootstrap tests.
- Kept the rebase cleanup limited to strict-lint shape fixes in touched files.

## Validation

[...]

PR #936, opened 2026-04-24
feat(documents): XLSX write via libxlsxwriter closes the round trip

mimeding:

Status: superseded; do not merge as-is.

XLSX write/emitter work is now represented through the current workbook workflow PR #1368 and existing emitter groundwork. Keep this branch as historical reference only.

> **Stacked on #929, which still carries #927 until #927 merges.** This branch has been rebased onto current `origin/main` (`fd2ffece`, #1015). Reviewers should focus on the XLSX writer commit and the small strict-lint rebase cleanup; the read-side and adapter-wrapper commits are present only because the lower stack is still open.

## Business rationale

This closes the spreadsheet round trip: osaurus can ingest a workbook into a typed `Workbook`, let agents reason over or modify it, and emit a fresh `.xlsx` file locally. That matters for finance, engineering, mining, operations, and other business workflows where a useful assistant must produce editable artifacts, not just summaries. It advances the file-fidelity pillar of the harness by keeping the workbook transformation offline, structured, and reviewable.

## Coding rationale

libxlsxwriter 1.2.4 is used as a plain SwiftPM dependency because it provides a stable C writer without bundling Homebrew, XCFrameworks, or generated binaries. The emitter consumes the osaurus-owned `Workbook` model introduced by the read adapter instead of carrying a second spreadsheet representation, so future tool and attachment work targets one shape. A1 references are parsed at the emitter boundary because the model already stores cells that way and libxlsxwriter wants zero-indexed row/column coordinates. The emitter always closes the workbook and deletes partial output on failure, making errors observable instead of leaving behind a misleading half-written artifact.

## What changed

- Added libxlsxwriter to `Packages/OsaurusCore/Package.swift`.
- Added `XLSXEmitter` for `Workbook`-backed `.xlsx` output.
- Registered the emitter with `DocumentAdaptersBootstrap`.
- Added round-trip tests that emit workbooks, parse them back through `XLSXAdapter`, and assert strings, numbers, booleans, formulas, merged ranges, and sheet names survive.
- Kept rebase cleanup limited to strict-lint shape fixes inherited from the stacked lower branch.

## Validation

- `git fetch origin && git rebase origin/main` — completed after resolving the startup-bootstrap overlap and `Package.resolved` origin hash.
- `swift build --package-path Packages/OsaurusCore` — passed.
- `swift build --package-path Packages/OsaurusCore -c release` — passed.
- `swift test --package-path Packages/OsaurusCore` — passed, 1468 tests in 198 suites, with the sandbox integration suite skipped by its normal env gate.
- `xcrun swift-format lint --strict` on every touched Swift file — passed.
- `swiftlint lint --strict` on every touched Swift file — passed.
- `git diff --check origin/main...HEAD` — passed.
- CLI gate skipped because this slice does not touch `Packages/OsaurusCLI`.

## Non-scope

- No workbook agent tools; those land in the next document-stack slice.
- No styles, number formats, column widths, fills, or explicit date typing.
- No legacy `.xls` or ODS support.
- No attachment-kind changes in this PR.
- No acknowledgements UI update for libxlsxwriter or its bundled C dependency notes.

## Residual risks

libxlsxwriter reports numeric C error codes rather than Swift-native errors, so the emitter maps failures into `DocumentAdapterError.writeFailed` with the affected cell or range plus the raw code for diagnosis. Formula cached results are not written yet; Excel-compatible readers may recalculate on open. Style fidelity remains intentionally out of scope until the read-side styles parser is tolerant of common OOXML defaults.


mimeding:
Realignment update: I am moving this writeback work to `osaurus-xlsx`. The round-trip tests, formula preservation, and cleanup-on-failure behavior are still valuable, but the dependency and emitter implementation should live in the XLSX plugin rather than `OsaurusCore`. Core should only expose a neutral plugin emitter registration contract.

mimeding:
Closing this old document-stack slice as superseded by the newer consolidated Business Document I/O Workbench direction. I will not revive these stale document PRs unless a specific missing capability is still absent on current main.

PR #943, opened 2026-04-24
feat: jang_config.json chat metadata + LFM false-thinking-block fix

jjang-ai:
Two features prepping osaurus for the DSV4 / Kimi K2.6 model-family rollout and fixing a user-visible bug already reported on LFM2-class models. **Stacked on top of #932** — review / merge #932 first.

## 1. `jang_config.json > chat > sampling_defaults` support

New JANG / JANGTQ bundles (DSV4-Flash-JANGTQ, Kimi K2.6 JANG snapshots, newer Gemma-4 / Qwen-3.6 JANG converters) stamp authoritative sampling defaults into `jang_config.json`. These can differ from the source model's generic `generation_config.json` — critically, **DSV4's `convert_dsv4_jangtq.py` reads `inference/generate.py` defaults directly and stamps `temperature=0.6`, while upstream HF ships `temperature=1.0`.**

`LocalGenerationDefaults` now consults both files and merges:

1. `jang_config.json > chat > sampling_defaults` (authoritative)
2. `generation_config.json` (HF fallback)

Merge semantics: primary fills first, fallback fills any field the primary omits. A DSV4 bundle with `jang_config` `temp=0.6` and HF `top_k=64` produces `{temp=0.6, top_p=0.95, top_k=64}` — the expected cross-file overlay.

Schema recognised (subset osaurus honors):

```json
{
  "model_family": "deepseek_v4",
  "chat": {
    "sampling_defaults": {
      "temperature": 0.6,
      "top_p": 0.95,
      "top_k": 20,
      "repetition_penalty": 1.05
    }
  }
}
```

Fields outside `chat.sampling_defaults` (quantization, source_model, crack_surgery, architecture, reasoning mode tables, tool_calling parser stamp) are intentionally ignored here — they belong to vmlx's model loader, not osaurus's sampling overlay.

## 2. LFM non-thinking-model false-positive fix

**Bug report:** \"even if the model doesn't support thinking, the response gets rendered in the thinking block.\"

**Root cause:** vmlx's reasoning-parser heuristic (`LLMModelFactory.swift:720-750`) routes ANY `model_type` that doesn't match `gemma4` / `mistral` / `gemma` prefixes to `think_xml` with `startInReasoning: true`. For LFM2 / LFM2-MoE / LFM2.5 (`lfm2*`), classic Llama instruct checkpoints, and other families that genuinely don't emit `<think>` tags, that mis-classification surfaces the model's **first output bytes** as `Generation.reasoning` events — which osaurus's UI renders in the thinking pane.

**Fix:** `GenerationEventMapper.map` grows a `supportsThinking: Bool` parameter (default `true`, backwards compatible). When `false`, `.reasoning(text)` events are re-labelled as `.tokens(text)` so the UI renders them in the response block — preserving delivery order across interleaved events and still filtering empty strings.

`ModelRuntime.generateEventStream` wires this up from `LocalReasoningCapability.capability(forModelId:).supportsThinking`, which reads the model's template text for `<think>` / `<|think|>` / `enable_thinking` markers. Models with none of those → `supportsThinking=false` → coercion engages → thinking block disappears for LFM2 et al.

This is an osaurus-side mitigation for a vmlx heuristic gap. A permanent upstream fix would widen the `none` branch of the reasoning-parser heuristic to cover `lfm2*`, but the osaurus mitigation is correct either way — the source of truth stays the local template capability.

## Test plan

- [x] **LocalGenerationDefaults** — 10 new tests covering DSV4 fixture, full-field population, missing-chat-subtree, malformed JSON, merge overlay semantics, filesystem round-trip for BOTH files / JANG-only
- [x] **GenerationEventMapper** — 4 new tests covering non-thinking coercion, interleaved order preservation, empty-delta filtering, default-value guard
- [x] `swift test --parallel`: **913 / 913 pass** across 121 suites
- [x] `swift build` clean
- [ ] Manual: chat with LFM2-1.5B — verify response renders in response block, NOT thinking pane
- [ ] Manual: chat with DSV4-JANGTQ4 once vmlx ships model_type registration — verify temp=0.6 applies

## Not in this PR (vmlx side)

- `deepseek_v4` model_type registration in `LLMModelFactory`
- DSML tool-call parser (new `.dsml` case in `ToolCallFormat`)
- Widening reasoning-parser heuristic's `.none` branch to include `lfm2*`

Osaurus doesn't need any of those to build / run; when vmlx ships them the PR's wiring picks them up transparently.

### DIFF
--- a/Packages/OsaurusCore/Managers/Model/ModelManager.swift
+++ b/Packages/OsaurusCore/Managers/Model/ModelManager.swift
@@ -952,6 +952,7 @@ extension ModelManager {
         cachedLocalModels = nil
         localModelsCacheLock.unlock()
         LocalReasoningCapability.invalidate()
+        LocalGenerationDefaults.invalidate()
     }
 
     /// Discover locally downloaded models. Cached until invalidated by model download/delete.
--- a/Packages/OsaurusCore/Services/LocalGenerationDefaults.swift
+++ b/Packages/OsaurusCore/Services/LocalGenerationDefaults.swift
@@ -2,25 +2,39 @@
 //  LocalGenerationDefaults.swift
 //  osaurus
 //
-//  Reads `generation_config.json` from a locally-installed model bundle and
-//  surfaces the sampling defaults (temperature / top_p / top_k / repetition_
-//  penalty) so osaurus can honor them when the OpenAI-wire request omits
-//  the corresponding field.
+//  Reads sampling defaults from a locally-installed model bundle and
+//  surfaces them (temperature / top_p / top_k / repetition_penalty) so
+//  osaurus can honor them when the OpenAI-wire request omits the
+//  corresponding field.
 //
-//  Hugging Face ships these defaults alongside every instruction-tuned
-//  checkpoint. Ignoring them serves, e.g., Qwen 3.5 397B-A17B at 0.7
-//  temperature when its training recipe specifies 0.6, or Gemma-4 26B-A4B
-//  with top_k disabled when the recipe specifies top_k=64. vmlx's
-//  `GenerationConfigFile` (Libraries/MLXLMCommon/GenerationConfigFile.swift)
-//  only decodes `eos_token_id` from this file, so reading sampling fields
-//  is osaurus's job.
+//  Two sources are consulted, in priority order:
 //
-//  JANG / JANGTQ bundles that ship a `generation_config.json` are read like
-//  any other model. Bundles that don't ship one (some JANG snapshots) fall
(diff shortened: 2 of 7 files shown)


jjang-ai:
Superseded by #944. This branch was authored before #932 + #934 merged to main and duplicates their logic at the GenerationEventMapper layer. #944 contains the clean version (jang_config.json chat.sampling_defaults overlay + vmlx pin bump), no overlap with existing osaurus coercion at ModelRuntime.streamWithTools.

PR #945, opened 2026-04-24
fix(speech): adopt FluidAudio 0.14 decoderState parameter

mimeding:
## Business rationale

This is defensive maintenance. FluidAudio 0.14.0 shipped on 2026-04-23; osaurus's \`Package.swift\` pins \`FluidAudio\` with \`from: \"0.13.4\"\`, which means any fresh \`swift package resolve\` picks 0.14.0. 0.14 removed the single-argument \`transcribe(_:)\` in favour of \`transcribe(_:decoderState:)\`, so the current \`SpeechService.swift\` on main compiles only against CI's cached pre-0.14 DerivedData. The next cache eviction (or anyone running a clean build) turns into four compile errors and a broken \`speech\` feature.

I hit this locally yesterday after the #931 resolver rerun — main built against the cached 0.14-compatible binaries but a fresh checkout failed with:

```
error: missing argument for parameter 'decoderState' in call
  SpeechService.swift:744:104
  SpeechService.swift:947:85
  SpeechService.swift:1466:72
  SpeechService.swift:1478:72
```

Narrow one-file fix so there's no scope creep with the stage-4 document work.

## Coding rationale

Four call sites needed the new parameter; grouping them around a tiny factory keeps each call site legible:

- \`SpeechDecoderStateFactory.make(for:)\` — reads \`manager.decoderLayerCount\` (async on FluidAudio 0.14+) and returns a ready \`TdtDecoderState\`.
- Batch file path (\`transcribeAudioFile\`) — builds a fresh state per call inside the detached task. One-shot call, no reuse benefit.
- Streaming finalize on mic stop — same pattern.
- \`TranscriptionWorker.finalizeSegment\` — reuses a single state across segments in a session (\`decoderState: TdtDecoderState?\` on the actor) so the decoder's KV cache stays warm between chunks. Matches FluidAudio's guidance for the streaming path.
- \`TranscriptionWorker.updatePreview\` — throwaway state per preview, so a noisy partial chunk can't poison the next confirmed segment's state.

## What changed and why

- \`Packages/OsaurusCore/Managers/SpeechService.swift\` — added the factory enum and threaded \`decoderState\` through four call sites. Added \`TranscriptionWorker.decoderState: TdtDecoderState?\` for cross-segment reuse. No behaviour change beyond complying with the new API signature.

## Validation

- \`swift build --package-path Packages/OsaurusCore\` — succeeds against FluidAudio 0.14.0 (locally resolved).
- \`swift test\` (full \`OsaurusCore\` package) — 916 of 916 passing.
- \`xcrun swift-format lint --strict\` on \`SpeechService.swift\` — clean.
- \`swiftlint\` on \`SpeechService.swift\` — no new violations; pre-existing warnings on lines 157 / 354 / 452 / 464 / 567 / 801 / 845 / 1028 / 1249 are all outside the edit ranges.

## Non-scope

- No \`FluidAudio\` version pin bump. \`from: \"0.13.4\"\` stays; this PR just makes the code compile against the already-pulled-in 0.14.0.
- No streaming-quality tuning. Reusing \`decoderState\` across segments is the right default for continuity; tuning the reset interval is out of scope.
- No behaviour change on the batch file path. One-shot state still makes sense there.

## Residual risks

- \`TranscriptionWorker.decoderState\` is an actor-isolated \`var\`, released with the worker. No cross-session leakage.
- Preview uses a throwaway state — slightly more compute per partial chunk but removes a whole class of "partial poison" bugs. Acceptable given previews are short-lived.
- If a future FluidAudio bumps the signature again, this factory is the single seam to update.

### DIFF
--- a/Packages/OsaurusCore/Managers/SpeechService.swift
+++ b/Packages/OsaurusCore/Managers/SpeechService.swift
@@ -528,6 +528,18 @@ final class SendableAsrManager: @unchecked Sendable {
     init(_ manager: AsrManager) { self.manager = manager }
 }
 
+/// FluidAudio 0.14 moved the decoder state onto the caller: every
+/// `transcribe(_:decoderState:)` invocation threads a fresh or reused
+/// `TdtDecoderState` so the library doesn't have to stash per-call
+/// mutable state behind its own lock. Decoder-layer count comes off
+/// the manager itself, so the factory is a thin `async` shim.
+private enum SpeechDecoderStateFactory {
+    static func make(for manager: AsrManager) async -> TdtDecoderState {
+        let decoderLayers = await manager.decoderLayerCount
+        return TdtDecoderState.make(decoderLayers: decoderLayers)
+    }
+}
+
 // MARK: - Speech Service
 
 /// Service for audio transcription using FluidAudio
@@ -741,7 +753,10 @@ public final class SpeechService: ObservableObject {
         defer { isTranscribing = false }
 
         do {
-            let result = try await Task.detached { try await wrappedManager.manager.transcribe(audioURL) }.value
+            let result = try await Task.detached {
+                var decoderState = await SpeechDecoderStateFactory.make(for: wrappedManager.manager)
+                return try await wrappedManager.manager.transcribe(audioURL, decoderState: &decoderState)
+            }.value
 
             return TranscriptionResult(
                 text: result.text,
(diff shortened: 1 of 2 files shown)


mimeding:
Amended with a Package.swift pin bump — `FluidAudio` `from: "0.13.4"` → `from: "0.14.0"`. The initial push failed `test-core` with `cannot find type 'TdtDecoderState' in scope`: the 0.14-API code doesn't compile against CI's cached pre-0.14 `FluidAudio` checkout, and the `from: 0.13.4` constraint alone let the cache key match main's, so the runner reused the stale artifact. Bumping the floor to 0.14.0 invalidates the cache key and pins the minimum to the version the code actually requires. Local: full `OsaurusCore` suite 916/916 green against 0.14.0. Pushing the amend now triggers a fresh CI run.

mimeding:
@tpae small defensive PR whenever you have time — `SpeechService` currently compiles only against CI's cached pre-0.14 `FluidAudio`; a clean checkout or cache eviction breaks the build with four `missing argument for parameter 'decoderState'` errors (FluidAudio 0.14 removed the no-state overload). One-file fix: tiny factory + four call-site updates + a floor bump to `from: "0.14.0"`. Full suite 916/916 green, `test-core` just came back green on the current revision. Narrow scope on purpose — zero overlap with the stage-4 document stack.

tpae (approved):

PR #954, opened 2026-04-26
[codex] Gate local thinking controls and SQLCipher import

mimeding:
## Summary

This keeps the fallback local thinking toggle hidden unless template inspection confirms both an `enable_thinking` switch and reasoning markers Osaurus can parse. It also cleans public compatibility wording to describe generic protocol-compatible clients, hardens two environment-sensitive tests, fixes the vendored SQLCipher Swift import on newer macOS SDKs, and aligns CI with the release workflow's pinned Xcode 26.4.1 toolchain.

## Business Rationale

Users should not see reasoning controls that the selected local model cannot actually honor. Hiding unsupported toggles reduces confusing model behavior and avoids implying that Osaurus can control reasoning for templates that only expose a kwarg without recognizable reasoning output. The SQLCipher import guard and CI toolchain alignment keep the current developer and GitHub test paths able to compile Osaurus after the storage-encryption baseline, which is necessary before broader LLM capability work can be reviewed safely. Generic compatibility wording keeps public docs focused on Osaurus-owned protocol support.

## Coding Rationale

`LocalReasoningCapability` already centralizes template inspection, and `ModelOptions` already gates fallback profile selection. Adding `isToggleableThinking` preserves the existing JANG/Gemma/Qwen detection while making the UI profile depend on the combined capability instead of a single kwarg. The SQLCipher change hides only the loadable-extension C API from Swift's Clang import, which Osaurus does not use, while preserving the core SQLite/SQLCipher API and adding vendor guard tests so future SQLCipher bumps keep the import boundary intact. The CI change keeps the test workflow on the same Xcode 26.4.1 pin already used by the release workflow and avoids an unrelated Xcode 26.3 `EventSource` transitive module build failure.

## Expected LLM Performance Impact

Fewer local-model dispatches with misleading thinking controls and fewer degraded first responses from models whose templates mention `enable_thinking` but do not produce parseable reasoning markers. This should improve model-option accuracy, reduce unsupported reasoning toggles, and make future capability changes easier to validate consistently. The compile and CI fixes do not change model behavior directly, but they keep the capability baseline buildable on current Xcode so follow-up LLM harness improvements can land without test-environment drift.

## Changes

- [x] Behavior change
- [ ] UI change
- [x] Refactor / chore
- [x] Tests
- [x] Docs

## Test Plan

[...]

PR #956, opened 2026-04-27
fix flaky tests and updated headers for sqlite collision prevention

tpae:
## Summary

fixed flaky tests

## Checklist

### DIFF
--- a/Packages/OsaurusCore/Models/Agent/Skill.swift
+++ b/Packages/OsaurusCore/Models/Agent/Skill.swift
@@ -403,7 +403,7 @@ public struct Skill: Codable, Identifiable, Sendable, Equatable {
                 // SystemPromptComposer fires for sandbox+autonomous agents
                 // without manual intervention. Users who don't want the
                 // injection can disable this skill from the skill catalog
-                // — `pluginCreatorSkillSection` honours the flag.
+                // — `PluginCreatorGate` honours the flag via `skillEnabled`.
                 enabled: true,
                 instructions: sandboxPluginCreatorInstructions,
                 isBuiltIn: true,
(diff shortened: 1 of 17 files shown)


mimeding:
CI update from checking this against #954/#955: #956 still fails `test-core`.

Observed failures in run https://github.com/osaurus-ai/osaurus/actions/runs/24996830968/job/73197126271:

- `SandboxManagerCleanupTests.cleanupAfterFailure_removesStaleContainerDirectory`
- `PluginCreatorInjectionTests.composeChatContext_injectsPluginCreatorWhenCatalogEmpty`

The direct `pluginCreatorSkillSection_returnsContentWhenSkillEnabled` case passed in this run, so the remaining plugin failure looks specific to the `composeChatContext` gate around `ToolRegistry.shared.dynamicCatalogIsEmpty()` / process-wide registry state rather than the built-in skill being unavailable.

Two concrete follow-ups that look worth checking:

- The sandbox cleanup tests now use `SandboxTestLock`, but `cleanupAfterFailure_removesStaleContainerDirectory` still uses `OsaurusPaths.container()` through `staleContainerDir`. It may also need `StoragePathsTestLock` or a temp `OsaurusPaths.overrideRoot` so concurrent storage/path tests cannot move the cleanup target while the test is suspended.
- The dynamic-catalog lock should cover every test that registers/enables/unregisters dynamic tools or redirects `ToolConfigurationStore.overrideDirectory`. In the current patch, the MCP tests set `ToolConfigurationStore.overrideDirectory = tempDir` inside `DynamicCatalogTestLock`, but I don’t see a restore of the previous override or temp cleanup; that global leak can affect later prompt-composition tests even when the tool itself is unregistered.

The run used Xcode 26.3. That was not the immediate failure here because the package built and tests launched, but #954/#955 already hit an unrelated EventSource transitive-module build failure on 26.3 before I aligned their CI pins to 26.4.1. After the flake fixes are green, it may still be worth aligning this branch/main CI with the release workflow’s Xcode pin.

PR #974, opened 2026-04-29
Add on-demand high-fidelity skills

mimeding:
## Applicability review (2026-06-06)

Status: stale/reference only; do not merge as-is.

The plugin/skill stack now includes marketplace import, artifact aggregation, skill search, and capability search/load paths. Revisit this direction only after #1358 and #1363 land and maintainers request an on-demand-skill follow-up.

## Business rationale
High-fidelity skills should make Osaurus more useful as a local agent harness without making every chat slower or more expensive by default. This advances the agents, plugins, and file-fidelity pillars by letting users discover business-file and contributor skills on demand, then load them only when the task needs that extra context.

## Coding rationale
The activation model stays backward compatible with existing Agent Skills metadata while adding Osaurus-specific discoverability, default-selection, and on-demand fields. Session-loaded skills are persisted in `SessionToolStateStore` so `capabilities_load` has real chat-lifetime effects instead of being a one-shot tool response. Startup injection filters to activation-selected skills, while on-demand skills stay indexed for `capabilities_search` and can still be loaded explicitly. The first-party pack is shipped as portable markdown under `skills/first-party` with marketplace metadata, keeping skill content maintainable outside app binaries.

## What changed
- Add backward-compatible Osaurus skill activation metadata for discoverable, default-selected, and on-demand behavior.
- Persist skills loaded through `capabilities_load` in session state so they stay available for the chat without becoming globally selected.
- Keep high-fidelity skills out of startup prompts unless selected or loaded, while indexing them for `capabilities_search`.
- Add a portable first-party high-fidelity skill pack under `skills/first-party` with `.claude-plugin/marketplace.json` import metadata.
- Update public docs to match the post-#893 single Chat/Agent loop and lightweight activation model.

## Validation
- `swift build --package-path Packages/OsaurusCore`: passed.
- `swift build --package-path Packages/OsaurusCore -c release`: passed.
- `swift test --package-path Packages/OsaurusCore`: passed, 1445 tests in 194 suites; sandbox integration skipped behind its env var.
- `xcrun swift-format lint --strict` on touched Swift files: passed.
- `swiftlint lint --strict` on touched Swift files: passed.
- `git diff --check`: passed.
- CLI gate skipped: this PR does not touch `Packages/OsaurusCLI`.

## Non-scope
- Does not add PPTX/XLSX business-file services; this PR only provides the lightweight activation/runtime skill foundation.
- Does not change model-provider routing or Work Mode behavior.

## Residual risks
[...]

PR #979, opened 2026-04-30
Use lightweight context for Foundation

### LINKED ISSUE #973: Default system prompt is too heavy for Foundation
### Checks

### Describe the bug

Apple's Foundation model is inherently limited, especially in context size. I was hoping to use it for some rudimentary scheduled stuff (e.g. price checking every morning using the browse tool since changedetection gets blocked as a bot), but trying to do almost anything exceeds the context window.

Perhaps a super trimmed down default context could be special cased when using Foundation as the model.

### Steps to reproduce

_No response_

### Provider type

Local model (MLX)

### Model name

Foundation

### Osaurus version / commit

0.18.2

### macOS version

26.4.1

### Apple Silicon chip

M4 Max 64 GB

### Xcode version

26.4.1

### Logs

```shell

```

### Screenshots

_No response_

mimeding:

Status: stale/reference only; do not merge as-is.

Foundation/context behavior has changed since this branch. Rebuild only if a current-main performance or context-size regression is reproduced and scoped to a small, focused PR.

## Business rationale
Apple Foundation is the lightest local chat path, but its small context window makes the default Osaurus harness feel unreliable if baseline tool schemas and memory consume the budget before the user's request arrives. This advances the harness pillar by making Foundation chats start lean while still preserving the agent's ability to add tools on demand when the task actually needs them.

## Coding rationale
- Reused `ContextSizeResolver` from current `main` instead of carrying a second Foundation-only detector, so context policy remains centralized.
- Treated tiny-context models as a baseline-tool trim rather than a hard tool ban, which keeps `capabilities_load` and other explicit session-loaded tools available when requested.
- Kept gated guide text, skill suggestions, and memory estimates out of the lightweight prompt so Foundation pays only for active capabilities.
- Added tests at the prompt-composer boundary because that is where tool visibility, memory disabling, and preview/send parity meet.

## What changed
- Rebased the PR onto `origin/main` and collapsed stale CI/TTS tail commits out of the branch.
- Updated context-size comments to describe default-baseline trimming rather than a global tool ban.
- Added the `includeBaselineBuiltIns` path through tool resolution and always-loaded-name snapshots.
- Updated chat budget preview memory accounting for context-size memory disablement.
- Added SessionPreflightCache coverage for Foundation default-baseline omission, session-loaded tools, and non-Foundation baseline preservation.

## Validation
- `swift build --package-path Packages/OsaurusCore`
- `swift build --package-path Packages/OsaurusCore -c release`
- `swift test --package-path Packages/OsaurusCore` (1439 tests, 192 suites)
- `xcrun swift-format lint --strict` on touched Swift files
- `swiftlint lint --strict` on touched Swift files
- `git diff --check`
- CLI gate skipped; this PR does not touch `Packages/OsaurusCLI`.

## Non-scope
- No changes to provider selection, model installation, or the broader CI workflow.
- No changes to the on-demand skills package in #974.
- No changes to the TTS/FluidAudio commits that previously leaked onto this branch.

## Residual risks
- Foundation can still exceed budget on very long conversations; this PR trims startup weight, not conversation history.
- GitHub checks restarted after the force-with-lease push and must finish green before merge.

Fixes #973.

### DIFF
--- a/Packages/OsaurusCore/Services/Chat/ContextSizeClass.swift
+++ b/Packages/OsaurusCore/Services/Chat/ContextSizeClass.swift
@@ -6,9 +6,8 @@
 //  prompt features that don't fit into very small windows. Apple's
 //  Foundation model has a ~4K window; even before any user message
 //  the always-loaded tool schemas push past it. The system-prompt
-//  composer reads this resolver at compose time and ORs the result
-//  into the agent's effective tools/memory disable flags so we never
-//  ship a request that's already over budget.
+//  composer reads this resolver at compose time so it can drop default
+//  prompt features before they crowd out the user's actual request.
 //
 
 import Foundation
@@ -17,12 +16,12 @@ import Foundation
 
 /// Coarse classification of a model's nominal context window. Three
 /// buckets are enough — the prompt composer only needs to decide
-/// whether to disable tools (tiny only) and/or memory (tiny + small).
+/// whether to trim baseline tools (tiny only) and/or memory (tiny + small).
 public enum ContextSizeClass: Sendable, Equatable {
     /// `<= 4096` tokens. Apple Foundation and any equally tight
-    /// future model. Tools, memory, and skill suggestions all auto
-    /// off — at this size even the always-loaded tool JSON schemas
-    /// cost more than the available budget.
+    /// future model. Memory, skill suggestions, and the default tool
+    /// baseline are off — explicit on-demand tools can still be added
+    /// when the agent or user asks for them.
     case tiny
 
     /// `<= 8192` tokens. Fits a reasonable chat schema but not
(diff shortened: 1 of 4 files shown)


tpae:
@mimeding thank you for the contribution, however, i feel this PR is a bit redundant because user should be able to disable tools and memory per agent configuration. we're not sure how the `foundation` model will evolve in the future, would rather prefer user does their own configuration vs. defining specific code paths.

tpae:
@mimeding i was able to address this issue from a critical bug that was found: https://github.com/osaurus-ai/osaurus/pull/996

mimeding:
Thanks for the follow-up. I agree this should not be advanced as-is: #996 covered the underlying issue, and the broader behavior belongs in per-agent configuration rather than Foundation-specific code paths.

I am marking #979 as superseded in the stabilization plan and will not push or request readiness for it unless you want a smaller follow-up.

mimeding:
Agreed. I will not continue this as a Foundation-specific code path. The better direction is per-agent user configuration for tools and memory, and this PR can be treated as superseded by the later critical-bug fix you linked.

mimeding:
Closing this as superseded. The Foundation-specific lightweight-context path is not the right direction; per-agent user configuration for tools and memory plus the later critical-bug fix are the better path.

Issue #1001, opened 2026-05-02
Add a toggle to auto-play agent responses with TTS, per-conversation


fortinmike:
### Checks

- [x] I have searched existing issues and discussions

### Area

Chat / conversation

### Problem statement

The new TTS support works great but you have to manually press the speaker button below each response to hear the agent speak.

### Proposed solution

I think the logical next step would be to add a per-conversation toggle to enable "auto-speak" when an agent reply comes in (disabled by default). This way you could theoretically have full voice-only conversations like #417 suggests and hear the agent respond from the background when it's done with a task and the app is not frontmost anymore.

**Per-conversation:** So you can decide on-the-fly if you want a conversation to be audible or not. Probably enabled in a single conversation at a time so agents don't try to speak over each other as messages come in, if multiple conversations are going on in parallel.

**Default to off:** Because otherwise it would get noisy fast and there could be times where the agent speaks when you didn't expect it to, such as in public spaces, etc.

### Contribution

- [ ] I would be willing to submit a PR for this feature

PR #1025, opened 2026-05-04
[codex] Sync refreshed provider service models

mimeding:
## Business rationale
Custom OpenAI-compatible providers can add models while Osaurus is running. After #1020, the picker refreshes visible model lists on open; this follow-up keeps the backing provider service in sync too, so a model shown as available is also reflected by the service that handles requests. That protects trust in the local multi-provider harness.

## Coding rationale
- Build on #1020's refreshConnectedProviders/refetchModels path rather than keeping a second refresh API.
- Update the existing RemoteProviderService actor only after the merged model list actually changes, preserving throttling/coalescing behavior and avoiding reconnects.
- Add a process-wide remote-provider test lock because Swift Testing .serialized only serializes tests within one suite; cross-suite singleton/notification races made ModelPickerItemCacheTests fail in full-suite runs while passing in isolation.
- Keep UI behavior unchanged; this PR is service state plus test isolation.

## What changed
- refetchModels(providerId:) now updates the connected service's model snapshot when provider state changes.
- Added RemoteProviderTestLock.
- Wrapped remote-provider refresh and model-picker cache tests that share provider/cache singletons.

## Validation
- swift build --package-path Packages/OsaurusCore
- swift build -c release --package-path Packages/OsaurusCore
- swift test --package-path Packages/OsaurusCore
- swift test --package-path Packages/OsaurusCore --filter RemoteProviderManagerRefreshTests
- swift test --package-path Packages/OsaurusCore --filter "RemoteProviderManagerRefreshTests|ModelPickerItemCacheTests"
- xcrun swift-format lint --strict Packages/OsaurusCore/Managers/RemoteProviderManager.swift Packages/OsaurusCore/Tests/Helpers/RemoteProviderTestLock.swift Packages/OsaurusCore/Tests/Model/ModelPickerItemCacheTests.swift Packages/OsaurusCore/Tests/Provider/RemoteProviderManagerRefreshTests.swift
- swiftlint lint --strict Packages/OsaurusCore/Managers/RemoteProviderManager.swift Packages/OsaurusCore/Tests/Helpers/RemoteProviderTestLock.swift Packages/OsaurusCore/Tests/Model/ModelPickerItemCacheTests.swift Packages/OsaurusCore/Tests/Provider/RemoteProviderManagerRefreshTests.swift
- git diff --check

## Non-scope
- Does not change picker UI refresh behavior from #1020.
- Does not add a second provider refresh API.
- Does not change provider discovery or connection policy.

## Residual risks
[...]

PR #1026, opened 2026-05-04
auto speak tts in chat

RaajeevChandran:
## Summary

addresses #1001 

## Changes

- [x] Behavior change
- [ ] Refactor / chore
- [ ] Tests
- [ ] Docs

## Checklist


mimeding:
@tpae maintainer CI note:

This external PR appears to be failing `test-core` on the same shared provider/model-picker race currently present on `main`, not necessarily because of the TTS changes here. The failed tests match the cluster fixed by #1025.

Recommended next step: merge/review #1025 first, then rerun this PR's checks. If it is still red after #1025 lands, that remaining failure should be treated as branch-specific and debugged here.

Why: rerunning now is likely to reproduce the known red-base failure and hide whether this TTS PR has its own issues.

PR #1037, opened 2026-05-06
Bump vmlx-swift-lm to b9da180 (Ling/ZAYA hardening + BatchEngine lifecycle)

jjang-ai:
## Summary

Bumps `vmlx-swift-lm` from `4a832400` to `88fc352b` (12 commits ahead on `osaurus-ai/vmlx-swift-lm` `main`, verified via `git ls-remote`). The bump brings two production-witnessed Ling 2.6 Flash JANGTQ multi-turn fixes plus the full ZAYA1 (Zyphra; `model_type=zaya`) CCA-attention port, and pulls forward adjacent runtime hardening (Gemma4 SWA paged-incompatible flip, audio media-salt fingerprinting, DSV4 reasoning-effort context strip, etc.).

## Production-witnessed Ling fixes the bump pulls in

| Symptom | Root cause | vmlx fix |
|---|---|---|
| Ling chat UI shows Stop button + spinner with no visible answer text on multi-turn | `generateTask` did not derive prompt tail from `TokenIterator.promptTokenIds`, so `ReasoningParser.forPrompt` fell back to a family stamp that opened reasoning state. The whole answer streamed to `.reasoning` instead of `.chunk`. | `a138f47` reconstructs the decoded prompt tail when the caller does not pass `promptTail`, so `ReasoningParser.forPrompt` reads the actual rendered prompt state. |
| Ling cross-turn cache-state desync (manifests as language drift on long multi-turn flows) | `BailingLinearAttention` / `BailingMLAAttention` applied RoPE with scalar `cache.offset`. For mixed-length B>1 batched decode the merged `BatchArraysCache.offset` was the max across slots, leaving non-leader slots with wrong RoPE positions and desynced recurrent GLA state across turns. | `88fc352` switches both attention paths to `applyRotaryPosition(rope, to:cache:)`. `BatchArraysCache` gains per-sequence `offsets: [Int]` + `offsetArray: MLXArray` + `advance(by:)` so the recurrent GLA state advances per-slot. |

## ZAYA1 CCA-attention (Zyphra; `model_type=zaya`)

vmlx-side, ZAYA dispatches via `LLMTypeRegistry.types["zaya"] = dispatchZaya` to a real `ZayaModel` (replacing the prior unsupported throw). Bundle variants supported: BF16 base + JANGTQ2 / JANGTQ4 / MXFP4 routed-expert. The hybrid cache topology (`ZayaCCACache`: KV + path-dependent `conv_state` + `prev_hs`) round-trips through:

- `BatchZayaCCACache` — per-slot CCA gather/scatter for batched decode
- `TQDiskSerializer.zayaCCA` LayerKind — disk-tier prefix reuse round-trip
- `BatchEngine` admission auto-flips `setHybrid(true)` + `setPagedIncompatible(true)` on first `ZayaCCACache` slot
- `LLMUserInputProcessor.defaultContext` clamps `enable_thinking=false` for `model_type=zaya`/`zyphra` (until live thinking-on parity is proven on JANGTQ2/JANGTQ4/MXFP4 — see vmlx `OSAURUS-RUNTIME-HANDOFF-2026-05-06.md` ZAYA Production Notes)
- `ToolCallFormat.zayaXml` — `<zyphra_tool_call>` wrapper around the standard XML function syntax

Osaurus-side wiring mirrors the existing Ling pattern:

- `ModelFamilyNames.isZayaFamily` — boundary-aware regex `(^|/)zaya[\-0-9]` matches `Zaya1-…`, `Zyphra/Zaya1-…`, `Zyphra/Zaya-S-…` and rejects adjacent names like `dataset/zayasaurus`, `lazyaardvark`, `dazaya-llm`.
- `ZayaRuntimeProfile` (registered before `AutoThinkingProfile`) reserves ZAYA IDs so the chat UI does not expose a Thinking toggle that would override the vmlx clamp.
- `MLXBatchAdapter.additionalContext` short-circuits to `enable_thinking=false` for ZAYA, so a stale persisted `disableThinking=false` cannot cross the caller-wins merge in `LLMUserInputProcessor.prepare`.
- `ModelRuntime.isKnownHybridModel` adds the ZAYA family for eager `setHybrid(true)` parity with vmlx's BatchEngine auto-flip on first `ZayaCCACache` slot admission.

## isKnownHybridModel cleanup

MiniMax M2 / M2.7 was historically in the eager-flip list with a `// gated SSM in some layers` comment, but vmlx's `MiniMaxModel` and `MiniMaxJANGTQModel` use only standard `KVCache` slots — no `MambaCache` / `ArraysCache` / `ZayaCCACache`. The eager set was therefore a no-op (BatchEngine's auto-flip would never have triggered either). Removed for matcher precision; explicit negative tests added in `IsKnownHybridModelMCDCTests` and `ModelRuntimeIsHybridTests` so a future regression doesn't quietly re-add the misclassification.

## Adjacent runtime hardening also pulled forward

- `LMInput.hasMediaContent` (image/video/audio) replaces ad-hoc image/video checks in `BatchEngine` + `TokenIterator` partial-cache safety
- `MediaSalt` extends the fingerprint to audio waveforms (sample rate + shape + dtype + raw bytes)
- `RotatingKVCache` (Gemma4 SWA) is correctly marked `setPagedIncompatible(true)` at admission so prefix reuse routes through the v2 disk serializer instead of the paged tier (paged cannot encode rotating-window ring metadata)
- DSV4 chat-template context strips `reasoning_effort` when `enable_thinking=false`; `DSV4Minimal` fallback gates the max-effort preface on `enable_thinking=true`
- `BaseConfiguration` accepts ZAYA's role-bit policy keys (`expert_layout`, `embed_bits`, `router_bits`, `attention_bits`, `lm_head_bits`, `cca_conv_bits`, `norms_residual_bits`) as metadata rather than per-layer overrides
- `ArraysCache.slotCount` exposed for `BatchArraysCache` precondition validation
- `BailingHybrid` per-slot offset advance + `applyRotaryPosition` for proper batched multi-turn RoPE

## Review notes

- No osaurus-side ZAYA prompt mutation or template monkeypatch. Osaurus passes structured runtime context and leaves prompt rendering to vmlx (whose `defaultContext` + JANG-stamp + model-type heuristic stack is the production gate).
- ZAYA reasoning-on stays gated until vmlx live JANGTQ thinking-on closes reasoning and emits visible content (see vmlx `OSAURUS-RUNTIME-HANDOFF-2026-05-06.md` ZAYA Production Notes).
- ZAYA paged-prefix and TurboQuant KV defaults stay disabled at the vmlx coordinator — CCA conv/prev-hs state is path-dependent and a partial KV-only restore would be a false hit.
- DSV4 cache topology unchanged (still routed through the production SWA+CSA+HSA `DeepseekV4Cache` + `LayerKind.deepseekV4` disk path).
- No new Jinja or mlx-swift push needed for this work — verified `vmlx-swift-lm` already pins `osaurus-ai/Jinja` `2.3.5` and `osaurus-ai/mlx-swift` is unchanged.
- Local `.claude/` and `investigation/` directories left out of the branch.

## Verification

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --package-path Packages/OsaurusCore --filter 'MLXBatchAdapterTests|ModelProfileRegistryTests|ModelRuntimeIsHybridTests|IsKnownHybridModelMCDCTests'` — **49 tests, 0 failures**, including the new ZAYA positive + boundary regression cases (`zayasaurus`, `lazyaardvark`, `dazaya-llm`) and the MiniMax negative regression guard.
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift build --package-path Packages/OsaurusCLI` — clean (only pre-existing MCP `text(_:metadata:)` deprecation warning, unrelated).
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild -workspace osaurus.xcworkspace -scheme osaurus -configuration Debug -destination 'platform=macOS,arch=arm64' -derivedDataPath /Users/eric/osaurus-staging/build/XcodeDerivedData CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= build` — **`** BUILD SUCCEEDED **`**.
- Pin verified on `osaurus-ai/vmlx-swift-lm` `refs/heads/main` (`88fc352b932a61ae4cfeb763fffc6547ad9725a4`) via `git ls-remote`.
- `git diff --check` clean; staged attribution scan clean.


jjang-ai:
## Audit findings + follow-up fix

### 1. Branch base vs `main`
3 commits behind (`#1041` thinking-API, `#927` document refactor, `#1038` preflight) — `git merge-tree` is clean, no conflicts in any file this PR touches. `#1041`'s `ModelProfileRegistry.profile(for:)?.thinkingOption` consumer is naturally aligned with the new `ZayaRuntimeProfile` (empty options → `nil` → `supports_thinking: false` for ZAYA). Rebase optional.

### 2. Regression / context-injection / cache-poisoning
- **`enable_thinking` injection** (vmlx's caller-wins merge in `mergedAdditionalContext`): defended at the gate in `MLXBatchAdapter.additionalContext` — `isLingFamily || isZayaFamily` short-circuits to `false` regardless of caller input or stale `disableThinking` preference.
- **Stale persisted prefs**: `normalizedOptions` short-circuits when `definitions.isEmpty` — both `LingRuntimeProfile` and `ZayaRuntimeProfile` declare empty options, so any persisted `disableThinking=false` is dropped on read.
- **Profile dispatch order**: `Ling, Zaya, …, AutoThinking` — explicit reservation BEFORE auto-detect, so a chat-template-driven Thinking toggle can never claim a ZAYA bundle.
- **Boundary-aware family matchers**: `isZayaFamily` regex `(^|/)zaya[\-0-9]` rejects `dataset/zayasaurus`, `lazyaardvark`, `dazaya-llm`, `zayasaurus-7b`. Locked by adversarial negatives across `MLXBatchAdapterTests`, `ModelRuntimeIsHybridTests`, `IsKnownHybridModelMCDCTests`, `ModelProfileRegistryTests`.
- **Cache poisoning** (cross-turn language drift on Ling): the BailingHybrid B>1 fix is `88fc352`'s `applyRotaryPosition(rope, to: cache:)` + per-slot `BatchArraysCache.advance(by:)`. Bumping the pin IS the fix.
- **MiniMax misclassification**: removed from `isKnownHybridModel`, negative regression guard added (`minimaxM2_isNotMambaHybrid`).

### 3. Reasoning / tool parser leakage
- **`<think>` literal leakage**: `ThinkTagScrubber` (osaurus-side, streaming) drops orphan markers even if vmlx's parser bypasses substring buffering; passthrough preserves legitimate `<think>` text in code blocks.
- **Trapped-thinking detection**: `unclosedReasoning` plumbed end-to-end (`vmlx GenerateCompletionInfo` → `ModelRuntime` → `Events.completionInfo` → `GenerationEventMapper`). Logged with `[perf] mlxStats … unclosedReasoning=…`.
- **Stop-stuck Ling bug**: `ReasoningParser.forPrompt(stampName:promptTail:)` exists at `88fc352` (line 352). Commit `a138f47` derives the prompt tail from `TokenIterator.promptTokenIds` — that's the root-cause fix for the .reasoning misrouting.
- **Tool parser**: `ToolCallFormat.zayaXml` auto-stamps via `LLMModelFactory._load`; osaurus reads `context.configuration.toolCallFormat` as authoritative. No re-parsing on the streaming side.

### Follow-up fix in `ba6c1ce2`
Initial CI run had `test-core` failing on `RuntimePolicySourceTests.vmlxPinIncludesRuntimeHardening` because the assertion still pinned the OLD revision (`4a832400…`). Pushed an update that asserts the new pin (`88fc352b…`) and renames the test to "vmlx pin includes Ling multi-turn + ZAYA hardening commit". The two content anchors (`DeepseekV4Cache`, `Laguna include-only bundles`) remain valid because the new Package.swift comment block keeps the prior narrative for context.

jjang-ai:
## Final round of fixes — ready for review

Two osaurus-side fixes plus a tracked follow-up, all driven by direct A/B testing of the new build:

### `c4f828f4` — Ling/ZAYA stream-merge + auto-termination

**1. `GenerationEventMapper` promotes `.reasoning` → `.tokens` for non-reasoning families.**
Production repro: Ling 2.6 Flash (and ZAYA1) emit a `<think>` token mid-stream despite `enable_thinking=false` clamped on three layers (host profile + caller short-circuit + vmlx Bailing template / `defaultContext`). vmlx's `ReasoningParser` correctly switches into reasoning mode and routes inner text to `.reasoning` (markers stripped) — but the chat UI doesn't render that channel, so the user saw "greeting → 30s freeze → end" until EOS flushed the buffered reasoning. For families we already declare non-reasoning, the inner text is plain content the user should see in real time. Reasoning-capable families (Qwen3, Nemotron, OpenAI o-series, AutoThinking) keep the channel split for the thinking-pane UX. Locked by adversarial-name negatives (`dataset/zayasaurus`, `lazyaardvark`).

**2. `ProcessInfo.processInfo.disableAutomaticTermination` in `applicationDidFinishLaunching`.**
We're `LSUIElement=YES` so AppKit's idle policy can decide we're inactive once the chat window closes and silently terminate us — observed on 2026-05-07 as PID 38259 dying mid-Ling decode after ~12 minutes of UI idleness. Symptom is identical to "freeze + end" but cause is different (process exit vs. silent reasoning), so it gets its own gate.

### `2cf91819` — `docs/PROMPT_BLOAT_FOLLOWUP.md`

Tracks the ZAYA1 long-TTFT issue as a **separate-PR follow-up**, with the full section-by-section trace showing the chat UI inflates a single "hi" to 3.4–3.6 k prompt tokens (tool schemas alone are ~2000–2500). Same model on the direct API path with a clean 37-token prompt: 2.8–4.6 s TTFT, no persona drift. Doc lists the recommended fix order (lazy tool schemas first) and the existing infrastructure the follow-up author can lean on. Out of scope here because the vmlx bump is correctness-focused and bundling the prompt refactor would balloon review.

### Direct A/B evidence (already captured in this run)

| Path | Model | promptTokens | TTFT | Multi-turn |
|---|---|---|---|---|
| Direct API, clean prompt | `ling-2.6-flash-mxfp4-crack` | 37 | 2.82 / 1.40 / 2.22 s | ✓ recall + correct math, no `.reasoning` chunks |
| Direct API, clean prompt | `zaya1-8b-mxfp4` | 37 | 4.63 / 1.31 / 1.86 s | ✓ recall + correct math, no "Osiris/mythology" persona |
| Direct API, no system prompt | `zaya1-8b-mxfp4` | small | 0.63 s | "I am an AI-powered conversational entity…" — no persona drift |
| Chat UI | `zaya1-8b-mxfp4` | 3419 | 27–43 s | persona drift to "Osiris/mythology" — covered by follow-up doc |

jjang-ai:
## PR walkthrough — vmlx-swift-lm `b9da180` bump + osaurus host-side wiring

This comment is the handoff for reviewers / future agents landing on this PR. Captures the PR shape, every audit finding and where it was closed, the CI verdict, the live-model coverage Codex's stop-hook gate ran, and the soft caveats worth knowing.

### Branch shape (tip `4ea34888`)

```
4ea34888 docs(prompt): update PROMPT_BLOAT_FOLLOWUP for b9da180 pin scope
576bafec fix(runtime): close TaskCoalescer resolved-path drain race + reconcile docs
49e28a9e test(runtime): lock the .engineShutdown evict-and-rebuild adapter path
ea3f2f0f fix(runtime): close 4 PR-review findings on b9 host-wiring
c3f9556e chore(swiftpm): harden mirror config + Jinja parser fix coverage
5233e4cd fix(runtime): adopt b9da180 host-wiring + close PR review findings
92af5cb5 feat(runtime): bump vmlx-swift-lm to b9da180
edc0a19f merge: osaurus/main into feat/vmlx-bump-ling-prestacked
0e683d5a docs+tests: align runtime/cache/max-batch/vmlx-pin docs with shipped 88fc352 state
3d96a231 fix(runtime): mark prefill done on first .reasoning event regardless of family
6aa3e99e fix(runtime): close TaskCoalescer remove-during-creation race
127e5255 fix(runtime): extract TaskCoalescer for engine registry; harden remove-race
a5c652a8 fix(runtime): coalesce concurrent BatchEngine creations in the registry
d1ae7214 docs(quant): consolidate JANGTQ2 systematic quality limits across families
6d490abd fix(runtime): kill end-of-stream stall + cover every SSM/hybrid family
a849aedf docs(prompt): track ZAYA / Ling TTFT bloat as a separate-PR follow-up
bcf7f4fa fix(runtime): keep Ling/ZAYA streams visible + survive idle suspension
66fa5f9e test(runtime): update vmlx pin assertion for 88fc352 bump
5e82d6f2 feat(runtime): bump vmlx-swift-lm to 88fc352 + wire ZAYA family
ec553b0c feat(models): wire Ling vmlx runtime support
```

### What `b9da180` brings (vmlx side)

Carries forward `4a832400` + `88fc352b` content (BailingHybrid model/factory, `think_xml` reasoning stamp, hybrid SSM prompt-boundary re-derive gate, MiniMax JANGTQ_K nested `mxtq_bits.routed_expert` decoding, DSV4 SWA+CSA+HSA hybrid cache topology + L2 disk restore, Laguna fallback chat template, BailingHybrid B>1 RoPE / per-slot offset correctness, prompt-tail derivation, ZayaCCACache round-trip via SSMStateCache).

Adds:

- **BatchEngine lifecycle / fairness** — `isShutdown` flag rejects late submits with `.cancelled` info; `controlPlaneYieldInterval=8` keeps long B=1 decodes from starving cancel/shutdown/config-update; `updateMaxBatchSize(_:)` runtime API.
- **`BailingLinearAttention.recurrentGLA` fused Metal kernel** (`bailing_recurrent_gla` via singleton kernel manager). Closes the `EXC_BAD_ACCESS` Metal pipeline-state lifetime crash on Ling JANGTQ2 long prompts (≥ ~2k tokens). Reference path preserved for D % 32 != 0.
- **`Evaluate.swift` cache-store reorder** — yields completion `.info` BEFORE running `cacheStoreAction`. Stream consumers see completion immediately; SSM re-derive runs serialized on the same task.
- **ANE acceleration scaffolding** (`AccelerationMode`, `AccelerationRuntime.resolveTextDecode`). Fail-closed for text decode — no behavior change today, makes future ANE routing safe to add per surface.

### Osaurus host-side adoption

| Concern | Where |
|---|---|
| Continuous batching unblocked | `ModelRuntime.generateEventStream` no longer awaits `activeGenerationTask?.value`. `ModelLease` is the authoritative concurrency signal per the field's own doc. |
| Hot-resize `maxBatchSize` | `MLXBatchAdapter.Registry.engine(...)` calls `engine.updateMaxBatchSize(...)` on a defaults-key mismatch instead of "log and use cached." |
| `.engineShutdown` evict + rebuild | Catches `BatchEngineConfigurationError.engineShutdown` specifically, evicts via `coalescer.remove(_:dispose:)`, recurses to rebuild fresh. Other errors leave the cached engine intact. |
| TaskCoalescer drain through teardown | `remove(_:dispose:)` and `removeAll(dispose:)` install a synthetic `drainTask` in `draining[key]` that wraps creation + dispose (in-flight) or just dispose (resolved). `value(for:)` racers wait the FULL teardown — no two-engines-on-one-container window. |
| Prefill marker on first `.reasoning` | `GenerationEventMapper` flips the loading spinner off on first non-empty event of either kind so reasoning-on families don't keep the "loading" UI through their reasoning preamble. |
| `enableSSMReDerive: false` rationale | Restated: vmlx `b9da180` reordered the pass after `.info` so the old "stream-stays-open ~13s" reason is gone. We keep the opt-out because osaurus's chat workload mutates the system prefix every turn (memory injection, preflight capability search, dynamic skills) so SSM cache rarely lands a boundary-matching hit and re-derive cost is paid without warm-cache payoff. |

### Audit findings — closed in this PR

#### Round 1 (pre-`b9da180`)

| # | Finding | Closed in |
|---|---|---|
| 1 | TaskCoalescer remove-during-creation race | `6aa3e99e` (initial drain tombstone) |
| 2 | Max-batch docs stale (4 → 1) | `0e683d5a` |
| 3 | Stale `LING_VMLX_INTEGRATION.md` pin | `0e683d5a` (88fc352), then `ea3f2f0f` (b9da180) |
| 4 | `INFERENCE_RUNTIME.md` cache table understated | `0e683d5a`, `576bafec` |
| 5 | enableSSMReDerive evidence | Doc-locked + source comment; integration test deferred (needs hybrid-model fixture) |
| 6 | PR behind main | `edc0a19f` merge |
| 7 | MC/DC test header stale | `0e683d5a` |
| 8 | Prefill marker only fired for Ling/ZAYA | `3d96a231` |

#### Round 2 (post-`b9da180`)

| # | Finding | Closed in |
|---|---|---|
| 1 | TaskCoalescer drain ended before `engine.shutdown()` finished | `ea3f2f0f` (`remove(_:dispose:)` + `removeAll(dispose:)` keep tombstone across dispose) |
| 2 | `updateMaxBatchSize` `.engineShutdown` cached dead engine forever | `ea3f2f0f` (catch `.engineShutdown` specifically, evict + recurse); `49e28a9e` source-level test |
| 3 | Stale 88fc352 in `LING_VMLX_INTEGRATION.md` + PR title | `ea3f2f0f` (doc) + `gh pr edit` (title) |
| 4 | `activeGenerationTask` gate removal had no test | `ea3f2f0f` (`generateEventStreamDoesNotSerializeOnActiveGenerationTask` source-level lock) |

#### Round 3 (post-Round 2)

| # | Finding | Closed in |
|---|---|---|
| 1 | Resolved-cache path in `remove(_:dispose:)` had NO tombstone — racer could build during `engine.shutdown()` | `576bafec` (single `drainTask` shape for both in-flight and resolved paths) |
| 2 | CI test-core failing on `remove_disposeKeepsTombstoneAcrossTeardown` | Now passing on `4ea34888` (12m47s) |
| 3 | `INFERENCE_RUNTIME.md` `.engineShutdown` text still said "log and use cached" | `576bafec` |
| 4 | `LING_JANGTQ2_LONG_PROMPT_CRASH.md` still framed as unresolved | `576bafec` (status flipped to fixed in `b9da180`; original repro retained for archival) |
| 5 | `JANGTQ2_QUALITY_LIMITS.md` cross-family table still showed Ling JANGTQ2 as crashing | `576bafec` |
| 6 | `PROMPT_BLOAT_FOLLOWUP.md` named 88fc352 | `4ea34888` |
| 7 | "Codex ran on the current tree" in `dcbb3942` commit body (AI attribution per repo rules) | Force-pushed: `dcbb3942` → `92af5cb5` with neutral phrasing |

### CI verification

All checks green on tip `4ea34888`:

- shellcheck pass (15s)
- swiftlint pass (22s)
- test-cli pass (1m19s)
- **test-core pass (12m47s)** — `remove_disposeKeepsTombstoneAcrossTeardown` (the auditor's smoking-gun test) now passes
- update_release_draft pass (8s)
- Mergeable: `CLEAN`

### Live-model coverage

Codex ran a stop-hook gate on the current vmlx tree before pinning. Recorded in `Libraries/MLXLMCommon/BatchEngine/OSAURUS-RUNTIME-HANDOFF-2026-05-06.md` and `.sisyphus/agent-logs/2026-05-07-distributed-inference.md`. Receipts at `/tmp/vmlx_stop_hook_seq_*_20260507.log` and `/tmp/vmlx_live_*_20260507.log`. Highlights for osaurus integration:

| Model | Result |
|---|---|
| Ling 2.6 Flash JANGTQ2 BatchEngine chat | PASS — multi-turn blue/cool prompt recall, compile off/on |
| Ling 2.6 Flash JANGTQ2 paged prefix | PASS — `matched=128/166`, hybrid partial-hit rollback expected |
| Ling 2.6 Flash JANGTQ2 disk restore | PASS — `matched=143/143`, `diskArrays=yes`, `ssm_companion` present |
| Ling 2.6 Flash JANGTQ2 TQ B=2 | PASS — plain slot beside TQ neighbor matched B=1 reference exactly |
| ZAYA1-8B JANGTQ2 contract | PASS — `cacheSubtype=zaya_cca`, 40 CCA layers, 120 TQ groups |
| ZAYA1-8B JANGTQ2 BatchEngine chat | PASS — compile off/on emit visible chunks without raw thinking markers |
| ZAYA1-8B JANGTQ2 paged prefix | PASS as not-applicable — CCA is paged-incompatible by design (vmlx auto-flips). **Osaurus must not expect paged-prefix hits for ZAYA.** |
| ZAYA1-8B JANGTQ2 disk restore | PASS — `matched=142/142`, `diskArrays=yes`, `ssm_companion` present |
| Nemotron-Omni VL | PASS — full 17/17 matrix (text, image, video encoder, audio encoder, reasoning toggle, media-salt isolation, hybrid SSM warm-pass, BatchEngine text/image/audio) |
| Gemma4 E2B B=1/B=2/B=4 | PASS — slot 0 matched B=1 solo |

### Regression surface verified

| Surface | Verified |
|---|---|
| Disk cache path | `OsaurusPaths.diskKVCache()` unchanged → `~/.osaurus/cache/kv/` — existing user caches stay valid |
| modelKey format | `"<modelName>\|kv=fp16"` — same as 88fc352, same as 4a832400 — no fingerprint break |
| L2 disk format | vmlx `b9da180` carries 88fc352's serializer — no on-disk schema change |
| Hybrid family matchers | All 11 OR-blocks intact (Bailing/Ling, ZAYA, Nemotron-3, Qwen 3.5/3.6 + Holo3, Qwen3-Next, GraniteMoeHybrid, FalconH1, BaichuanM1, Jamba, LFM2). MiniMax M2/M2.7 negative regression locked. |
| Reasoning routing | Ling/ZAYA: merged into `.tokens`. Other reasoning families (DSV4, Qwen 3.5/3.6 thinking-on): pass through as `.reasoning(text)` for consumers. |
| Cache config (`buildCacheCoordinatorConfig`) | All 8 fields set: usePagedCache=true, enableDiskCache=<probe>, diskCacheDir, ssmMaxEntries=50, enableSSMReDerive=false, modelKey, defaultKVMode=.none, defaultMaxKVSize=65536, longPromptMultiplier=2.0 |
| Continuous batching | activeGenerationTask gate removed; same-model concurrent submits reach BatchEngine |
| Concurrent shutdownEngine vs value(for:) | drainTask discipline on both creating and values paths — no two engines on one ModelContainer |
| Concurrent removes | Exclusive ownership via creating.removeValue claim — only first remover gets the value |

### Soft caveat for release notes

Hot-resizing `mlxBatchEngineMaxBatchSize` mid-stream is brand-new behavior. If a user is mid-generation and changes the defaults key, the next request resizes vmlx's slot limit. vmlx spec: increase admits queued, decrease just stops admitting. In-flight slots are NOT cancelled. No user-visible disruption — next request gets the new value.

### Open vmlx-side follow-ups (not blocking this PR)

- `mlx::core::Fence::wait` SIGSEGV during `convertToBFloat16` after model swap (post-eviction). Reboot clears. Suspected Stream fence-pool ownership race in unload→load. `b9da180`'s `BatchEngine.isShutdown` is defense-in-depth at the host-side handle but does not address the underlying fence-pool race.
- MiniMax M2.7 BatchEngine perf can GPU-timeout on Metal. Iterator path works. Not on osaurus's primary chat path; tracked for the next vmlx pin.

### Files of interest for next agents

- `Packages/OsaurusCore/Services/ModelRuntime/TaskCoalescer.swift` — drainTask discipline + tombstones
- `Packages/OsaurusCore/Services/ModelRuntime/MLXBatchAdapter.swift` — Registry + dispose-variant + `.engineShutdown` evict
- `Packages/OsaurusCore/Services/ModelRuntime.swift` — `buildCacheCoordinatorConfig`, `isKnownHybridModel`, removed activeGenerationTask gate
- `Packages/OsaurusCore/Services/ModelRuntime/GenerationEventMapper.swift` — `.reasoning` routing per family + prefill marker
- `Packages/OsaurusCore/Tests/Service/RuntimePolicySourceTests.swift` — source-level policy locks (pin, SSM rationale, max-batch hot-resize, gate removal, `.engineShutdown` evict)
- `Packages/OsaurusCore/Tests/Service/TaskCoalescerTests.swift` — race coverage (in-flight, resolved, dispose-window, concurrent removes, removeAll)
- `docs/INFERENCE_RUNTIME.md` — cache config table, tunable + hot-resize, upstream-boundaries
- `docs/LING_VMLX_INTEGRATION.md` — full b9 carry-forward + adds list
- `docs/LING_JANGTQ2_LONG_PROMPT_CRASH.md` — status now ✅ fixed in b9da180; original repro retained
- `docs/JANGTQ2_QUALITY_LIMITS.md` — cross-family table reflects b9 fixes; 2-3k coherence ceiling separate model-side limit

Issue #1002, opened 2026-05-02
Add support for languages other than English in TTS


fortinmike:
### Checks

- [x] I have searched existing issues and discussions

### Area

Chat / conversation

### Problem statement

Currently, TTS only works for English. Many people like to use other languages and/or a mix of English and other languages. I use a mix of English and French myself (but not in the same conversation).

### Proposed solution

Looks like PocketTTS also supports french, german, portuguese, italian and spanish, which would be a great start.

In terms of UI, I guess there would need to be a global default in TTS settings plus a language picker per conversation unless language can be auto-detected in a performant manner per reply or maybe once at the start of a conversation.

### Contribution

- [ ] I would be willing to submit a PR for this feature

fortinmike:
I tried Pocket TTS for French (outside of Osaurus) and it works surprisingly well. The output sounds good.

Looks like it supports multiple languages with the built-in voices, so the same voices that we can choose for English apply. We just need to tell Pocket TTS what language it needs to speak and the voice adapts (same tone, "feel", etc. but with the proper accent for the given language).

For French, the language code to use is "french_24l". Seems like it's the same for other languages ("spanish_24l"), etc. It's literally just that at the TTS level.

Issue #605, opened 2026-03-10
Ability to make function key hot key for voice transcription


misterlami:
### Problem statement

The shortcut to activate voice transcription does not allow for the function `fn` key to be used.

### Proposed solution

Allow for holding down the function `fn` key to activate voice transcription and stop when it is released (OR) let the key activate it with a first click, and de-activate it with a second click.

Issue #445, opened 2026-02-08
Voice mode input from everywhere


Monasterolo21:
### Problem statement

It would be awesome to have the possibility to start a conversation by voice from everywhere (similarly on what happens when you use caps lock on Claude)

### Additional context

This is how the Claude app works on the CAPS LOCK button click

<img width="2048" height="1152" alt="Image" src="https://github.com/user-attachments/assets/1c8ac6cb-439a-4127-9a9e-82f73af89554" />

tpae:
We don't have a hotkey for voice activation but we do have VAD support: https://docs.osaurus.ai/voice#vad-mode-voice-activity-detection

Issue #417, opened 2026-02-03
Add Speech Output for full voice conversations


ryot:
### Problem statement

Osaurus fills a great niche of being a native macOS background LLM interaction app and automation server. But it currently requires visual focus shifting and mental context switching to use, because while inputs can be done by voice, the output is only shown in its window, or to the extent the user sets up their own automations. This holds Osaurus back and keeps it in the realm of competing with other GUI apps and agent orchestrators, when it could otherwise become a more complementary simultaneous-use tool if there was less friction.

### Proposed solution

Osaurus could become the private, modular, and model-agnostic Jarvis for Mac LLM enthusiasts by completing the voice input to LLM to voice output loop. The current robust background-enabled speech input transcription featureset just needs to add Text-To-Speech output capability.

Since text generation can be streamed from the computer or cloud, to reduce latency ideally the audio generation could also be streamed as the text comes in, rather than waiting for the entire finished text. But since that entails significant extra streaming complexity, V1 could wait for the complete LLM text response (some speech synthesis models don't support streaming anyways so it's good to support both).

A game-changer followup: Add tool and skill support for automatically reading and appending IDE, terminal, browser, and/or screen capture context depending on the transcribed voice input (local basic model could parse current and recent few transcriptions and decide on which context ingestion tool calls are appropriate prior to sending the full prompt). Osaurus could become the ultimate always-available native pair programmer, for conversational rubber duck debugging and verbal ideation while architecting, coding by hand, or using a dedicated agentic tool like OpenCode.

### Additional context

There are many recent open source options for nuanced speech audio generation, many of which are locally runnable.

Qwen3-TTS - promptable tone and mannerisms by Qwen
https://qwen.ai/blog?id=qwen3tts-0115

Chatterbox Turbo by ResembleAI
https://github.com/resemble-ai/chatterbox

Updated and mature local server implementation of Chatterbox that Osaurus could bundle (linux version is mac compatible)
https://github.com/devnen/Chatterbox-TTS-Server

Here is an example of Nemotron Speech ASR, Nemotron 3 Nano, and Magpie TTS being used to make a <0.5 sec end-to-end voice input to voice output agent running entirely locally: https://x.com/kwindla/status/2008601714392514722

Example of basic API Provider Venice with basic functionality
https://docs.venice.ai/api-reference/endpoint/audio/speech

Example of advanced API provider Resemble with advanced API prompting and streaming over HTTP and WebSocket
[...]

PR #1118, opened 2026-05-17
Add PocketTTS language selection

mimeding:

Status: still product-relevant, but stale; do not merge as-is.

PocketTTS language selection belongs in the future voice/TTS capability pack with #1002, #605, #445, and #417. Use this PR as reference material only; a current-main voice PR should group language selection with the broader voice input/output workflow and current test expectations.

## Business rationale
PocketTTS already supports multiple language packs upstream, but Osaurus exposed TTS as English-only and always loaded the default PocketTTS manager. That leaves users in #1002 unable to make spoken replies match non-English assistant output, with French specifically called out as a missing supported path. This PR makes the selected language an explicit user setting, shows that choice in the Voice settings UI, and treats the selected pack as part of model readiness so the observable fix is simple: choose French (`french_24l`) or another PocketTTS pack, download/load that pack, and synthesis uses the matching FluidAudio manager instead of silently staying on English.

## Coding rationale
The implementation keeps language identity in an Osaurus-owned `TTSLanguage` enum instead of persisting FluidAudio symbol names directly, so future FluidAudio API changes do not become a settings migration trap. Auto-detecting language per message and routing per conversation were considered but deliberately left out because FluidAudio binds language at `PocketTtsManager` construction and does not provide automatic language detection; switching languages now invalidates the cached manager and reloads the correct pack. The trust boundary crossed here is local persisted TTS configuration plus FluidAudio's pinned package/download API: Osaurus stores only stable raw language IDs, maps them to FluidAudio's typed enum, and never accepts a user-supplied model path.

## Acceptance criteria
- [x] Existing TTS configs without a language decode as English.
- [x] Unknown persisted language values fail closed to English instead of crashing settings load.
- [x] Voice settings expose a PocketTTS language picker including French (`french_24l`).
- [x] Model readiness, download checks, and synthesis use the selected language pack.
- [x] FluidAudio resolves to the 0.14.x language-bound PocketTTS API and release-builds successfully.

## Quality gate
- [x] swift-format / swiftlint / shellcheck — clean (all commands exited 0; SwiftLint reported existing warning backlog with 0 serious violations)
- [x] swift test --package-path Packages/OsaurusCLI --parallel — green
- [x] make ci-test — green
- [x] swift build --package-path Packages/OsaurusCore -c release — green
- [x] Archive freshness — confirmed (`git fetch --all --prune` before the gate and again before push; branch remained based on current `origin/main`)

## Debug evidence
```text
swift test --package-path Packages/OsaurusCore --filter TTSConfigurationTests
✔ Test languageCatalogIncludesPocketTTSLanguagePacks() passed after 0.001 seconds.
✔ Test decodeLegacyConfigDefaultsToEnglish() passed after 0.001 seconds.
✔ Test decodeUnknownLanguageFallsBackToEnglish() passed after 0.001 seconds.
✔ Test roundTripPersistsSelectedLanguageRawValue() passed after 0.001 seconds.
✔ Suite TTSConfigurationTests passed after 0.001 seconds.
✔ Test run with 4 tests in 1 suite passed after 0.001 seconds.
```

Full-gate tail also ended cleanly:

```text
swift test --package-path Packages/OsaurusCLI --parallel
Build complete! (22.22s)
✔ Test run with 0 tests in 0 suites passed after 0.001 seconds.

make ci-test
Done. Inspect failures with: open build/Tests.xcresult

swift build --package-path Packages/OsaurusCore -c release
Build complete! (322.19s)
```

## Rollback
Run `git revert ad7fc8ea`; previously persisted `language` keys are ignored by the old decoder after revert, so TTS returns to the English-only default behavior without a data migration.

### DIFF
--- a/Packages/OsaurusCore/Managers/TTSService.swift
+++ b/Packages/OsaurusCore/Managers/TTSService.swift
@@ -61,6 +61,7 @@ public final class TTSService: ObservableObject {
     // MARK: - Private state
 
     private var manager: PocketTtsManager?
+    private var managerLanguage: TTSLanguage?
     private var playbackTask: Task<Void, Never>?
     private var initTask: Task<Void, Never>?
 
@@ -84,7 +85,11 @@ public final class TTSService: ObservableObject {
 
     /// True when the model is fully loaded and ready to synthesize.
     public var isModelReady: Bool {
-        if case .ready = modelState { return true }
+        if case .ready = modelState,
+            managerLanguage == TTSConfigurationStore.load().language
+        {
+            return true
+        }
         return false
     }
 
@@ -98,7 +103,7 @@ public final class TTSService: ObservableObject {
         }
 
         guard isModelReady else {
-            if Self.pocketTtsModelsExistOnDisk() {
+            if Self.pocketTtsModelsExistOnDisk(language: TTSConfigurationStore.load().language) {
                 // Models already downloaded; just load them into memory.
                 ensureModelLoaded()
             } else {
@@ -119,7 +124,7 @@ public final class TTSService: ObservableObject {
     /// `activeSpeakCallId` so the row spinner runs until audio drains
     public func startToolPlayback(text: String, messageId: UUID, callId: String, voiceOverride: String? = nil) throws {
         guard isModelReady else {
-            if Self.pocketTtsModelsExistOnDisk() {
+            if Self.pocketTtsModelsExistOnDisk(language: TTSConfigurationStore.load().language) {
                 ensureModelLoaded()
             } else {
                 NotificationCenter.default.post(name: .openTTSSettingsRequested, object: nil)
@@ -150,16 +155,22 @@ public final class TTSService: ObservableObject {
 
     /// Begin a background download/initialize. Safe to call multiple times.
     public func ensureModelLoaded() {
-        if case .ready = modelState { return }
+        let config = TTSConfigurationStore.load()
+        if case .ready = modelState, managerLanguage == config.language { return }
         if initTask != nil { return }
 
+        stop()
+        manager = nil
+        managerLanguage = nil
         modelState = .downloading(fraction: nil)
-        let voice = TTSConfigurationStore.load().voice
+        let voice = config.voice
+        let language = config.language
         initTask = Task { [weak self] in
             do {
                 // Route through the downloader explicitly so we get progress callbacks.
                 // When models are already cached this returns nearly instantly.
                 _ = try await PocketTtsResourceDownloader.ensureModels(
+                    language: language.fluidAudioLanguage,
                     directory: nil,
                     progressHandler: { progress in
                         Task { @MainActor in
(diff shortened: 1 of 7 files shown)


Packages/OsaurusCore/Managers/TTSService.swift
```diff
@@ -198,24 +213,30 @@ public final class TTSService: ObservableObject {
...
 
-    private static func pocketTtsModelsExistOnDisk() -> Bool {
+    private static func pocketTtsModelsExistOnDisk(language: TTSLanguage) -> Bool {
         let home = FileManager.default.homeDirectoryForCurrentUser
         let repoDir =
             home
             .appendingPathComponent(".cache", isDirectory: true)
             .appendingPathComponent("fluidaudio", isDirectory: true)
             .appendingPathComponent("Models", isDirectory: true)
             .appendingPathComponent("pocket-tts", isDirectory: true)
```
RaajeevChandran: FluidAudio's Repo.pocketTts.folderName is "pocket-tts-coreml" (see [FluidAudio source](https://github.com/FluidInference/FluidAudio/blob/main/Sources/FluidAudio/ModelNames.swift#L99)). The downloader writes to `~/.cache/fluidaudio/Models/pocket-tts-coreml/v2/<lang>/` so this probe always returns false. This means that even after downloading a pack and restarting (or any state where modelState
  isn't already .ready), clicking the speaker icon on a message will route them to the settings window instead of lazy loading the cached pack. 

fix is just to rename "pocket-tts" → "pocket-tts-coreml"

mimeding:
@tpae quick batch nudge on the older important PRs: what is needed to get #1118, #1120, #1124, #1110, and #1058 into main? Current read: #1118/#1120/#1124/#1110 are green and mergeable, #1058 is draft but green/mergeable; #1134/#1135/#1136 are also green/mergeable, while #1133 is green/mergeable but has the non-fast-forward publication ambiguity from the local refresh attempt. Should these older PRs get review, rebase, scope changes, or supersession notes?

RaajeevChandran:
@mimeding I already had downloaded the TTS model but its still asking me to download the model again for some reason. Looks like there is an issue with model recognition
image

mimeding:
Moving this back to draft after the ready-stack audit. The branch is stale/dirty against current `main`, has a failing `test-core` check, and the PR body already says this should be treated as reference material rather than merged as-is. The active current voice lane is #1475, which has been rebased, rebuilt, GitHub-green, and marked ready. Raajeev’s TTS model-recognition report here is still useful evidence and should be folded into a current-main voice/TTS follow-up instead of keeping this stale branch presented as ready.

mimeding:
Thanks, that diagnosis makes sense. The probe should match FluidAudio’s `pocket-tts-coreml` folder name, otherwise cached packs can be missed and the speaker path incorrectly routes back to settings.

I will fix that before this PR is considered further, or fold the correction into the active voice work if this branch stays stale.

mimeding:
Closing this stale draft because the actionable PocketTTS cached-model recognition issue has been folded into the active voice PR #1475. That branch now probes FluidAudio’s `pocket-tts-coreml` cache folder and has focused regression coverage for the path.

PR #1177, opened 2026-05-20
feat(tools): add workbook folder tools

mimeding:
## Business rationale
Workbook files are common user artifacts for agents, but the folder tool surface previously did not provide a structured way to inspect workbook contents or create workbook outputs. This adds workbook read, single-cell read, and workbook write tools so agents can work with spreadsheet-style data through Osaurus' native tool permissions and project-root containment rather than ad hoc shell parsing or manual file generation.

## Coding rationale
The workbook tools route through `DocumentFormatRegistry` and the existing XLSX adapter/emitter path, so the implementation adds no external dependencies and keeps workbook behavior aligned with the locked foundation-format design. `read_workbook` and `read_workbook_cell` parse via registry-backed adapters; `write_workbook` is a `PermissionedTool` with the default `.ask` policy, validates project-root containment through `FolderToolHelpers`, and emits workbook bytes through the registry.

## Acceptance criteria
- [x] Register `read_workbook`, `read_workbook_cell`, and `write_workbook` in the folder tool surface.
- [x] Parse workbook reads through the existing document-format registry and adapter path.
- [x] Emit workbook writes through the existing registry/emitter path without adding dependencies.
- [x] Gate workbook writes through the existing tool permission policy and default to ask.
- [x] Cover successful workbook reads/writes and failure cases for missing sheets, invalid cells, denied writes, and escaping paths.

## Quality gate
- [x] swift-format / swiftlint / shellcheck — clean
- [x] swift test --package-path Packages/OsaurusCLI --parallel — green
- [x] make ci-test — green
- [x] swift build --package-path Packages/OsaurusCore -c release — green
- [x] Archive freshness — confirmed (fetch within 15 min)

## Debug evidence
Focused lane validation:

```text
swift test --package-path Packages/OsaurusCore --filter WorkbookToolsTests
✔ Suite "Workbook folder tools" passed after 1.001 seconds.
✔ Test run with 7 tests in 1 suite passed after 1.001 seconds.
```

Full gate evidence from `/tmp/osaurus-coord/evidence/T-J-WORKBOOK-TOOLS/20260520T014352Z`:

```text
swift-format lint --strict --recursive Packages App
# empty output, exit 0

swiftlint lint --reporter github-actions-logging
Done linting! Found 1119 violations, 0 serious in 584 files.

find scripts -name '*.sh' -print0 | xargs -0 shellcheck --severity=warning
# empty output, exit 0

swift test --package-path Packages/OsaurusCLI --parallel
Build complete! (26.43s)
[...]

PR #1197, opened 2026-05-21
added folder plugin hints

tpae:
## Summary

when selecting a folder context, detect file types and load tools that are needed

## Checklist


mimeding:
Reviewed with the #1177 tool-surface concern in mind. This direction looks right to me: context-aware, bias-only plugin injection for folders with xlsx/csv/pptx files, no new always-on default tools, and it respects installed plugins plus the filtered catalog/agent allowlist. CI is green and I do not see blockers.

PR #1240, opened 2026-05-24
feat: add mirror server option for improved availability

miguelAngelo1999:
## Summary

Adds a configurable mirror option so users can switch to an alternative server when the primary one is slow or unavailable.

## Changes

- \ModelMirror.swift\ — enum with mirror options, URL construction, UserDefaults persistence
- \HuggingFaceService.swift\ — branches on selected mirror; alternate API response decoder
- \ModelDownloadService.swift\ — \
esolveURL\ delegates to \ModelMirror.current.downloadURL(...)\
- \ModelMirrorSection.swift\ + \ConfigurationView.swift\ — Settings UI picker to switch between mirrors

## Testing

- Verified downloads complete successfully from both mirror options
- Settings picker persists selection across app restarts


tpae:
hi @miguelAngelo1999 , i dont see all the changes in the PR

mimeding:
Helper triage from the current plan pass:

- The first local compile issue in `ModelMirrorTests` is fixable by renaming the test helper field from `Type` to `type` and mapping it with `CodingKeys` (`case type = "Type"`).
- After that fix, the branch still does not compile because the tests reference `ModelMirror`, but the checked-out PR head does not contain a `ModelMirror` type under `Packages/OsaurusCore`.
- So this PR needs the mirror implementation restored/adapted to the current tree, or the stale `ModelMirrorTests` removed/rewritten, before it can be revived.

One extra maintainer-review note: the PR description mentions a TLS bypass delegate for self-signed corporate proxies, which crosses a security/trust boundary and should be explicit opt-in if this moves forward.

PR #1243, opened 2026-05-25
fix: preserve exposeToNetwork setting across launches

miguelAngelo1999:
The expose to local network setting resets on every launch because the Bonjour agent sync unconditionally sets exposeToNetwork=false when no agents have bonjourEnabled. Fix: only turn it ON when an agent requires it, never force OFF.


tpae (approved):
LGTM! Thank you for your contribution!

PR #1267, opened 2026-05-27
Fix zero free space on external model volumes

atomtanstudio:
## Summary

- fall back to `attributesOfFileSystem(.systemFreeSize)` when `volumeAvailableCapacityForImportantUsage` reports `0`
- keep the modern capacity query as the preferred source when it returns a positive value
- add a regression test for external volumes that expose writable free space through the legacy filesystem attributes

## Why

On some external model volumes, `volumeAvailableCapacityForImportantUsage` can return `0` even though the same path has real writable free space according to filesystem attributes. That causes model download preflight checks to fail immediately with zero available space, while already-installed models can still run from the same volume.

The storage helper already centralizes the free-space lookup used by model download preflight and the system monitor. This change keeps that central behavior, but treats a modern `0` value as unusable when the legacy query has a positive free-space value.

## Test

```bash
DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --package-path Packages/OsaurusCore --filter ModelDownloadServiceStorageTests/zeroImportantCapacityFallsBackToLegacyFreeSpace
```

### DIFF
--- a/Packages/OsaurusCore/Services/SystemMonitorService.swift
+++ b/Packages/OsaurusCore/Services/SystemMonitorService.swift
@@ -189,8 +189,8 @@ class SystemMonitorService: ObservableObject {
         // the dashboard reported `Available: 0 GB` even when Finder showed
         // tens of GB free. `OsaurusPaths.volumeFreeBytes` uses the modern
         // URL-keyed `.volumeAvailableCapacityForImportantUsageKey` first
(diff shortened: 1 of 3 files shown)


RaajeevChandran (approved):

RaajeevChandran:
thanks for your contribution! looks good to merge

PR #1320, opened 2026-06-02
opt in telemetry

RaajeevChandran:
## Changes

- [x] Behavior change
- [ ] Refactor / chore
- [ ] Tests
- [ ] Docs

## Checklist


bazzmann:
Don't forget to change the [Terms](https://osaurus.ai/terms): "Osaurus does not collect any personal data, usage data, **or telemetry**." — And I think it be a good thing to explain the why, about the _no telemetry -> telemetry opt-in_ change, although it's a natural step to keep things working.

RaajeevChandran:
@bazzmann we will be updating our terms to reflect this change. As for why we decided to add opt-in telemetry, we want to improve the overall stability of the app and also know where users are getting stuck to provide a better experience.

Osaurus is being used by both technical and non-technical users. We have been getting tonnes of feedback from the non-technical folks about certain parts of the app and its just incredibly hard for us to precisely understand their pain points and reduce that friction without telemetry. So we wanted to make informed decisions that are backed by actionable metrics instead of blindly guessing and hoping that the fixes we implement will resolve the underlying problems

bazzmann:
@RaajeevChandran sure, I understand this and I'm with you. It was just a friendly reminder as I see Terms are updated to April 2026 and many people are sensitive to telemetry presence or not, too. Especially in EU. And, as you told, many of them are non-technical users so unable to understand or correctly read a commit.

Thanks for your hard work!

RaajeevChandran:
Thanks for using osaurus!

PR #1361, opened 2026-06-05
Fix plugin tool-loop default iterations

jjang-ai:
## Summary
- Raise plugin-host completion default max iterations from 1 to 8 so omitted `max_iterations` can complete model -> tool -> tool-result loops.
- Add a source guard so the default cannot silently regress back to one-shot tool handling.

## Validation
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path Packages/OsaurusCore --filter RuntimePolicySourceTests/pluginHostInferenceDefaultsToMultiIterationTools --jobs 1 --no-parallel`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path Packages/OsaurusCore --filter PluginHostAPITests --jobs 1 --no-parallel`
- `git diff --check`
- No-sign Release rebuild via `scripts/live-proof/build-keychain-free-osaurus.sh build/DerivedData-nemotron-ultra-archheadroom-20260604-222621`
- Live Qwen 35B MXFP4 CRACK MTP tool-call smoke on rebuilt app: HTTP 200, `get_time` tool call emitted with `{"city":"Los Angeles"}`, `finish_reason: tool_calls`, no prose/tool-tag leak, elapsed 3.98s.

## Notes
- This fixes the plugin-host default that can make Qwen-style agent/tool callers stop before tool/result follow-up when callers omit `max_iterations`.
- The direct OpenAI chat-completions smoke proves parser/tool-call emission; the host iteration behavior is covered by source tests because `/v1/chat/completions` intentionally returns tool calls rather than executing plugins inline.


jjang-ai:
Sentry crash pass for this PR:

- Queried unresolved `apple-macos` issues for the last 14d. Top fatal groups are MLX/Metal command-buffer lifecycle or MLX null-access crashes, not plain OpenAI/tool parser errors.
- Representative groups:
  - `APPLE-MACOS-10` / issue `7526513243`: `-[_MTLCommandBuffer addCompletedHandler:]` assertion, latest frame `Evaluate.swift:1728 TokenIterator.next`, breadcrumb shows `gemma-4-e4b-it-4bit`, `batch=1`.
  - `APPLE-MACOS-14` / issue `7526889658`: `mlx::core::default_stream` bad access, frame `Evaluate.swift:3108 generateLoopTask`.
  - `APPLE-MACOS-17` / issue `7526976482`: command encoder already encoding assertion.
  - `APPLE-MACOS-3T` / issue `7530281408`: explicit Metal OOM, breadcrumb shows `gemma-4-12b-it-mxfp8`, `batch=1`.
- Conclusion: current evidence does not support “user enabled multi-batch” as the main cause; sampled crashes are `batch=1`. Context/memory pressure is still plausible for the explicit OOM group, but the previous breadcrumb used `promptTokens=...` and Sentry filtered it as prompt content, hiding the exact token count.
- Added commit `934e8d7a` to this PR: breadcrumb now records `input_tokens=...` instead of `promptTokens=...`, plus a source guard test. That makes the next Sentry crash answer whether context growth is the trigger without logging prompt contents or changing generation behavior.

Validation added for the breadcrumb commit:
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path Packages/OsaurusCore --filter RuntimePolicySourceTests/sentryInferenceBreadcrumbsExposeTokenCountWithoutPromptFilter --jobs 1 --no-parallel`
- `git diff --check`

jjang-ai:
Branch update / crash-warning note:

- Rebased PR #1361 onto latest `origin/main` (`f54d9c7f`). Current PR head is `d0a4e949`.
- I did not add a hard large-context limiter. That would change user behavior and could block legitimate long-context runs.
- I also did not inject warning text into assistant output, because the current runtime stream has content/reasoning/tool/stats channels but no non-content advisory channel. Mixing warnings into model text would pollute chats and API responses.
- Current safe change is telemetry-only: crash breadcrumbs now expose `input_tokens`, `batch`, and model id so the next Sentry event can prove whether the crash was context growth, batch/concurrency, or a vMLX/Metal lifecycle bug.
- If we want user-visible warnings without blocking, the right follow-up is a dedicated non-content advisory/UI channel, not prompt/output mutation.

jjang-ai:
Sentry follow-up pushed in `7f5b4956`.

- Fixed `APPLE-MACOS-2N` (`NSGenericException: Collection ... mutated while being enumerated`) by removing the `NSPasteboard.readObjects(forClasses:)` conversion path from clipboard file detection.
- Clipboard file detection now reads explicit file URL pasteboard types only, avoiding AppKit type-conversion enumeration.
- Added a source guard in `ClipboardContentDiagnosticsTests` so this path does not regress.
- Privacy: no clipboard payloads, paths, prompts, tool args, or response text are logged; existing diagnostics stay count/type-only.

Validation:
- `git diff --check`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path Packages/OsaurusCore --filter ClipboardContentDiagnosticsTests --jobs 1 --no-parallel`

jjang-ai:
Update: added the resource-exhaustion chat error path from the Sentry pass.

What changed:
- Resource/allocation failures now show an actionable chat error: ran out of system resources, free memory/unload other models/use a smaller or more-quantized model, then retry.
- Non-resource runtime errors are preserved unchanged, so this does not mask template/model/runtime failures like unsupported model type.
- This is presentation-only. It does not change sampling, prompts, tool parsing, model behavior, cache behavior, or runtime policy.

Validation run locally:
- `git diff --check`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path Packages/OsaurusCore --filter ChatErrorMessagesTests --jobs 1 --no-parallel`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path Packages/OsaurusCore --filter ClipboardContentDiagnosticsTests --jobs 1 --no-parallel`

Sentry/privacy notes:
- No Sentry test event was sent.
- No prompts, outputs, tool args, filenames, paths, URLs, or clipboard payloads are logged. The telemetry changes in this PR remain counts/stats only.
- The top MLX/Metal crash groups are still runtime-level follow-up work; this PR adds safer reporting/user messaging and fixes the actionable pasteboard enumeration crash.

PR #1371, opened 2026-06-05
Gate local model loads on available memory

jjang-ai:
## Summary

- Add live available-memory accounting to the local model load gate so impossible loads refuse before MLX/Metal schedules work.
- Count shallow `.safetensors` shard layouts when exact known shard names are absent, preventing `weightsBytes == 0` from bypassing preflight.
- Surface `available_memory_bytes` and `required_available_bytes` in `/health` `ram_feasibility`.
- Keep Qwen/plugin tool loops on the configured chat max-tool-attempt budget instead of tiny hard-coded defaults.

## Root Cause

Sentry issue `7530281408` / `APPLE-MACOS-3T` is a Metal OOM during local model load, not a Gemma template/parser issue. The latest event showed a 24 GB-class Mac (`Mac14,2`) crashing after breadcrumb `model.load begin model=qwen3.6-35b-a3b-mxfp4`, with `kIOGPUCommandBufferCallbackErrorOutOfMemory`.

The old gate considered projected resident footprint but did not compare the incoming load plus KV/headroom against currently available memory. On a memory-constrained machine, Metal could still be asked to allocate and crash the app.

## Validation

- `git diff --check`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path Packages/OsaurusCore --filter 'RuntimePolicySourceTests/(modelRuntimeWeightSizePreflightIsManualMultiModelOnly|pluginHostInferenceDefaultsToMultiIterationTools)' --jobs 1 --no-parallel`
- No-sign Release app build passed:
  - `build/DerivedData-ram-gate-tool-loop-20260605-155733/Build/Products/Release/osaurus.app`
  - ad-hoc signed, no signing identity
  - app build used pinned `osaurus-ai/vmlx-swift@525fc6dc17650efd420e7f9adfc1c37e40650a73`
- Live API proof on the rebuilt app at `127.0.0.1:4242`:
  - `qwen3.6-35b-a3b-mxfp4-crack-mtp` returned `qwen live ok`, `finish_reason=stop`
  - OpenAI tool-call round trip returned `lookup_status({"node":"node-2"})`, then consumed the tool result and answered normally
  - no raw tool tags, leaked parser envelopes, or random early stop observed
- Live cache/resource telemetry after the Qwen tool-result turn:
  - `incoming_weights_bytes=19788916296`
  - `kv_headroom_bytes=3957783259`
  - `required_available_bytes=23746699555`
  - topology: `layers=40`, `kvLayers=10`, `mambaLayers=30`, `companion=ssm`, `restore=disk-backed`
  - `disk_l2_hits=1`, `disk_l2_misses=5`, `disk_l2_stores=5`
  - `ssm_companion_hits=1`, `ssm_companion_misses=0`, `ssm_companion_rederives=0`

## Residual Risk

- Local proof machine has 128 GB RAM, so Qwen MXFP4 correctly loads here. The Sentry production crash is on a 24 GB-class Mac. This PR builds and exposes the refusal path, but exact refusal proof still needs a constrained-memory machine/profile or a test hook that does not alter production behavior.
- Keychain-free dev launch cannot prove built-in memory DB tools because existing `~/.osaurus` SQLCipher DBs do not open without keychain. OpenAI tool-call parser/runtime proof is unaffected.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -410,7 +410,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "044bf8ad412e701be9f7cfc9e6b84d8894c22de7"
+        "revision" : "7a7481f9793d253cb2dacff31cb0b3f18a307ef6"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -410,7 +410,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "d03beb22b0f9421b30b30c4e0dcadcb61d4d89d7"
+        "revision" : "45a1560d96a10899a7c20f4797b2769d26e3294d"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -410,7 +410,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "45a1560d96a10899a7c20f4797b2769d26e3294d"
+        "revision" : "044bf8ad412e701be9f7cfc9e6b84d8894c22de7"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -410,7 +410,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "525fc6dc17650efd420e7f9adfc1c37e40650a73"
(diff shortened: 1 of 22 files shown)


jjang-ai:
Runtime pin update pushed.

- vMLX `main` is now `7a7481f9793d253cb2dacff31cb0b3f18a307ef6` (`Keep Nemotron streaming opt-in until proven`).
- This PR now pins Osaurus to that revision in `Packages/OsaurusCore/Package.swift` and all three resolved files.
- Local source proof: `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path Packages/OsaurusCore --filter RuntimePolicySourceTests --jobs 1 --no-parallel` passed: 78 tests, 0 failures.
- vMLX proof before pin: `NemotronHJANGTQDispatchFocusedTests` passed; release JPREG Nemotron Ultra JANGTQ regular path completed 3/3 coherent turns without the streaming fatal/OOM crash, TQ disk round-trip PASS, low footprint after quiesce.

Important boundary: the attempted auto offset-addressed Nemotron streaming path was a live speed regression (~0.3 tok/s), so vMLX now keeps that streaming backend explicit/diagnostic-only instead of auto-enabling it. Regular JANGTQ remains the default until a faster streaming path is actually proven. Nemotron Ultra speed is therefore still partial, not claimed as the 8-10 tok/s target in this PR.

jjang-ai:
Follow-up CI fix pushed after the rebased full `test-core` job failed twice.

Root cause from downloaded `test-core-xcresult-2`: `ModelMediaCapabilities.from(directory:modelId:)` returned text-only for a missing `config.json` Nemotron Omni model because the broad Nemotron thinking-family guard ran before the documented fallback to `from(modelId:)`, and that guard includes Omni.

Fix: preserve `config_omni.json` as the strongest signal, fall back to `modelId` when `config.json` is missing/unreadable, then apply the text-only Nemotron Ultra guard only for non-Omni thinking models with readable config.

Local verification:
- `swift test --package-path Packages/OsaurusCore --filter ModelMediaCapabilities --jobs 1 --no-parallel` passed: 34 tests.
- `swift test --package-path Packages/OsaurusCore --filter CapabilityFromDirectoryTests --jobs 1 --no-parallel` passed, including `unreadableConfigFallsBackToModelId`.
- `swift test --package-path Packages/OsaurusCore --filter RuntimePolicySourceTests --jobs 1 --no-parallel` passed: 78 tests.
- `git diff --check` clean.

Issue #903, opened 2026-04-21
System prompts not injected at runtime


DBlankenship1967:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest `main` or release

### Describe the bug

System prompts are saved correctly to disk (confirmed via ~/.osaurus/agents/*.json) but are not being applied at runtime. Model has no knowledge of any instructions in the system prompt.
Reproduced with two agents — one with a full prompt, one with a single line ("Your name is Gerald"). In both cases the model behaves as if no system prompt exists.

### Provider type

Local model (MLX)

### Osaurus version / commit

0.16.16

### macOS version

Tahoe 26.4.1

### Apple Silicon chip

M4 Pro

### Logs

```shell

```

RaajeevChandran:
Can you check if both these toggles are disabled for your agent?

<img width="896" height="420" alt="Image" src="https://github.com/user-attachments/assets/1e747de1-7300-43b3-b77c-0390538f45dc" />

DBlankenship1967:
Hi Raajeev, 

Yes, I just doublechecked and both of those toggles are set just as you pictured in your screenshot. I've been confused about memory and how it works between the default agent and specialist agents generally.

EDIT: I just tested again on the latest build. Here's screenshot: 

<img width="871" height="1372" alt="Image" src="https://github.com/user-attachments/assets/5c1e962f-8a97-4b47-b024-874c9e120378" />

RaajeevChandran:
Could you show a screenshot of the system prompt in this agent?

DBlankenship1967:
Hi Raajeev, you got me thinking and it dawned on me that almost all of my currently installed models are Gemma 4 variants. I tested all of them and they all fail in the same way. I tested with a Cydonia build and it got it immediately. So I guess it's either the Gemma 4 family or something with my prompt. Anyway, here's my prompt for this agent.

<img width="1260" height="1582" alt="Image" src="https://github.com/user-attachments/assets/db68f6f7-6c8f-4597-98dd-0baed9a20b89" />

DBlankenship1967:
Quick followup: I did some more testing. While none of my Gemma 4 models could apparently read the system prompt, copying and pasting it at the start of a new chat did work.

mimeding:
Contributor drive-by — I traced the system-prompt path from disk to the MLX request, and there are a couple of specific things that would pin down the actual failure point. Could you share any of the following when you have a moment?

[...]

PR #1527, opened 2026-06-16
Add system prompt injection proof trace

mimeding:
## Summary
- add a source-level system prompt injection trace on PromptManifest so #903 evidence can show a configured agent prompt reached the static persona/prefix surface
- add RuntimeProofSystemPromptEvidence and require it alongside the existing live prompt probe before a system-prompt row can be marked proven
- document that source trace alone is contract coverage, not runtime proof

## Validation
- git diff --check
- SCRIPT_INPUT_FILE_COUNT=4 ... swiftlint lint --strict --use-script-input-files
- OSAURUS_TEST_ROOT=/tmp/osaurus-test swift test --package-path Packages/OsaurusCore --filter SystemPromptInjectionTraceTests
- OSAURUS_TEST_ROOT=/tmp/osaurus-test swift test --package-path Packages/OsaurusCore --filter RuntimeProofValidationTests

Refs #903

### DIFF
--- a/Packages/OsaurusCore/Services/Chat/PromptManifest.swift
+++ b/Packages/OsaurusCore/Services/Chat/PromptManifest.swift
@@ -10,6 +10,15 @@
 import CryptoKit
 import Foundation
 
+// MARK: - PromptSectionID
+
+/// Stable IDs for sections that other proof and budget paths inspect by name.
+public enum PromptSectionID {
+    public static let platform = "platform"
+    public static let persona = "persona"
+    public static let memory = "memory"
+}
+
 // MARK: - PromptSection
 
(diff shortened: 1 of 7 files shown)


mimeding:

- Replaced the hardcoded prompt section id with shared section constants.
- Routed injection tracing through the same prompt rendering path used by the composer.
- Added regression coverage for the shared renderer path.

Validation:
- `git diff --check`
- `swift test --package-path Packages/OsaurusCore --filter SystemPromptInjectionTraceTests`
- `swift test --package-path Packages/OsaurusCore --filter RuntimeProofValidationTests`

tpae (approved):

PR #1557, opened 2026-06-17
Add Business Document Studio service

mimeding:
## Summary

Adds `BusinessDocumentStudioService`, a format-neutral core service for inspecting, previewing, and explicitly exporting business documents through the existing document registry and workflow services.

This consolidates CSV/TSV, XLSX, PDF/PPTX, rich-text, and text-fallback routing behind one Codable inspection surface so UI, plugin, and attachment flows do not each need to reimplement format-specific preview/export decisions.

## What changed

- Added a bounded Business Document Studio inspection model with summary, registry roles, security metadata, previews, and export options.
- Routed CSV/TSV previews and exports through `CSVTableWorkflowService`.
- Routed workbook inspection and XLSX export through `WorkbookWorkflowService`, with bounded sheet/row/cell samples.
- Wrapped PDF/PPTX workflow previews and missing-emitter diagnostics without pretending text writes are binary package creation.
- Added guarded text fallback export with destination containment, overwrite protection, byte limits, and structured package target rejection.
- Documented the service in the features and high-fidelity output safety audit docs.

## Safety notes

- No default agent tool is added.
- No arbitrary-path HTTP endpoint is added.
- Structured package extensions are rejected for text fallback writes, and package export target extensions must match the requested structured format.
- Preview sizes are capped by policy and format-specific parsing still uses the existing document limits.

## Validation

- `swiftlint lint Packages/OsaurusCore/Services/Documents/BusinessDocumentStudioService.swift Packages/OsaurusCore/Tests/Documents/BusinessDocumentStudioServiceTests.swift`
- `git diff --check`
- `bash scripts/i18n/check.sh`
- `swift test --package-path Packages/OsaurusCore --filter BusinessDocumentStudioServiceTests`
- `swift test --package-path Packages/OsaurusCore --filter 'BusinessDocumentStudioServiceTests|CSVTableWorkflowServiceTests|WorkbookWorkflowServiceTests|PDFPPTXWorkflowServiceTests|DocumentFormatRegistryTests'`
- `OSAURUS_TEST_ROOT=/tmp/osaurus-business-document-studio-test make test` — 3763 tests passed
- `make cli`
- `swift build --package-path Packages/OsaurusCore -c release`

Note: the keychain-disabled full-suite mode still fails an existing keychain lifecycle expectation where the bystander secret is nil under disabled keychain mode. The normal full core suite passed.

## Rollback

Revert commit `d6138991`.

### DIFF
--- a/Packages/OsaurusCore/Services/Documents/BusinessDocumentStudioService.swift
+++ b/Packages/OsaurusCore/Services/Documents/BusinessDocumentStudioService.swift
@@ -667,15 +667,15 @@ public struct BusinessDocumentStudioService: Sendable {
         guard url.isFileURL else {
             throw BusinessDocumentStudioError.destinationIsNotFileURL(url)
         }
-        if !policy.allowOverwrite, FileManager.default.fileExists(atPath: url.path) {
-            throw BusinessDocumentStudioError.destinationAlreadyExists(url)
-        }
         if let allowedDirectory = policy.allowedDirectory {
             let root = allowedDirectory.standardizedFileURL.resolvingSymlinksInPath().path
             let parent = url.deletingLastPathComponent().standardizedFileURL.resolvingSymlinksInPath().path
             guard parent == root || parent.hasPrefix(root + "/") else {
                 throw BusinessDocumentStudioError.destinationOutsideAllowedDirectory(url)
             }
+        }
+        if !policy.allowOverwrite, FileManager.default.fileExists(atPath: url.path) {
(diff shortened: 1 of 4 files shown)


mimeding:

- Reordered destination validation so containment is checked before overwrite/existence checks.
- Added regression tests proving existing and missing outside paths return the same containment error without leaking filesystem existence.

Validation:
- `swift test --package-path Packages/OsaurusCore --filter BusinessDocumentStudioServiceTests`
- `git diff --check`

tpae (approved):

PR #1565, opened 2026-06-18
Add shared global proxy controls

mimeding:
## Summary
- Move global proxy parsing and URLSession proxy shaping into a shared networking package while preserving Core compatibility aliases.
- Add Settings proxy status text for disabled, configured/applied, and invalid proxy states with localized strings.
- Apply the persisted global proxy policy to plugin repository refreshes and plugin artifact downloads, with focused repository tests and updated docs.

## Validation
- swift test --package-path Packages/OsaurusRepository
- OSAURUS_DISABLE_KEYCHAIN_FOR_TESTS=1 OSAURUS_TEST_ROOT=/tmp/osaurus-test-a1 swift test --package-path Packages/OsaurusCore --filter 'GlobalProxyConfigurationTests|ProviderNetworkDiagnosticsTests|PrivacyLocalizationTests'
- swift test --package-path Packages/OsaurusCore --parallel --num-workers 1
- git diff --check
- scripts/i18n/check.sh


mimeding:

- Removed session invalidation during proxy refresh so a caller holding the previous shared session is not handed an invalidated object during a concurrent refresh.
- Added a loopback regression test that keeps an old retained session usable across refresh.

Validation:
- `swift test --package-path Packages/OsaurusRepository --filter RepositoryGlobalProxySettingsTests`
- `swift test --package-path Packages/OsaurusRepository`
- `git diff --check`

tpae (approved):

PR #1568, opened 2026-06-18
Surface local model cache imports in Models

mimeding:
## Summary
- Adds Hugging Face cache path configuration and import visibility in the Models view.
- Lets users point Osaurus at pre-downloaded local model folders without redownloading them.
- Keeps incomplete or imported-model status available behind a collapsed panel so the default Models view stays quiet.

## Validation
- `swift test --package-path Packages/OsaurusCore --filter 'ExternalModelLocatorTests|HuggingFaceDownloadPathTests|ModelDownloadServiceStorageTests'` — 32 tests passed after rebase onto current `origin/main`.
- `bash scripts/i18n/check.sh` — passed.
- `git diff --check` — passed.
- `make app XCODEBUILD_FLAGS="CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO"` — attempted locally; the CLI target completed and the app target stalled idle in Xcode/SwiftBuild after dependency compilation, so the run was interrupted and GitHub CI is the current build signal.
- GitHub CI — green; merge state CLEAN.

## Notes
- The external model details remain available for debugging/import flows, but are collapsed by default to avoid noisy model-library rows for users with many incomplete local downloads.
- This branch was rebased onto current `origin/main` to replace the previous unrelated `test-core` failure with a fresh CI run.

### DIFF
--- a/Packages/OsaurusCore/Tests/Service/ExternalModelLocatorTests.swift
+++ b/Packages/OsaurusCore/Tests/Service/ExternalModelLocatorTests.swift
@@ -385,8 +385,20 @@ struct ExternalModelLocatorTests {
 
         ExternalModelLocator.testRootsOverride = [(root: hfRoot, source: .huggingFaceCache)]
         ExternalModelLocator.rescan()
-        #expect(FileManager.default.fileExists(atPath: OsaurusPaths.externalModelsManifestFile().path))
+        let manifestFile = OsaurusPaths.externalModelsManifestFile()
+        #expect(FileManager.default.fileExists(atPath: manifestFile.path))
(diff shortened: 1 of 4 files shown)


mimeding:

- Tightened the external model manifest reload test so it clears the test root override before invalidation/reload.
- This now proves the manifest path is actually used rather than rediscovering the temporary source directory.

Validation:
- `swift test --package-path Packages/OsaurusCore --filter ExternalModelLocatorTests`
- `git diff --check`

mimeding:
Moving this back to draft. The incomplete-model surface can get noisy for users, so this needs a UX pass before it should be considered ready.

Issue #22, opened 2025-08-26
Benchmarks for current models, more sizes


rosmur:
### Problem statement

The docs show results for only llama3.2 3B which is quite old and very small. Can results of Osaurus vs ollama vs LMStudio vs llama.cpp be shared for current models, bigger sizes and different architectures. Listed below are some current, most used models

1. Qwen3-Coder 
2. Qwen3-A3B-30B-Instruct (MOE)
3. GLM-4.5-Air
4. Mistral Small 3.2 24B(Dense)
5. GPT-OSS-20b (MOE with reasoning)

rosmur:
@epsimatic you just contradicted yourself..

epsimatic:
Did I?

For benchmarks, we only care about performance (throughput and TTFT) and we don't care about model's actual output. llama3.2 3B, a typical dense 3B model, would do fine there.

For real tasks, we care about model output — the actual text, and llama3.2 is clearly lacking behind Gemma3 or Qwen3.

skills-up:
Given that `README.md` is comparing 4 bit quantized version's performance on Osarus with full model's performance on other tools, I think it will under perform for some time (until some massive performance improvements are gained).

PR #1571, opened 2026-06-18
Add model benchmark evidence reports

mimeding:
## Summary
- Add deterministic JSON and Markdown benchmark evidence reports for OpenAI-compatible model comparisons.
- Record model id, prompt case, timing, token/s, optional RSS memory evidence, and pass/partial/failed/unproven status.
- Mark successful generation rows without token/s as failed, and document the report workflow for issue #22.

## Validation
- python3 -m py_compile scripts/benchmark/benchmark_models.py scripts/benchmark/test_benchmark_models.py
- python3 scripts/benchmark/test_benchmark_models.py
- python3 scripts/benchmark/benchmark_models.py --help
- bash -n scripts/benchmark/run_bench.sh
- make -n bench-models
- git diff --check

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ WORKSPACE := osaurus.xcworkspace
 DERIVED := build/DerivedData
 XCODEBUILD_FLAGS ?=
 
-.PHONY: help cli app install-cli serve status test ci-test clean bench-setup bench-ingest bench-ingest-chunks bench-run bench bench-models evals-prep evals evals-verbose evals-report evals-all evals-all-verbose evals-all-report evals-capture-screen evals-loop evals-matrix evals-diff evals-contribute evals-compat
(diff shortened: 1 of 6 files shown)


tpae:
@mimeding i would avoid adding this PR in favor of this: https://github.com/osaurus-ai/osaurus/tree/main/Packages/OsaurusEvals#recording-a-run-committed-snapshot--history

mimeding:
Update after second-opinion review:

- Benchmark rows that succeed without positive token/s are now reported as `unproven` instead of being blended into ordinary pass/fail totals.
- Aggregate status now only passes when every row has complete passing evidence.
- Added a focused `make bench-models-test` target for the benchmark report scripts.

Local validation:
- `make bench-models-test`
- `git diff --check`

mimeding:
Agreed. I will avoid pushing this as a separate benchmark/report source and align any future model evidence work with the existing committed eval snapshot/history flow instead. This PR can be treated as superseded.

mimeding:
Closing this as superseded per maintainer direction. Future benchmark/model evidence work should use the existing committed eval snapshot/history path rather than adding a separate report source.

PR #1574, opened 2026-06-18
Add native audio file attachments for capable local models

### LINKED ISSUE #1183: Support native audio files with Gemma4
### Checks

### Area

Other

### Problem statement

The Gemma4 model supports audio natively according to [the Gemma docs](https://ai.google.dev/gemma/docs/capabilities/audio).  

I tried dragging in a wav file into Osaurus with Gemma4 selected, but it gave an error that audio was not supported.  Likewise if I choose the paperclip, it shows wav files as greyed out.  I think this is a limitation of Osaurus rather than Gemma4.

### Proposed solution

Each model would have a capabilities descriptor (something like a model card), and Osaurus would tailor the UI according to that model's capabilities.

It would be similar to how it differentiates between Vision enabled and non-vision models, but for audio.  

### Alternatives considered

Writing inference code snippet.

### Additional context

I realize this is a bit of an edge case / advanced feature, but I think it would be a really cool addition.

### Contribution

- [x] I would be willing to submit a PR for this feature

mimeding:
## Summary
- Enable audio file attachments only when local bundle facts prove native audio input support.
- Keep name-only, unsupported, and unproven audio models gated with visible typed errors instead of advertising capability.
- Keep Gemma4 audio blocked as `partial`/diagnostic evidence until live Osaurus/vMLX runtime proof exists, even when installed bundle metadata contains `embed_audio`, `audio_tower`, or `audio_config` markers.
- Centralize audio file classification, picker/drop UTTypes, format normalization, and attachment size limits.
- Thread local model directories into runtime policy so audio requests are accepted only with capability evidence.

Fixes #1183.

## Validation
- `OSAURUS_DISABLE_KEYCHAIN_FOR_TESTS=1 OSAURUS_TEST_ROOT=/tmp/osaurus-test-a10 swift test --package-path Packages/OsaurusCore --filter 'ModelMediaCapabilitiesMCDCTests|MLXServiceRuntimePolicyTests|DocumentParserShimTests|LocalModelDetectionTests|ChatAttachmentSecurityTests|MultimodalContentPartTests'`
- `git diff --check`
- `scripts/i18n/check.sh`

## Maintainer Feedback Addressed
- Gemma4 multimodal audio is no longer promoted to usable support from metadata alone. Installed Gemma4 audio markers now surface as partial/blocked diagnostics until there is live runtime proof.

### DIFF
--- a/Packages/OsaurusCore/Managers/Model/ModelManager.swift
+++ b/Packages/OsaurusCore/Managers/Model/ModelManager.swift
@@ -230,6 +230,7 @@ final class ModelManager: NSObject, ObservableObject {
         NotificationCenter.default.publisher(for: .localModelsChanged)
             .receive(on: DispatchQueue.main)
             .sink { [weak self] _ in
+                ModelMediaCapabilities.invalidateComposerDescriptorCache()
                 self?.refreshDownloadStates()
             }
             .store(in: &cancellables)
@@ -1638,6 +1639,7 @@ extension ModelManager {
         localModelsCacheCondition.unlock()
         LocalReasoningCapability.invalidate()
         LocalGenerationDefaults.invalidate()
+        ModelMediaCapabilities.invalidateComposerDescriptorCache()
     }
 
     nonisolated static func localModelsScanDiagnosticJSONObject() -> [String: Any]? {
--- a/Packages/OsaurusCore/Models/Chat/MEDIA-ATTACHMENT-USER-GUIDE.md
+++ b/Packages/OsaurusCore/Models/Chat/MEDIA-ATTACHMENT-USER-GUIDE.md
@@ -17,9 +17,8 @@ this:
 | Loaded model | Drop zone accepts | File picker shows |
 |---|---|---|
 | Nemotron-3-Nano-Omni (any quant tier) | image + audio + video | image + docs + audio + video |
-| Installed Gemma 4 bundles with audio tensors in the weight map | image + audio | image + docs + audio |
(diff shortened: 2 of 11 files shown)


mimeding:

- Added a shared cached model media capability descriptor lookup with invalidation on local model/cache changes.
- Updated chat attachment UI paths to use the cached lookup instead of repeating descriptor disk scans from computed SwiftUI state.
- Added reuse/invalidation/source-change coverage.

Validation:
- `swift test --package-path Packages/OsaurusCore --filter ModelMediaCapabilitiesMCDCTests`
- `git diff --check`

mimeding:
Update after second-opinion review and rebase:

- Rebased onto current main.
- Stabilized the media capability cache test suite by serializing the shared-cache coverage.
- Re-ran the focused media capability coverage locally.

Local validation:
- `OSAURUS_DISABLE_KEYCHAIN_FOR_TESTS=1 OSAURUS_TEST_ROOT=/tmp/osaurus-pr1574-recheck-2 swift test --package-path Packages/OsaurusCore --filter ModelMediaCapabilitiesMCDCTests`
- `git diff --check`

mimeding:
Moving this back to draft. The Gemma4 multimodal-audio issue needs current-main proof and a narrower fix before this should be review-ready.

mimeding:
Closing this draft intentionally for now. Native audio attachment support should come back after the Gemma4 multimodal audio behavior is reproduced and the media path has live evidence, so we do not claim model-facing audio support before the compatibility bug is understood.

PR #1629, opened 2026-06-22
Add unified evidence report registry

mimeding:
## Summary
- Add a canonical evidence report registry for eval, benchmark, runtime, live-proof, run-trace, provider, and custom artifact summaries.
- Register existing local report descriptors without moving artifacts, with filtering, dedupe, unavailable/error rows, and stable JSON encoding.
- Redact sensitive metadata before summaries are stored or serialized.

## Validation
- `swift test --package-path Packages/OsaurusCore --filter EvidenceReportRegistryTests` passed: 6 tests
- `git diff --check`
- Scoped SwiftLint on touched Swift files: 0 violations

## Notes
- This is the shared model/service foundation only; persistence and UI wiring are left for follow-up PRs that migrate existing report surfaces onto the registry.

### DIFF
--- a/Packages/OsaurusCore/Models/Evidence/EvidenceReportRegistry.swift
+++ b/Packages/OsaurusCore/Models/Evidence/EvidenceReportRegistry.swift
@@ -1,0 +1,302 @@
+//
+//  EvidenceReportRegistry.swift
+//  osaurus
+//
+//  Typed summaries for locally-produced eval, benchmark, runtime, live-proof,
+//  run-trace, and provider evidence artifacts. The registry is intentionally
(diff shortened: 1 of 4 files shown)


mimeding:
Ready for review.

This is the shared evidence registry prerequisite for upcoming model, eval, and Computer Use report consumers. Current state: non-draft, merge-clean, and all required checks are green (`test-core`, `test-cli`, `swiftlint`, `shellcheck`, release drafter). The updated development plan now treats this as the single report/evidence registry gate so follow-up work does not create another reporting system.

tpae (approved):

Issue #85, opened 2025-10-04
Disable confirm on quit or implement a more graceful approach


aptonline:
### Problem statement

As a user, when I shutdown my device I have to manually confirm I want to quit Osaurus leading to unnecessary actions.

### Proposed solution

Add shutdown detection to automatically quit the app when shutting down vs accidentally choosing quit (where you would want confirmation). 

Alternatively add a toggle so that this check can be disabled/enabled if needed.

### Alternatives considered

N/A

### Additional context

More notable when using laptop as tend to shutdown rather than sleep.

PR #1686, opened 2026-06-24
Repin vmlx-swift: fix Mistral3/Pixtral VLM model-load crash

jjang-ai:
## Summary
Bumps vmlx-swift `6b77b1ee` → `3f122322` (osaurus-ai/vmlx-swift#84, merged) to fix a **fatal model-load crash** on Mistral3 / Ministral / Pixtral **VLM** bundles packaged in the standard Hugging Face layout. This is the Sentry crash `EXC_BREAKPOINT … rope_parameters['rope_theta'] is required`.

## Root causes fixed in vmlx#84
1. **Flat `rope_theta`** — VLM language attention hard-required a nested `rope_parameters['rope_theta']` and `fatalError`'d otherwise. Standard Mistral/Ministral configs put `rope_theta` flat at the top level. Fixed in both the fp and JANGTQ attention paths.
2. **Split processor config** — the VLM processor decoded image-processor fields only from a nested `image_processor` key; the split HF layout keeps image params flat in `preprocessor_config.json` and tokens in `processor_config.json`, so the loader threw `keyNotFound`. Now accepts both nested and flat layouts (Mistral3 + Pixtral). Gemma4 (nested `processor_config.json`) is unaffected.

The external models that trigger the crash ship **both** layouts, so both fixes are required to actually load them.

## Live proof (dev app)
On `mlx-community/Mistral-Small-3.1-24B-Instruct-2503-4bit` (unmodified):
- **Pre-fix** (May-23 installed app): the server **crashes on model load** (rope `fatalError`).
- **Post-fix** (this repin, built + installed): loads as `Mistral3VLM`, server stays up, `/v1/chat/completions` returns coherent text — *"The ocean is a vast body of saltwater that covers approximately 71% of the Earth's surface."* (HTTP 200, 38.8 tok/s). Image pipeline also verified (correct patch-token count, vision tower runs).

## Changes
- 4 vmlx pin sites bumped to `3f122322`.
- `ImageGenerationBridgeContractTests` / `RuntimePolicySourceTests` proven-revision assertions + changelog updated to the new SHA.

Supersedes the engine-only fix in vmlx-swift#83 (closed).

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "6b77b1ee682d81b3f84d84ac455de565fea3f348"
+        "revision" : "29d05174151894f2a4072c93079736860480ae4e"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "6b77b1ee682d81b3f84d84ac455de565fea3f348"
+        "revision" : "29d05174151894f2a4072c93079736860480ae4e"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.swift
+++ b/Packages/OsaurusCore/Package.swift
@@ -34,7 +34,7 @@ let package = Package(
         // live model, cache, parser, API, and UI evidence.
         .package(
             url: "https://github.com/osaurus-ai/vmlx-swift",
-            revision: "6b77b1ee682d81b3f84d84ac455de565fea3f348"
+            revision: "29d05174151894f2a4072c93079736860480ae4e"
         ),
         // FluidAudio 0.14.3 added a breaking `language:` parameter to TTS
         // calls that osaurus's `TTSService` doesn't pass. Pinning to the
--- a/Packages/OsaurusCore/Services/ModelRuntime/SwiftTransformersTokenizerLoader.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime/SwiftTransformersTokenizerLoader.swift
@@ -381,10 +381,26 @@ private struct TokenizerBridge: MLXLMCommon.GenerationPromptControllableTokenize
         // ChatML/Hermes format) and may mis-route to a ChatML fallback. Apply the
         // complete native Mistral template so reasoning / tools / vision all work
         // natively. Detected by Mistral's tekken special tokens, not by name.
+        //
+        // `convertTokenToId` is NOT a reliable presence test: BPE/Unigram
+        // tokenizers return the unknown-token id for any absent token (see
+        // BPETokenizer `tokensToIds[token] ?? unknownTokenId`), so `!= nil` is
+        // true for every string on a tokenizer that declares an <unk> (e.g. a
+        // Gemma SentencePiece pack). Require each tekken token to round-trip —
+        // `convertIdToToken(convertTokenToId(t)) == t` — which only holds when
+        // the token genuinely exists in the vocab (unk-mapped strings decode
+        // back to "<unk>", not to themselves).
+        func hasExactToken(_ token: String) -> Bool {
+            guard let id = upstream.convertTokenToId(token),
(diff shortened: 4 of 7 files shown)


jjang-ai:
## Live dev-app cross-function matrix (fixed build, :1337)

Built osaurus against the repinned vmlx, installed it, and ran the matrix on the previously-crashing `mlx-community/Mistral-Small-3.1-24B-Instruct-2503-4bit`:

| Function | Result | Evidence |
|---|---|---|
| VLM load (Sentry crash) | ✅ | loads as `Mistral3VLM`; pre-fix May-23 build crashed the server here |
| Multi-turn + context | ✅ | recalled name + number, `7×6=42` |
| Real image / VL | ✅ | correctly read "hello" from a real image (validates the flat-layout image-processor decode) |
| Stress burst (4×) | ✅ | all correct, ~0.35s each, stable |
| Model switch (qwen↔mistral) | ✅ | no crash on churn |
| Other-model regression | ✅ | `qwen3-4b` loads + replies on the fixed build |
| RAM / stability | ✅ | 10.5 GB; server alive through the whole run |
| Tool calling | ⚠️ separate | emits `<function=…>` non-JSON + leaks `<\|im_start\|>` — a chat-template mismatch for this external pack, **not** the load fix (the vmlx change touches only Mistral3/Pixtral model-load files, zero tokenizer/template/tool code). Tracked separately. |

The load crash (rope + split-processor) is fully resolved and the VLM is fully usable for chat + vision.

jjang-ai:
## Update: folded in vmlx #85 (chat-template fix) + tool-leak diagnosis

Re-pinned to vmlx `6d50965` (main), which now carries **three** fixes:
1. #84 — flat `rope_theta` + split processor-config crash (the original Sentry crash).
2. #85 — load the model's own `chat_template.json` for Mistral/Pixtral VLMs.

### Chat-template fix (#85) — what it does
The stress test surfaced `<|im_start|>` ChatML tokens leaking in **tool mode**. Root cause: this pack ships its chat template only in `chat_template.json` (Mistral `[INST]`/`[IMG]`), and vmlx's `resolveChatTemplateSidecarSubstitution` only loaded a JSON sidecar that `isVisionChatTemplate()` recognized — which only knew Qwen/Gemma markers, not Mistral's `[IMG]`. So the model got a ChatML default. The fix adds a gated `genericJsonTemplate` path; **verified live** that the model now loads its real `[INST]` template (shim materialized with `[INST]`/`[IMG]`, no `<|im_start|>`). Codex-reviewed for VLM-path non-disruption (Qwen/Gemma/Zaya/LFM2/jinja/inline all untouched).

### Remaining tool-mode leak — separate, model-pack limitation (NOT this fix)
With `[INST]` now correctly loaded, the `<|im_start|>` in tool output is the **model hallucinating** ChatML in its *output*, because this mlx-community pack's `chat_template.json` has **no tool block** (`[AVAILABLE_TOOLS]`). With tools passed, the model gets no tool grounding and improvises a Hermes/ChatML format that `MistralToolCallParser` (expecting `[TOOL_CALLS]`) doesn't parse → raw text leaks as content. osaurus's own Mistral bundles ship tool-capable templates and are unaffected. A proper fix (Mistral tool-aware fallback template when tools are present + the bundle template lacks `[AVAILABLE_TOOLS]`) is a separate, larger follow-up.

### Other known-separate item
Multi-image (2+ images) → HTTP 500 = the documented osaurus **n=1 image cap** (all VLMs), unrelated to these fixes.

**Net:** load crash fixed, model loads + chats + does single-image vision correctly, and now loads its correct base template. Tool calling for this external pack remains limited by its template (tracked separately).

jjang-ai:
## Mistral 3.x now works natively end-to-end (load + chat + tools + vision + reasoning)

Repinned to vmlx-swift `41caf0af` (#84 + #85 + #86) and added the osaurus tokenizer-bridge Mistral case.

**Root causes (all real, no guards):**
1. Load crash — flat `rope_theta` + split processor config (#84).
2. Template — pack ships only the HF vision `chat_template.json` (no tools); osaurus's `TokenizerBridge` had no Mistral case → ungrounded → ChatML leak (#85 + bridge case + #86 template).
3. Tools — 2503/2506 tekken has no `[ARGS]` token; it emits the JSON-array `[TOOL_CALLS][{"name":..,"arguments":..}]`, which the parser dropped (#86 parser).

**Live proof (Mistral-Small-3.1-24B, clean debug-free build):**
- Tools: `tool_calls=[get_weather{"city":"Tokyo"}]`; round-trip → *"The weather in Tokyo is sunny with a temperature of 22°C."*
- Multi-turn (7×8=56), single-image VL (reads "hello"), reasoning (no `<think>`/`<|im_start|>`/`[INST]` leaks), plain chat — all native.
- Regression: qwen3-4b tools unaffected (bridge + parser are Mistral-gated).

The bridge Mistral case mirrors the existing per-family selectors (Zaya/LFM2/Step37/Nemotron/Gemma3); no other family is touched.

PR #1690, opened 2026-06-25
Tolerate and reap corrupt coordinator lock files

jax-0n-git:
## Problem

`CoordinatorLockService.loadLock` decoded each lock file with `try decoder.decode(...)` and let a `DecodingError` propagate. One empty/truncated/corrupt lock file (a process crashing mid-write, a short write, external tampering) therefore broke `list()`, `reapExpired()`, and `coord status` for **every** resource.

Worse, `reapExpired()` calls `list()` first, so it threw before it could delete anything — the poison file could never be cleaned up automatically and required manual filesystem intervention.

## Fix

- `loadLock` now treats an undecodable file as absent (returns `nil`) instead of throwing, so `list()`/status skip it.
- `reapExpired()` garbage-collects undecodable lock files in addition to expired ones, so a poison file is removed rather than lingering forever.

## Testing

- New `CoordinatorLockServiceCorruptFileTests` plants a corrupt lock file next to a valid one and asserts `list()` returns only the good lock, `reapExpired()` does not throw, and the corrupt file is removed. It fails on the bug and passes with the fix.
- Full `OsaurusCLI` suite green, `swift-format lint --strict` clean.

### DIFF
--- a/Packages/OsaurusCLI/Sources/OsaurusCLICore/Services/CoordinatorLockService.swift
+++ b/Packages/OsaurusCLI/Sources/OsaurusCLICore/Services/CoordinatorLockService.swift
@@ -105,6 +105,7 @@ public struct CoordinatorLockService {
 
     @discardableResult
     public func reapExpired(now: Date = Date()) throws -> [CoordinatorLock] {
+        removeUndecodableLockFiles()
         let locks = try list()
         let expired = locks.filter { $0.isExpired(now: now) }
         for lock in expired {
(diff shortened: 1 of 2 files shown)


jax-0n-git:
One design note for reviewers: `reapExpired()` now garbage-collects any lock file that `loadLock` can't decode. Today that only ever matches genuinely corrupt/truncated files — a real I/O read error still throws and is left in place (the `do/catch` distinguishes a nil return from a throw). The one thing to be aware of going forward: if `CoordinatorLock`'s `Codable` shape is ever changed **incompatibly**, every previously-valid lock would become "undecodable" and get reaped on the next pass. For these short-TTL, single-user coordinator locks that's likely harmless (they'd just be re-acquired), so I kept the fix minimal — but if you'd prefer to harden against it I'm happy to gate the corrupt-file deletion on a minimum age (e.g. only reap undecodable files older than the longest TTL) or add a schema-version field. Let me know which you'd prefer.

PR #1700, opened 2026-06-25
Reject directory/symlink named *.dylib in dev-mode install detection

jax-0n-git:
## Summary

`InstalledPluginsStore` treats a version directory that has no `receipt.json` as a valid "dev mode" install when it contains a `.dylib`. Both detection sites used a bare `contentsOfDirectory(atPath:)` + `hasSuffix(".dylib")` string check with **no file-type test**:

- `installedVersions(pluginId:)` — the receipt-less fallback (`InstalledPluginsStore.swift:111`)
- `latestInstalledVersion(pluginId:)` — the `current`-symlink fast path (`InstalledPluginsStore.swift:167`)

Because the check is pure string matching, a *subdirectory* (or symlink) merely named `Foo.dylib` — with no real loadable binary — is accepted as a valid version. That false positive then flows into `allInstalledPluginIds()`, `latestInstalledVersion()`, and `repairDanglingCurrentSymlinks()`, which can repoint `current` at a binary-less directory.

The real install path already guards the identical decision behind `PluginInstallManager.isRegularPayloadFile` (`.isRegularFileKey` && `!.isSymbolicLinkKey`), whose doc comment warns that symlinked payloads are unsafe. This mirrors it: a private `containsLoadableDylib(in:)` helper that only accepts a regular, non-symlink `*.dylib` file, used at both sites.

## Changes
- [x] Bug fix
- [x] Tests

## Test Plan
`swift test` in `Packages/OsaurusRepository` (27 tests green, +3 new) + `swift-format lint --strict` clean. New regression tests: a directory named `Plugin.dylib` is rejected at both the `installedVersions` and `current`-symlink paths (each fails on the pre-fix logic, passes after), while a regular `.dylib` dev-mode install is still detected (guards against over-correction). No MLX/GUI dependency.

### DIFF
--- a/Packages/OsaurusRepository/InstalledPluginsStore.swift
+++ b/Packages/OsaurusRepository/InstalledPluginsStore.swift
@@ -78,6 +78,31 @@ public final class InstalledPluginsStore: @unchecked Sendable {
         return receipt
     }
 
+    /// Returns true if `directory` directly contains a regular (non-symlink) file with a
+    /// `.dylib` extension. Mirrors `PluginInstallManager.isRegularPayloadFile` so a
+    /// directory or symlink merely *named* `*.dylib` is not mistaken for a loadable plugin
+    /// binary by the dev-mode ("no receipt.json") validity check.
+    private func containsLoadableDylib(in directory: URL) -> Bool {
+        let fm = FileManager.default
+        guard
+            let urls = try? fm.contentsOfDirectory(
+                at: directory,
+                includingPropertiesForKeys: [.isRegularFileKey, .isSymbolicLinkKey],
+                options: [.skipsHiddenFiles]
+            )
+        else {
+            return false
+        }
+        return urls.contains { url in
+            guard url.pathExtension == "dylib",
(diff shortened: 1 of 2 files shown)


jax-0n-git:
One behavioral note for reviewers: `containsLoadableDylib(in:)` lists the version directory with `contentsOfDirectory(at:includingPropertiesForKeys:options:[.skipsHiddenFiles])` and matches on `pathExtension == "dylib"`, whereas the replaced code used `contentsOfDirectory(atPath:)` + `hasSuffix(".dylib")`. Two edge-case deltas follow: a *hidden* payload like `.foo.dylib`, and a file literally named `.dylib` (empty basename), are no longer counted as dev-mode installs.

This is intentional — it matches the real install-time payload detector `PluginInstallManager.findFirstDylib`, which already uses `.skipsHiddenFiles` + `pathExtension == "dylib"` + `isRegularPayloadFile`. So the dev-mode validity check is now consistent with how an actual payload is discovered. Real receipts store `dylib_filename = finalDylibURL.lastPathComponent` (normal, non-hidden names), so no valid install regresses. Happy to drop `.skipsHiddenFiles` if you'd rather keep the old hidden-file behavior.

PR #1708, opened 2026-06-25
Add eval watcher scoreboard artifacts

mimeding:
## Summary

- Adds a watcher-oriented eval workflow that stores local and frontier report bundles and refreshes a cross-run scoreboard.
- Uses the unified evidence registry as the report source of truth and keeps historical comparison data while gating only the selected release candidate.
- Supports a configurable regression allowance without hiding failed, errored, zero-case, or all-skipped runs.
- Fails closed on missing, corrupt, incomplete, mismatched, or unavailable artifacts and preserves infrastructure exit errors.

## Reliability details

- Release-candidate selection verifies both the validated artifact ID and the canonical artifact directory.
- Reused artifact IDs prefer the newest registered report instead of preserving stale evidence.
- Baseline comparisons gate coverage loss and severity escalation, including pass-to-skip, failed-to-error, and skipped-to-failure transitions.
- Raw reports are stored by role and model so local and frontier lanes cannot overwrite each other.
- Watcher cancellation stops the active child before refreshing the scoreboard.
- Channel and artifact identifiers are restricted before being used in paths or status JSON.

## Validation

- `swift test --package-path Packages/OsaurusEvals`: 195 tests passed, 3 resource-dependent tests skipped.
- Focused final report and scoreboard suites: 18 tests passed.
- `shellcheck --severity=warning scripts/evals/eval-watcher-report.sh`
- `git diff --check`
- Makefile plan-only smoke verified a single stable timestamped output directory.

## Risk

This changes eval artifact generation and release-gate reporting only. It does not change prompts, tools, model routing, runtime defaults, or the agent loop. Live local/frontier model runs remain an on-demand operational requirement; fixture and unit validation does not claim live model evidence.


mimeding:
Moving this back to draft because the branch is now dirty against current main even though the required checks were green. I will refresh the branch before asking for review again so the ready queue only contains merge-clean PRs.

tpae (approved):

PR #1710, opened 2026-06-25
Add Business Document I/O Workbench

mimeding:
## Summary

- Expands Business Document Studio into an import, preview, structured-extraction, and export workbench for PDF, document, spreadsheet, presentation, and delimited inputs.
- Surfaces unsupported formats, validation failures, and missing emitters explicitly.
- Keeps attachment handoff on the existing chat attachment API and labels the current handoff as validation-only rather than claiming prompt injection.
- Stages exports in isolated temporary locations, serializes document transitions, propagates cancellation, and uses atomic destination ownership.

## Security and correctness

- Rejects destinations outside the allowed export directory, including existing and dangling symlink escapes.
- Canonicalizes absolute and relative symlink targets, dot components, missing path tails, and the macOS `/var` to `/private/var` alias.
- Prevents a destination created during rendering from being overwritten without consent.
- Preserves structured-package extension ownership and rejects unsafe fallback exports.
- Explicit user document exports remain outside #2078 chat-session Changes/Undo tracking; agent file writes in combined mode remain tracked by #2078.

## Validation

- 43 business-document workflow tests passed before the final containment hardening.
- The complete affected service suite passed 17/17 after the containment fix; the failing baseline was reproduced with 11 issues before repair.
- 64 current-main execution, change-tracker, tool-resolution, and sandbox integration tests passed.
- Scoped strict SwiftLint passed with 0 violations.
- Localization and `git diff --check` passed.
- Unsigned Release app build passed after the current-main rebase and final fix.

## Draft gate

This PR remains draft until the native workflow is proven with screenshots and manual interaction:

- real open and save panels;
- document switching and preview click-through;
- export artifact surfacing in the app;
- unsupported and missing-emitter states;
- attachment handoff into an actual chat.

Path validation is fail-closed, but a writable intermediate parent can still be swapped between path validation and rename. Eliminating that residual TOCTOU requires directory-descriptor-relative filesystem operations rather than additional string/path checks.

### DIFF
--- a/App/osaurus/osaurusApp.swift
+++ b/App/osaurus/osaurusApp.swift
@@ -95,6 +95,14 @@ private extension osaurusApp {
         CommandGroup(after: .newItem) {
             Divider()
 
+            Button {
+                BusinessDocumentStudioLauncher.openDocumentPicker()
+            } label: {
+                Text(verbatim: L("Open Business Document..."))
+            }
+
+            Divider()
+
             Button(vadToggleLabel) {
                 toggleVAD()
             }
--- a/Packages/OsaurusCore/Resources/Localizable.xcstrings
+++ b/Packages/OsaurusCore/Resources/Localizable.xcstrings
@@ -29800,6 +29800,9 @@
     "Business Document Studio" : {
       "shouldTranslate" : false
     },
+    "Business Document Workbench" : {
+      "shouldTranslate" : false
+    },
     "Bundle" : {
       "localizations" : {
         "de" : {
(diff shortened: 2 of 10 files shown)


mimeding:
Revision update: rebased this draft PR onto current main and folded in the document-handoff/export hardening pass. The workbench now validates the structured attachment payload honestly instead of implying a chat/workspace insertion happened, guards against stale slow parses overwriting newer imports, and blocks misleading structured exports such as workbook bytes written to .txt or extensionless paths.\n\nLocal validation: focused BusinessDocumentStudio suite passes (31 tests), i18n check passes, and diff hygiene passes. A full keychain-disabled core test run hit the existing Keychain-sweep test because that test intentionally verifies real Keychain persistence, so I did not treat that as document-workbench signal. This PR remains draft while GitHub checks rerun.

PR #1712, opened 2026-06-25
Add `osaurus serve --supervise` keep-alive loop to survive app quit

jax-0n-git:
## Motivation

The Osaurus server — and the in-app scheduler that only ticks while the app is running — lives inside `Osaurus.app`. `osaurus serve` is a one-shot *client*: it launches the app, requests serving, polls `/health` once, and exits. If the app later **quits, crashes, or the machine logs out**, nothing brings the server back. There's no `launchd` / LaunchAgent backbone anywhere in the app, so an Osaurus you rely on for background/scheduled work silently stops the moment the GUI app isn't running.

## What this adds

A long-lived **supervise mode**:

```
osaurus serve --supervise [--interval N] [--port N] [--expose --yes]
```

`--supervise` runs a loop that probes `/health` every `--interval` seconds (default 15) and, whenever the server is down, relaunches it via the **same** launch + serve-notification path the one-shot mode already uses. It never exits, so it can run under a `launchd` **KeepAlive** LaunchAgent: launchd keeps the supervisor alive, the supervisor keeps the server alive — giving quit/crash/logout/reboot resilience. `--port` / `--expose` are honored exactly as in one-shot mode.

## Design

The loop is a small `ServerSupervisor` value (`OsaurusCLICore/Services/ServerSupervisor.swift`) whose collaborators — health check, relaunch, sleep, log — are injected, so the behavior is unit-testable without a real server, app, or wall-clock sleep. `serve` constructs the production supervisor (real `ServerControl.checkHealth` + `AppControl.launchAppIfNeeded` + serve notification) and runs it. Everything is in the light, MLX-free `OsaurusCLI` package.

## Deploying as a LaunchAgent

`~/Library/LaunchAgents/ai.osaurus.serve.plist`:

```xml
<dict>
  <key>Label</key>            <string>ai.osaurus.serve</string>
  <key>ProgramArguments</key> <array>
      <string>/opt/homebrew/bin/osaurus</string>
      <string>serve</string><string>--supervise</string>
  </array>
  <key>RunAtLoad</key> <true/>
  <key>KeepAlive</key> <true/>
  <key>StandardOutPath</key>  <string>/tmp/osaurus-serve.out.log</string>
  <key>StandardErrorPath</key> <string>/tmp/osaurus-serve.err.log</string>
</dict>
```

```sh
launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/ai.osaurus.serve.plist
```

(If you'd like a `osaurus serve install-agent` / `uninstall-agent` convenience command to generate and load this plist, happy to add it as a follow-up.)

## Tests

`ServerSupervisorTests` drives the loop with injected collaborators and a bounded iteration count: it relaunches on every down tick, never relaunches while healthy, and stops relaunching once the server recovers. Full `OsaurusCLI` suite stays green (91 tests). `swift-format lint --strict` clean. Verified end-to-end: `osaurus help` shows the new flags and the executable builds.

Mutation-checked: inverting the down-detection makes the relaunch tests fail, confirming they pin the behavior.

### DIFF
--- a/Packages/OsaurusCLI/Sources/OsaurusCLI/OsaurusCLI.swift
+++ b/Packages/OsaurusCLI/Sources/OsaurusCLI/OsaurusCLI.swift
@@ -105,9 +105,13 @@ struct OsaurusCLI {
             osaurus - CLI for Osaurus
 
             Usage:
-              osaurus serve [--port N] [--expose] [--yes|-y]
+              osaurus serve [--port N] [--expose] [--yes|-y] [--supervise] [--interval N]
                                       Start the server (default: localhost only). If --expose
                                       is set, a warning prompt will appear unless --yes is provided.
+                                      With --supervise, run a long-lived keep-alive loop that
+                                      relaunches the server whenever it goes down (probe every
+                                      --interval seconds, default 15) — intended to run under a
+                                      launchd KeepAlive LaunchAgent so the server survives quit.
               osaurus stop            Stop the server
               osaurus mcp [--access-key KEY]
                                       Run MCP stdio server proxying to local HTTP. When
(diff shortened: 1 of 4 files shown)


tpae (approved):

PR #1713, opened 2026-06-25
Don't treat a bare `tools uninstall` name as a CWD path (avoid deleting the wrong directory)

jax-0n-git:
## Problem

`tools uninstall <target>` resolves `target` as a filesystem path **first**:

```swift
let tURL = URL(fileURLWithPath: target)
var isDir: ObjCBool = false
if fm.fileExists(atPath: tURL.path, isDirectory: &isDir), isDir.boolValue {
    dirToRemove = tURL          // <- matched before the tools-root / plugin-id lookup
}
```

`URL(fileURLWithPath:)` resolves a **relative** string against the current working directory. So `osaurus tools uninstall foo`, run from any directory that happens to contain a subdirectory named `foo`, matches `<cwd>/foo` and `removeItem`s it — **deleting an unrelated local directory instead of the installed plugin, with no confirmation.** A normal plugin-id argument can trigger silent data loss.

## Fix

Gate the path interpretation behind an explicit path indicator — a leading `/`, `./`, `../`, `~`, or any embedded `/` — mirroring `ToolsInstall`'s existing convention (`src.hasPrefix("/") || src.hasPrefix("./") ...`). A bare name can then only resolve to a folder under the tools root or a plugin whose `receipt.json` id matches. Explicit paths (`./plugin`, `/abs/path`, `~/path`) still work exactly as before.

The resolution is extracted into a testable `resolveTargetDirectory(_:root:fileManager:)` (the old logic lived inline in `execute()`, which `exit()`s and isn't unit-testable).

## Tests

Adds `ToolsUninstallTests` (OsaurusCLI is a light, MLX-free package): a bare name must not resolve to a same-named CWD directory; explicit absolute and `./name` paths still resolve; a folder under the tools root resolves by name. `tools uninstall` had no prior test coverage.

3-state verified: removing the gate (unconditional path mode) makes the bare-name test resolve `collide` to `<cwd>/collide` — the exact directory that would be deleted — and the fix makes it return nil. `swift-format lint --strict` clean; full OsaurusCLI suite green.

### DIFF
--- a/Packages/OsaurusCLI/Sources/OsaurusCLICore/Commands/Tools/ToolsUninstall.swift
+++ b/Packages/OsaurusCLI/Sources/OsaurusCLICore/Commands/Tools/ToolsUninstall.swift
@@ -15,55 +15,7 @@ public struct ToolsUninstall {
         }
         let fm = FileManager.default
         let root = Configuration.toolsRootDirectory()
-        var dirToRemove: URL?
-        // If target looks like a path and exists, use it
-        let tURL = URL(fileURLWithPath: target)
-        var isDir: ObjCBool = false
-        if fm.fileExists(atPath: tURL.path, isDirectory: &isDir), isDir.boolValue {
-            dirToRemove = tURL
-        } else {
-            // Try direct folder under Tools root (directory name is plugin_id)
-            let candidate = root.appendingPathComponent(target, isDirectory: true)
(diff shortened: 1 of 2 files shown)


tpae (approved):

tpae:
hey @jax-0n-git , thank you for your contribution. Are you on Twitter? happy to tag you for the release

jax-0n-git:
Hey, @tpae - happy to help. I'm on Twitter as @Jungian_coach

PR #1726, opened 2026-06-26
Add authenticated GitHub plugin import

mimeding:
## Summary

Makes large GitHub plugin and skill imports practical while tightening authentication, checkpoint, supply-chain, and URL safety.

- Uses the shared Keychain-backed GitHub credential already used for plugin browsing and updates, with explicit in-memory and environment fallbacks.
- Adds an Advanced secure token entry in the import sheet with clear shared-credential semantics.
- Uses bounded recursive Git tree discovery and fails closed on truncated, duplicate, oversized, unknown-size, or excessively deep trees.
- Resumes interrupted manifest discovery from atomic checkpoints bound to marketplace, source-tree, and manifest fingerprints.
- Uses collision-resistant checkpoint identities and refreshes recursive trees at every top-level import so retries cannot validate stale checkpoints against cached data.
- Rejects ambiguous GitHub URLs, unsafe source paths/refs, private or credential-bearing imported MCP URLs, and alternate IPv4 encodings used to bypass SSRF checks.
- Redacts GitHub tokens and secret-bearing URLs from user-facing errors and never serializes credentials into checkpoints.

## Security

- One shared GitHub Keychain entry; no duplicate importer credential store.
- Per-request authentication resolution prevents stale session headers.
- Explicit file-count, byte-count, depth, entry-count, and per-file limits.
- Checkpoint manifest substitution, repository/ref filename collisions, stale same-session retries, tree truncation, duplicate paths, token leakage, URL ambiguity, path traversal, and encoded private-address cases are covered by tests.
- Clearing the credential in the import sheet explicitly clears the shared token used by imports and plugin updates.

## Validation

- 45 token, shared-auth, and Keychain tests passed.
- 130 plugin/import/SSRF/MCP security tests passed; one environment-dependent live test skipped.
- 24 focused GitHub import/checkpoint tests passed after final cache-lifecycle hardening.
- Localization validation passed.
- `git diff --check` passed.
- Unsigned Release CLI and app build passed.
- Final security reviews found no blocking issues.

## Known boundaries

- DNS rebinding requires connect-time transport enforcement outside this importer’s string-level URL policy.
- Checkpoint hashes detect stale or inconsistent cached content; they are not an HMAC against a local user who can already modify Osaurus application data and installed plugins.
- GitHub environment tokens remain process-scoped fallbacks when no saved shared token is available.


tpae:
Does it store the token in the keychain?

mimeding:
Good question. This PR does not store the GitHub token in Keychain. It only reads an explicit in-memory token provider or `GH_TOKEN` / `GITHUB_TOKEN` from the process environment and uses that for GitHub API requests.

The checkpoint/progress state does not include token material, and diagnostics redact token-like values before surfacing errors. If you prefer a persistent credential path, I can keep this importer stateless and add a separate, explicit follow-up for an importer-scoped Keychain token setting.

tpae:
Normal users won't be able to pass in the environment, but if we could show advanced settings under import where user can copy/paste keys (persist in keychain), would be helpful

mimeding:
Revision update: added importer-scoped GitHub token storage for the import sheet's Advanced section. The saved token lives in a dedicated Keychain service separate from plugin/MCP/provider secrets, is never displayed back in the UI, and resolves after explicit/test tokens but before GH_TOKEN/GITHUB_TOKEN. I also added security coverage for control-character rejection, disabled-Keychain behavior, token-source precedence, checkpoint/error no-leak behavior, and an import-layer test proving the saved token is sent as the Authorization header.\n\nI moved this PR back to draft while the new GitHub checks run. Once checks are green and merge state is clean again, it can go back to ready for review.

Issue #1714, opened 2026-06-25
Feature request: native durable multi-agent task board (Kanban-style orchestration) for spawned/remote agents


jax-0n-git:
### 1. The gap

Osaurus already has the two hard halves of multi-agent orchestration:

- **Agent spawn / delegation** — the `spawn` and `local_delegate` tools, plus the recently-merged delegation/remote-agent work (#1682/#1676/#1657/#1656/#1636/#1628/#1623).
- **An always-on runtime to host it** — the proposed `osaurus serve --supervise` (#1712, my open PR) would make the server survivable under launchd KeepAlive — the precondition for any agent to *reliably* hand work to another.

What's missing is the **durable connective tissue between spawns**: a shared, observable task board that survives app restarts and records what each agent is doing. Today, when a parent agent fans out to children, there's no first-class place that holds:

- **roles** — which agent/persona owns which unit of work;
- **statuses** — a lifecycle (queued → running → blocked → review → done) you can watch;
- **dependencies** — "child B can't start until child A finishes";
- **an audit trail** — what was claimed, by whom, when, and how it ended;
- **crash recovery** — a worker that dies mid-task gets reclaimed instead of silently stranding the work.

`spawn` answers "start an agent." It doesn't answer "give me a board where 6 agents pull work, never collide on the same item, and I can see at a glance which one is wedged." That's the orchestration-observability layer, and it's exactly the part that gets painful precisely *because* the delegation/p2p work is succeeding — more concurrent agents means more need for a durable, shared, inspectable queue.

### 2. Precedent — Hermes ships this and it works

[NousResearch/hermes-agent](https://github.com/NousResearch/hermes-agent) has a production Multi-Agent Kanban that solves this exact problem, and several of its design choices map cleanly onto Osaurus's existing posture. (I run both Hermes and Osaurus locally; the specifics below are from reading Hermes' source — a v0.15.1 local clone, with upstream at 0.17.0, so minor details may differ, but the architecture is stable.)

[...]

PR #1727, opened 2026-06-26
Add durable agent task board

mimeding:
## Summary

Addresses #1714 by adding a durable local task-board foundation for future spawned and remote agent orchestration.

- adds task, event, run, and dependency models
- stores the board in an always-encrypted SQLCipher database
- records append-only task events
- supports lease-based atomic claims, stale lease reclaim, and crash-style recovery
- rejects dependency cycles
- documents the storage threat model and the non-goals for this foundation PR

This does not add agent teams, channel inboxes, remote execution, or UI workflows.

## Validation

- `git diff --check`
- `OSAURUS_DISABLE_KEYCHAIN_FOR_TESTS=1 OSAURUS_TEST_ROOT=/tmp/osaurus-test-task-board swift test --package-path Packages/OsaurusCore --filter 'AgentTaskBoardStoreTests|StorageDatabaseCatalogTests|StorageOptInEncryptionTests'`

## Review Notes

- Claude review found no blocking issues after building the package targets and running the relevant storage tests.
- Grok compact review found no visible blocking issue in the shown foundation diff.

## Risk

The board is storage/API only. Future callers should open it off the main thread because first open may touch the storage key. The docs state that SQLCipher protects offline database theft, not a compromised local process.

### DIFF
--- a/Packages/OsaurusCore/Identity/StorageKeyManager.swift
+++ b/Packages/OsaurusCore/Identity/StorageKeyManager.swift
@@ -370,6 +370,7 @@ public final class StorageKeyManager: @unchecked Sendable {
         let encryptedArtifacts = [
             OsaurusPaths.chatHistoryDatabaseFile(),
             OsaurusPaths.agentChannelMessagesDatabaseFile(),
+            OsaurusPaths.agentTaskBoardDatabaseFile(),
             OsaurusPaths.memoryDatabaseFile(),
             OsaurusPaths.methodsDatabaseFile(),
             OsaurusPaths.toolIndexDatabaseFile(),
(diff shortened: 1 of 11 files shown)


tpae:
This one has risk of diversion in the upcoming changes.

mimeding:
That makes sense. I intended this as a storage/API foundation only: encrypted task, event, run, and dependency records with lease/reclaim semantics. It does not add agent-team protocol behavior, routing, UI, channel inboxes, or remote execution.

If this risks pulling the upcoming orchestration design in the wrong direction, I am fine deferring or closing it and revisiting after the agent/channel protocol direction is settled.

mimeding:
Closing this draft intentionally to reduce the open PR queue and avoid advancing a conflicted implementation before the protocol design is settled. The task-board direction still makes sense, but it should come back after the local/remote agent interaction model is agreed in #1714 so it does not conflict with Agent Channels or future remote agents.

Issue #1737, opened 2026-06-28
Can't continue or export conversations after update to 0.20.9


prtc:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest `main` or release

### Describe the bug

After the last update 0.20.9 I can't continue nor export previous conversations. I see the conversation is intact when viewed in Memory pane as "Transcript turn" entries. The handle of the conversation is also visible when I am in Agent panel, but selecting it doesn't load the previous turns and Export create nearly empty files (only the Title, no transcripts turns). 

### Provider type

Remote provider

### Model name

Anthropic Sonnet 4.6

### Osaurus version / commit

v0.20.9

### macOS version

Tahoe 26.5.1

### Apple Silicon chip

M4

### Logs

```shell

```

PR #1749, opened 2026-06-29
fix continuing and exporting conversations after the encryption update

RaajeevChandran:
## Summary

addresses #1737

- fixed a regression where finished conversations could not be continued or exported after updating, showing only the title on export
- stopped silently dropping chat history writes when storage was deferred during the at-rest encryption migration or a key rotation
- queued those writes and re-flushed them once storage became ready instead of losing the turns
- repaired conversations already orphaned by the update by recovering their turns from the memory transcript and persisting them back on next open
- added tests covering the deferred-save flush and the orphaned conversation heal

## Changes

- [x] Behavior change
- [ ] Refactor / chore
- [ ] Tests
- [ ] Docs

## Checklist


universenz:
@RaajeevChandran does this fix/repair chat history that was lost after app restarts?

Issue #1719, opened 2026-06-25
Plugin importer is unauthenticated → large repos (e.g. anthropics/knowledge-work-plugins, 212 skills) instantly hit GitHub's 60/hr limit and can never import


jax-0n-git:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest release (0.20.9)

### Environment

- Osaurus 0.20.9, macOS 26.5.1, Apple M5 Max
- Settings → Plugins → **Import** ("Import from GitHub")

### Describe the bug

The plugin importer (Settings → Plugins → Import → "Import skills from any GitHub repository") **cannot import a large plugin repo** because it makes its GitHub API calls **unauthenticated** (60 requests/hour per IP). Importing a repo with many plugins/skills exhausts that budget in a **single** attempt and fails with:

> **Import Failed** — GitHub rate-limited this app. Try again in 59 min.

The key symptom that proves this is structural, not transient: **each retry restarts the ~59-minute countdown** (observed 59 → 52 → 59 across attempts). Waiting and retrying never succeeds, because the act of attempting the import re-exhausts the entire hourly budget before it can finish enumerating the repo.

### Reproduction

1. Settings → Plugins → **Import**.
2. Enter `https://github.com/anthropics/knowledge-work-plugins` (Anthropic's official knowledge-work plugins — **212 `SKILL.md` files** across 11 first-party plugins + partner integrations; 22 `plugin.json` manifests in-tree).
3. Continue.
4. → "GitHub rate-limited this app. Try again in 59 min." Retrying after any wait re-triggers it and resets the countdown.

Smaller repos (e.g. the individual `anthropics/claude-plugins-official` plugins) import fine because their file/skill count stays under the 60-call budget.

### Root cause (analysis)

The importer appears to walk the repo's contents per-file/per-skill via the unauthenticated GitHub REST API. A repo with 200+ skill files needs **>60 calls** to enumerate, which exceeds the unauthenticated **60 req/hr** limit in one burst. Independent confirmation: GitHub's `/rate_limit` core counter for this IP read healthy between attempts, yet Osaurus's importer still failed instantly — consistent with the importer blowing the budget within a single import and/or tripping the secondary (abuse) limit on a rapid burst.

### Suggested fixes (any one largely resolves it)

[...]

Issue #1718, opened 2026-06-25
MCP-provider tools need two turns: capabilities_load defers the schema to the next turn, so the agent loops on `todo` and the output stream corrupts (reproduces on cloud grok-4.3)


jax-0n-git:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest release (0.20.9)

### Environment

- Osaurus 0.20.9, macOS 26.5.1, Apple M5 Max (128 GB)
- Agent model: **`xai/grok-4.3`** (cloud) — a custom agent with a **manual tool selection** that includes an **MCP-provider tool** (`miyo_search`, a local stdio MCP server registered with `executionHost: host`)
- Autonomous Exec **enabled**; Sandbox **on** (see workaround re: sandbox off)

### Describe the bug

When a custom agent is asked to use an MCP-provider tool, the in-app Agent loop **stalls before ever calling the tool** and the final generation **corrupts**.

The sequence (verbatim trace below):

1. The agent correctly calls `capabilities_load {"ids":["tool/miyo_search"]}`.
2. The result says: *"Tool 'miyo_search' loaded and available. Note: loaded tools are callable NOW — call them by name with JSON arguments. **Their full schemas will appear in the tool list from the next user turn.**"*
3. Because the tool's **schema is not present this turn** (only "next user turn"), the model never constructs the `miyo_search` call. Instead it loops on the built-in `todo` tool **~14 times** (toggling the same item complete/incomplete).
4. After the loop, the final assistant turn **degenerates into corrupted output**: control-character-interleaved `reasoning:` markers and a leaked stats footer, e.g.:
   ```
   <FFFE>reasoning:The<FFFE>reasoning: user<FFFE>reasoning: asked<FFFE>reasoning: to ... <FFFE>stats:11;0.0000;stop=stop
   ```

**Sending a second user turn** (e.g. "call miyo_search now with query: water") — at which point the schema is finally in the tool list — makes the model call the tool correctly and answer. So the tool, the MCP connection, and the model are all fine; the failure is the **one-turn gap** between `capabilities_load` and schema availability for MCP-provider tools, plus the degenerate `todo`-loop + stream corruption that fills that gap.

### Reproduction (condensed trace)

```
User: Use miyo_search to find my dreams about water

Assistant → capabilities_load {"ids":["tool/miyo_search"]}
Tool     → {"ok":true,"result":{"text":"Tool 'miyo_search' loaded and available.\n"}}
[...]

Issue #358, opened 2026-01-20
Unsupported model type: hunyuan_v1_dense


agldk-dot:
### Problem statement

It would be great to support one of the best translation model yet **HY-MT1.5-7B-mlx-8Bit**

Error: Unsupported model type: hunyuan_v1_dense

### Additional context

https://huggingface.co/m-i/HY-MT1.5-7B-mlx-8Bit

Issue #1422, opened 2026-06-08
Osaurus doesn't follow XDG directory specifications or Apple specifications for data storage


lpar:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest `main` or release

### Describe the bug

Osaurus doesn't consistently follow [the XDG directory specifications](https://specifications.freedesktop.org/basedir/latest/) or [Apple's specifications](https://developer.apple.com/documentation/foundation/using-the-file-system-effectively) for where to store configuration files and application data. Instead, it dumps a `.osaurus` directory into the home directory.

### Steps to reproduce

1. Run osaurus
2. Look in home directory.

### Provider type

Local model (MLX)

### Model name

any

### Osaurus version / commit

0.19.15

### macOS version

26.5.1

### Apple Silicon chip

M2 Pro 32GB

### Xcode version

n/a

### Logs

```shell

```

jax-0n-git:
A couple of notes for anyone tracking this:

The location is **centralized**, not scattered — every path derives from `OsaurusPaths.root()` (`Packages/OsaurusCore/Utils/OsaurusPaths.swift`), and on current `main` `defaultRoot` still hardcodes `~/.osaurus`. So this is a single by-design choice, which is good news for fixing it cleanly.

It also looks like this is already being addressed in-flight by **PR #1573 ("Add app data location resolver and legacy fallback audit")**, open as of this writing — it adds an `AppDataLocationResolver` that resolves data/config/cache from Apple/XDG-standard candidates (`~/Library/Application Support/Osaurus`, `$XDG_DATA_HOME`/`$XDG_CONFIG_HOME`/`$XDG_CACHE_HOME` under `osaurus/`) with a read-only fallback to the legacy `~/.osaurus`, and rewires `OsaurusPaths.root()/.config()/.cache()` through it. Net effect: **fresh installs land in the Apple/XDG-standard location**, while existing installs keep reading `~/.osaurus` so nothing moves underneath them.

That "no automatic migration" stance is deliberate and was pre-established by the earlier audit work (PR #1474 / `docs/STORAGE.md` § Storage Location Standards): the Keychain DEK + HKDF-salt sidecar are paired with the root, sandbox tooling references `~/.osaurus` literally, and plugin/container trees can be many GB — so relocating an *existing* root is treated as an explicit data-safety decision rather than a silent move.

[...]

PR #1761, opened 2026-06-30
Fix sweep: Mistral chat fix + hunyuan_v1_dense + Rampart PII (repin) + plugin auth (#1719) + MCP triage

jjang-ai:
Consolidated, all-in-one sweep. Repins vmlx-swift `9e0b60f → 5b096a3c` to bring in three merged engine fixes, plus the osaurus-side auth fix, the MCP-issue triage, and an integration doc.

## Engine fixes (via the vmlx-swift repin)
- **Mistral chat looping / incoherence** (vmlx-swift #100). Family detection used `convertTokenToId("<|im_end|>") != nil` as an existence test, but that returns the `<unk>` id for absent tokens → tool-bearing Mistral was misrouted to a **ChatML/Gemma** template (`<|im_start|>`/`<|turn>`/`<think>`/`<tool_call>`), tokens it lacks → incoherent + never-stops looping. osaurus agent use always injects tools, so it hit ~every Mistral chat. Fixed by round-tripping all sentinel checks. Proven via `BENCH_TEMPLATE_SMOKE`: tool cases now render native `[SYSTEM_PROMPT]…[AVAILABLE_TOOLS]…[INST]…[/INST]`.
- **`hunyuan_v1_dense` support** (#358 / vmlx-swift #97). Live-proven EN→FR / EN→zh.
- **Rampart PII NER library** (vmlx-swift #98) + a 4-bit-quant load fix I added (it couldn’t load its own published model). Neural-output parity vs the reference proven exact.

## osaurus-side
- **#1719** — plugin importer authenticates via `GITHUB_TOKEN`/`GH_TOKEN` (60→5,000 req/hr); token-absent path byte-identical; 9-case unit test (CI-green).
- **`docs/VMLX_MODEL_INTEGRATION.md`** — how osaurus consumes vmlx models: `model_type` dispatch, layers/embeds, quant (auto via `Load.swift` for LLMs vs the standalone RampartPII path), tokenizer, chat-template/jinja sidecar precedence, and per-model wiring.
- **`docs/FIX_SWEEP_TRIAGE.md`** — MCP + crash issues already resolved on main (#1718/#416/#587/#1129/#1228/#443/#647/#662/#789/#615/#1496/#903/#823/#1422), several *fixed-but-unreleased*.

## Verification
- Mistral / hunyuan / Rampart: engine-level live proofs (template smoke; RunBench decode; reference parity) — see each vmlx PR.
- #1719: unit-tested, CI-green.
- Repin: pin-contract guardrail tests updated to the new SHA; CI `test-core` rebuilds osaurus against the new vmlx.

Note: full in-app dev-app E2E (multiturn coherence across compaction/reasoning/VL) still wants the Codex Computer Use TCC grant; engine-level proofs stand in where the GUI is blocked.


jjang-ai:
Expanding this into a **consolidated fix-sweep PR** covering multiple reproducible open issues (not just #1718). In-progress triage + fixes for: **#1718** (MCP capabilities_load schema gap — *fix landed*), and investigating **#1422** (XDG/Apple data dirs), **#903** (system prompts not injected), **#789** (Search tool never found), **#662** (Invalid request format), **#1719** (unauthenticated plugin importer). Each fix is being reproduced + root-caused before inclusion; will list per-issue status + tests as they land. Marking WIP.

jjang-ai:
@tpae good catch — you're right. My initial agent-run change mutated the rendered tools block mid-run, which is exactly the prefix-bytes change that busts paged-KV (the #1725 regression your #1760 reverted). I'm **dropping that approach** and rebasing this branch onto main so it picks up #1760's append-only `CapabilitiesLoadTool.loadedSchemaBlock` — since /agents/run appends the tool result to the conversation, the loaded schema reaches the model same-turn there too, with the `<tools>` prefix frozen (no KV reset). So #1718 is covered by #1760 across surfaces; no separate agent-run mutation needed. Will re-verify no lcp=0 reset and refocus this fix-sweep PR on genuinely-open issues (e.g. #1719 unauthenticated plugin importer). Thanks for the heads-up + the #1760 context.

PR #1796, opened 2026-07-02
Disconnect must not evict the BatchEngine (cold-load hangup crashed the server)

jjang-ai:
## Problem

Any client that disconnected while its request was still **prefilling** could kill the whole server with the next request. 100%-reproducible:

1. fresh server → streaming chat that cold-loads the model
2. hard socket close (~463 bytes in, mid-prefill)
3. immediate second request → **SIGABRT**, crash report shows two threads both inside `BatchEngine.startSoloFastPath → TokenIterator.prepare → asyncEval`:

```
AGX...CommandBuffer tryCoalescingPreviousComputeCommandEncoderWithConfig:
failed assertion 'A command encoder is already encoding to this command buffer'
```

This matches the production Sentry cluster on the same assertion (APPLE-MACOS-33 / APPLE-MACOS-D6).

## Root cause

Every disconnect hook routes to `ModelRuntime.cancelGeneration`, which called `Registry.shutdownEngine(for:)` — that both shuts down **and removes** the engine from the registry. The dropped request's prefill producer is not stopped by that (prefill had no cancellation points in vmlx), so it keeps encoding GPU work; the follow-up request then builds a **fresh `BatchEngine` on the same `ModelContainer`** and starts its own prefill. Two producers on the shared GPU command queue → Metal encoder assertion → abort. This is exactly the scenario the Registry's own single-flight comment warns about.

Warm-model disconnects never crashed because decode checks `Task.isCancelled` between tokens — the orphan exits within one token. Prefill was the wide-open window.

## Fix (two layers)

- **This repo**: `cancelGeneration` no longer touches the registry — it cancels the wrapper task (which terminates the vmlx stream and cancels the producer) and leaves the engine registered. The engine's solo-path guard then queues follow-up requests on the **same** engine until the producer has actually returned. No rebuild, no concurrent producers.
- **vmlx #111** (pin advanced to `822888f6`): chunk-level `Task.checkCancellation()` in all five chunked-prefill loops, so the orphan producer exits within one ~512-token chunk instead of encoding the rest of the prompt for a dead request. Unit-tested engine-side (cancelled prepare encodes zero chunks; live tasks bit-identical).

Either layer alone prevents the crash; together they also stop wasting GPU on dead requests and keep TTFT for the retry low.

## Proof

Live proof on the dev-built app follows in a comment: the deterministic repro before/after, warm-disconnect control, and a post-fix coherence check.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "822888f6f79079c5683676a13b1ac084dd798908"
+        "revision" : "634d11d3a9489cc09ffbf33ae1235d9aaf6b9821"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "634d11d3a9489cc09ffbf33ae1235d9aaf6b9821"
+        "revision" : "cee0f8e234a352fcc6d09fba9a78dd24c0b15238"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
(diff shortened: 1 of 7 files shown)


jjang-ai:
## Live proof at the final pin (cee0f8e2) — dev-built app

The disconnect crash needed **three stacked fixes**; each was uncovered because the previous fix changed the crash signature instead of eliminating it:

1. **This PR's serving fix** — `cancelGeneration` no longer evicts the engine (was: retry builds a second engine on the same container → double-prefill `'A command encoder is already encoding'` SIGABRT).
2. **vmlx #112** — `BatchEngine.shutdown` awaits its producers + drains before returning (was: the disconnect-triggered cold-load `unload` freed weights under the orphan producer → SEGV in AGX encoder creation; the model lease is held by the serving wrapper, not the producer).
3. **vmlx #113** — disk-cache restore evals serialized under `MLXCacheIOLock` (was: once the prompt had a warm disk entry, the orphan's restore evals raced the follow-up's input tokenization → `'Completed handler provided after commit call'` — the Sentry APPLE-MACOS-BV cluster).

(vmlx #111 chunk-level prefill cancellation bounds all the orphan windows.)

### Result

**12/12 disconnect iterations across 3 fresh-boot cycles survive** — cold-load disconnect + immediate retry (previously 100% fatal), warm disconnects, and warm-disk-entry disconnects. Every follow-up returned the byte-identical coherent temp-0 response, same server PID throughout each cycle.

### No-regression sanity on the same build

- stop-string truncation exact: `'one, two, three, four, '`
- 23,141-token disk-tier replay after fresh boot: **3.0s, correct answer** — restores still hit under the new serialization.

Production impact: this class maps to the Sentry clusters APPLE-MACOS-33 / D6 (encoder-already-encoding) and APPLE-MACOS-BV (completed-handler-after-commit).

PR #1828, opened 2026-07-04
Recover bare capability load ids

mimeding:
## Summary
- allow `capabilities_load` to recover a bare live tool, skill, or plugin id when it has exactly one match
- preserve closed-vocabulary behavior by rejecting ambiguous bare ids with exact retry ids
- reject disabled skills consistently through the normal skill loader path
- add focused coverage for bare tool, skill, plugin, ambiguity, case-insensitive, and whitespace id handling

## Validation
- `git diff --check`
- `OSAURUS_DISABLE_KEYCHAIN_FOR_TESTS=1 OSAURUS_TEST_ROOT=/tmp/osaurus-test swift test --package-path Packages/OsaurusCore --filter CapabilityToolsTests` (38 tests passed)
- `OSAURUS_DISABLE_KEYCHAIN_FOR_TESTS=1 OSAURUS_TEST_ROOT=/tmp/osaurus-test make test` was attempted locally but the full-suite runner stopped making progress after `queryTemplateEscapesInputWithoutAddingSiblingParameters`; run was interrupted and CI remains the required full-suite gate.

## Reviews
- Grok review completed after retry without unsupported effort flag and with a prompt-file diff: no blocking correctness or actor-isolation issues.
- Fable/Claude reviewed an earlier revision and found no blocking implementation issues; actionable test-isolation findings were addressed. Final Fable retry stalled twice at the 5-minute limit, so it was recorded as unavailable for the final diff rather than blocking this draft.

## Support Context
Addresses Wave 0 capability-load recovery from the Discord support plan for reports where agents could not successfully load Search/capabilities after discovery.

### DIFF
--- a/Packages/OsaurusCore/Tests/Tool/CapabilityToolsTests.swift
+++ b/Packages/OsaurusCore/Tests/Tool/CapabilityToolsTests.swift
@@ -262,7 +262,10 @@ struct CapabilitiesDiscoverToolTests {
                 let previousOverride = ToolConfigurationStore.overrideDirectory
                 ToolConfigurationStore.overrideDirectory = tempDir
                 try? FileManager.default.createDirectory(at: tempDir, withIntermediateDirectories: true)
-                defer { ToolConfigurationStore.overrideDirectory = previousOverride }
+                defer {
+                    ToolConfigurationStore.overrideDirectory = previousOverride
+                    try? FileManager.default.removeItem(at: tempDir)
+                }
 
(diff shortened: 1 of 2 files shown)


mimeding:
Superseded by #1810, which includes bare tool, skill, and plugin ID recovery plus canonical casing, ambiguity handling, agent grants, disabled-state checks, index fallback, and broader regression coverage. Closing this duplicate so #1810 remains the single capability-reliability review target.

PR #1841, opened 2026-07-04
Hydrate spilled document attachments in prompts

mimeding:
## Summary
- Fixes R015: spilled document/PDF attachments now hydrate from `documentRef` blobs when rebuilding user prompt text for follow-up turns.
- Keeps attached document XML escaping on the recovered blob text so hostile PDF text cannot forge wrapper tags or control markers.
- Avoids re-hydrating large document blobs on every agent-loop prompt rebuild by caching the rendered attached-document prompt block.
- Keeps injected context freeze logic from rendering whole document attachments just to decide whether a turn is multimodal.

## Support mapping
- Discord support plan R015: "Crash after first PDF prompt; second PDF crashes" / PDF second-prompt crash or exclamation-output class.
- Root cause addressed here: persisted large PDFs spill to `.documentRef`; the old prompt renderer only read inline `.document` content and silently omitted recovered PDFs on later prompt rebuilds.

## Validation
- `git diff --check`
- `OSAURUS_DISABLE_KEYCHAIN_FOR_TESTS=1 OSAURUS_TEST_ROOT=/tmp/osaurus-pdf-second-prompt-test swift test --disable-index-store --package-path Packages/OsaurusCore --filter ChatAttachmentSecurityTests`
- `OSAURUS_DISABLE_KEYCHAIN_FOR_TESTS=1 OSAURUS_TEST_ROOT=/tmp/osaurus-pdf-queued-send-test swift test --disable-index-store --package-path Packages/OsaurusCore --filter ChatSessionQueuedSendTests`
- `OSAURUS_DISABLE_KEYCHAIN_FOR_TESTS=1 OSAURUS_TEST_ROOT=/tmp/osaurus-pdf-spillover-test swift test --disable-index-store --package-path Packages/OsaurusCore --filter AttachmentSpilloverTests`

## Reviews
- Fable medium review: no blockers. Notes recorded for follow-up: missing-blob UX/telemetry, extra parity tests, and possible future boolean blob-existence checks to avoid media reads in the freeze guard.
- Grok review: no blockers. Notes recorded for follow-up: freeze-guard parity coverage and spilled structured-document prompt coverage.

### DIFF
--- a/Packages/OsaurusCore/Tests/Chat/ChatAttachmentSecurityTests.swift
+++ b/Packages/OsaurusCore/Tests/Chat/ChatAttachmentSecurityTests.swift
@@ -77,6 +77,45 @@ struct ChatAttachmentSecurityTests {
         #expect(message.contains("Summarize"))
     }
 
+    @Test func buildUserMessageText_hydratesSpilledDocumentRefs() async throws {
+        try await StoragePathsTestLock.shared.run {
+            let root = FileManager.default.temporaryDirectory.appendingPathComponent(
+                "osaurus-chat-document-ref-tests-\(UUID().uuidString)"
+            )
+            try FileManager.default.createDirectory(at: root, withIntermediateDirectories: true)
+            OsaurusPaths.overrideRoot = root
+            StorageKeyManager.shared._setKeyForTesting(
+                SymmetricKey(data: Data(repeating: 0x45, count: 32))
+            )
+            defer {
+                OsaurusPaths.overrideRoot = nil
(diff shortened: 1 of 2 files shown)


mimeding:
Superseded by #1945, which preserves repeated-turn document hydration while adding hash/size verification, bounded reads, path containment, explicit missing/corrupt states, transcript recovery, and export coverage. Closing this narrower implementation so there is one attachment-recovery review target. Please review and merge #1945 instead.

PR #1865, opened 2026-07-06
Fix residency-default doc, expose batch diagnostics in /health, log effective policy

mimeding:
## What

Three small observability/correctness fixes found during a runtime audit:

1. **Doc/code mismatch fixed** — `docs/INFERENCE_RUNTIME.md` said the idle-residency default is `Immediately`; the code default has been `ModelIdleResidencyPolicy.defaultWarm = .afterSeconds(900)` (15 minutes) since the residency policy landed. The doc now matches the code.
2. **`/health` gains a `batch_diagnostics` block** — prefix-cache hits/misses, disk-L2 hits/misses/stores, SSM companion counters, native-MTP model count and depth summary, queue depths, and TurboQuant compression count, aggregated across resolved BatchEngines (`MLXBatchAdapter.snapshotDiagnostics()`). `null` until the first engine exists.
3. **Server start logs the resolved inference policy** — one line with `maxBatchSize`, eviction policy, and idle-residency policy (stable tokens, not localized strings).

## Why

The Settings panel already renders the batch diagnostics, but external tooling can't see them. With `osaurus bench` (#1863) measuring TTFT and throughput, these counters are what let a before/after difference be *attributed*: a TTFT drop should correspond to `prefix_hits` incrementing; a decode speedup on an MTP bundle should correspond to `native_mtp_models` ≥ 1 rather than a silent AR fallback. The startup policy line serves the same goal for bug reports — every benchmark or issue can state exactly which knobs were in force without asking the reporter to screenshot Settings.

The doc fix removes a footgun: users reading INFERENCE_RUNTIME.md were told the server unloads models immediately by default, and could misattribute the (actual, 15-minute) residency behavior to a leak.

## Measurement

Documentation + observability only; no runtime behavior change. Verified by incremental build plus a live `/health` fetch on a local Release build showing the new block populated after one request.

## Risk & rollback

`batch_diagnostics` reads existing actor-isolated snapshot APIs already used by the Settings UI; worst case is a few extra awaits per `/health` call. Revert = one commit.


mimeding:
Adversarial-review hardening pushed (2256336c): the /health batch-diagnostics fetch now races a 1 s deadline and degrades to null (with a timeout marker) instead of wedging the liveness endpoint behind a sick engine actor; inert defaults-domain docs corrected.

PR #1887, opened 2026-07-06
Add a native text selection assistant

mimeding:
## Summary

- Captures selected text through macOS Accessibility when the focused field can be verified as safe, without changing the clipboard.
- Adds a compact selection assistant with Summarize, Rewrite, Extract actions, and Ask commands. The selected text is attached to the draft and is never sent automatically.
- Retains the existing assistant-message Copy action and does not modify text in the source application.

## Privacy and safety

- Shares one Accessibility text policy with Computer Use so secure-field and readable-role decisions cannot diverge.
- Password, secure, unknown, unavailable, and unverified fields fail closed without posting Cmd+C when the source app is known.
- Missing Accessibility permission has a distinct actionable result instead of being reported as an unverified field.
- Synthetic Copy is reserved for the rare case where macOS cannot identify a source application; its bounded transaction restores the complete clipboard baseline or refuses to publish the capture.
- Native and synthetic captures enforce the same 256 KiB UTF-8 limit.
- Diagnostics include only the source application, character/byte counts, route, and typed outcome. Selected text is never logged.
- The overlay opens at its UI deadline during native capture while preserving the short focus grace required after synthetic Copy.

## Validation

- `swift test --package-path Packages/OsaurusCore --filter ClipboardContentDiagnosticsTests` (35 tests)
- `bash scripts/i18n/check.sh`
- focused strict SwiftLint
- `git diff --check`
- unsigned Release CLI and app build with code signing disabled

## Manual verification

- Select text in TextEdit, invoke the configured chat hotkey, choose an action, review the generated draft, and send explicitly.
- Verify missing Accessibility permission reports the required recovery action.
- Verify empty selections, secure fields, and unclassified fields show typed outcomes without changing the clipboard.


mimeding:
Earlier validation completed successfully: focused diagnostics tests and SwiftLint passed, and CI run 29138348141 was green across test-core, test-cli, test-evals, shellcheck, and swiftlint. A current-main repair is now addressing delayed pasteboard updates and rapid repeated hotkey invocations before the PR is presented again for review.

hiSandog:
One edge case worth covering is a selection that changes the pasteboard only after the bounded wait expires, especially when the source app is slow or under load. A regression test with a delayed pasteboard update could verify that the overlay reports the unavailable state without later consuming that delayed selection as if it belonged to a subsequent invocation. It may also be useful to confirm that two hotkey invocations close together cannot let the first capture overwrite the second invocation’s diagnostic state.

PR #1894, opened 2026-07-06
Show live tool-call progress card during buffered tool generation

jjang-ai:
## What

Shows a live "Preparing tool call" card in the native chat while a local model is still generating a tool call, replacing the frozen typing indicator during a long buffered tool write (e.g. a large file).

Pairs with the vmlx-swift engine event in osaurus-ai/vmlx-swift#118 (`Generation.toolCallProgress`) — this PR is the consumer side that turns those deltas into UI.

## The issue (reproduced)

On the native chat with a local MLX model and a file-writing tool, while the model generates the tool call the chat shows the **typing indicator (bouncing dots) with no progress** for the whole duration of the write; the tool card only appears once the call is fully parsed.

Reproduced at the wire level against a local dev build (`gemma-4-e2b`, sandbox `write_file`, ~1.3 KB file):

```
events           : 8
SENTINEL LEAKS   : 0
tool_name        : write_file
args parse OK    : True  path='/tmp/osaurus_probe.py'  content_len=1318
biggest inter-event gap: 4.57s  (immediately before the tool_call delta)
```

That single **4.57s window with zero stream events** is the frozen UI. It scales to tens of seconds for a real file.

## Root cause

The engine (vmlx-swift) buffers the entire tool-call envelope and emits nothing until the call closes — correct, so a half-written envelope never leaks into visible content. During that window the assistant turn has no visible content and no `pendingToolName`, so `ContentBlock` falls back to the plain typing indicator (the `pendingToolName == nil` branch at `ContentBlock.swift:499`).

The live tool-call rendering added in #1683 assumed tool arguments stream incrementally; on the local path they don't, because the engine had no channel to surface in-flight progress. This is the consumer-side half of that gap — the engine event (vmlx-swift#118) is the other half.

## The change

Consumes the new engine event end to end:

- **`GenerationEventMapper`**: maps `Generation.toolCallProgress` → `ModelRuntimeEvent.toolCallProgress` and clears the prefill spinner (the model IS producing, just inside an envelope).
- **`ModelRuntime.streamWithTools`**: encodes it behind the shared `￾` sentinel via a new `StreamingToolCallProgressHint`, only before a tool is committed.
- **`ChatView`** native loop: decodes the hint and seeds a neutral `pendingToolName` placeholder, flipping the frozen dots to the existing shimmering pending-tool card. The committed `StreamingToolHint` name overwrites the placeholder when the envelope closes.
- **`ToolDisplayName`**: `pendingToolSentinel` + localized "Preparing tool call" label.

## Scoping (deliberately narrow — not a blanket change)

- **No wire change**: the sentinel is already dropped from every OpenAI-compatible / Ollama / Anthropic / Responses SSE writer (`StreamingToolHint.isSentinel`), verified live (`SENTINEL LEAKS: 0`). The HTTP API is byte-for-byte unchanged.
- The raw envelope text is **never rendered as message text** and **never committed as a tool call** — the progress branch only toggles the pending-tool display state. `pendingToolName` is display-only (memoizer key + typing-indicator-vs-card decision + card title); it is never executed or persisted, and the committed name overwrites the placeholder.
- Non-streaming (`respondWithTools`) drops the event, matching `.reasoning`/`.prefillProgress`.
- The pending-tool card view is unchanged (its existing shimmer signals liveness); no new block types.

## Pin

Bumps the vmlx-swift pin to the commit carrying `Generation.toolCallProgress`. To be finalized to the merged vmlx-swift commit before this merges.

## Testing

- **Wire (live, dev build)**: `write_file` tool call parses correctly (no regression), `SENTINEL LEAKS: 0` (wire unchanged), and the 4.57s buffered dead-window reproduced — the window this fills in the UI.
- **Unit**: `GenerationEventMapper` forwards each progress delta as `.toolCallProgress` while still surfacing the committed call as `.toolInvocation`; empty deltas produce no event.
- Native multi-turn UI verification (card appears during a large-file write; content + reasoning streaming unaffected) is the remaining live check.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "53840914f693e9e1305fbbacb1ecc8e5c1e9625f"
+        "revision" : "8d0ef7e115f40bf428d05eee6f6a21ab0ddffb65"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "bdf6a1555ee0a71c299e2868bd776a37046bed4b32bd2a923dda5a9558799da7",
+  "originHash" : "5cf19a8dd43fb459ec28796d00e256a9e088916ada0d1265e9c5c4603097e749",
   "pins" : [
     {
       "identity" : "aachartkit-swift",
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "53840914f693e9e1305fbbacb1ecc8e5c1e9625f"
+        "revision" : "8d0ef7e115f40bf428d05eee6f6a21ab0ddffb65"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.swift
+++ b/Packages/OsaurusCore/Package.swift
@@ -43,10 +43,14 @@ let package = Package(
         // eval/asyncEval/item + synchronize/clearCache to kill the Metal
         // concurrent-encoder crash class) and #117 (NormConventionResolver:
         // an unrecognized norm_convention defers to the vote instead of
-        // silently disabling the (1+weight) shift).
+        // silently disabling the (1+weight) shift). Now also carries the
+        // incremental tool-call envelope progress event (`Generation
+        // .toolCallProgress`) so the app can show a live "preparing tool call"
+        // card during a long buffered tool write (e.g. a large file) instead of
+        // a frozen typing indicator. Additive — existing consumers unaffected.
         .package(
             url: "https://github.com/osaurus-ai/vmlx-swift",
-            revision: "53840914f693e9e1305fbbacb1ecc8e5c1e9625f"
+            revision: "8d0ef7e115f40bf428d05eee6f6a21ab0ddffb65"
         ),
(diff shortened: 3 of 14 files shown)


jjang-ai:
### Live validation (dev build, `:1337`)

Exhaustive + stress sweep, all green — the core invariant is **0 `U+FFFE` sentinel leaks on the wire** in every case (the hint must stay internal to the native UI):

| scenario | result |
|---|---|
| tool call (gemma-4-e2b, qwen3-0.6b) | parses; args intact; 0 leaks |
| plain content streaming | streams; 0 leaks; `finish=stop` |
| multiturn after tool result | natural-language continuation (qwen3); 0 leaks |
| tools available, no call needed | no hallucinated call; 0 leaks |
| **reasoning + tool call** (qwen3 thinking-on) | reasoning streams (546 ch) **and** call fires; 0 leaks — `<think>` path uncorrupted |
| **concurrency ×4** tool-call streams | 4/4 ok; 0 leaks; app alive |
| **disconnect ×3** mid-buffer | app survives every abort; clean recovery after |
| multi-tool-call in one turn | 0 leaks |
| **model switch ×4** (gemma↔qwen3) | all clean; app alive (no switch corruption) |

The `#1866` GPU-stream/Metal-concurrency fix is preserved — the vmlx pin here is a direct descendant of it.

### Follow-up fix included (`c7753ba`)

A self-review found the seeded `pendingToolName` placeholder could persist if a turn was **cancelled mid-envelope**, ended on EOS/max_tokens inside the envelope, or the buffered envelope reclassified to plain text (no committed `.toolCall`) — leaving a stuck "Preparing tool call" card on the finalized turn. Fixed with a `defer` in `processStreamDeltas` that clears the placeholder on every exit path (no-op for genuine pending calls, since the committed tool name overwrites the sentinel first). Re-verified: no streaming regression.

### Note on overlap

This overlaps #1891, which implements the same tool-call-progress feature with a richer live file-diff preview. The two can't both merge as-is (duplicate `Generation.toolCallProgress` case + different vmlx pin + same ChatView cascade). Flagging so we consolidate into one — this branch is the minimal, safe-for-all-tools foundation (never renders raw envelope text); #1891's file-diff preview could layer on top for `write_file`.

PR #1902, opened 2026-07-07
Derive idle-residency default from RAM tier (2 min / 15 min / resident)

mimeding:
## What

Derives the **default** idle-residency policy from the host's RAM tier (via `ChipProfile`, #1861) instead of using a flat 15 minutes everywhere:

| Host RAM | Default policy |
|---|---|
| ≤ 16 GiB | unload after **2 minutes** |
| 16–128 GiB | **15 minutes** (unchanged — the existing `defaultWarm`) |
| ≥ 128 GiB | unload after **60 minutes** (`.never` stays an explicit Settings-only choice) |

Only the *default* moves. An explicit user selection always wins; strict single-model eviction, manual unload, `clearAll`, and memory cleanup all still override residency exactly as before; and a persisted policy survives decode untouched.

**Stacked on #1861** (first commit is the ChipProfile detection layer).

## Why

The flat 15-minute default is the wrong trade at both ends of the hardware range:

- On ≤ 16 GiB machines, idle model weights are the main driver of memory pressure — the jetsam class that `InferenceLoadCoordinator` exists to mitigate. Two minutes still covers rapid follow-up turns while releasing RAM well before the system starts compressing and swapping.
- On ≥ 128 GiB machines the trade inverts: idle RAM is abundant and the avoided reload is what the user feels. Today the 15-minute timer silently unloads, and the next request pays a full cold load. Per review feedback, the big-RAM default is a 60-minute window rather than `.never`, so weights always leave on their own eventually and memory-pressure relief never depends on user action.

## Measurement

On the dev machine (M5 Max, 128 GB):

- **What a reload costs** (the thing this default avoids on big machines): from the committed bench baseline (#1863), a cold load adds **4,485 ms vs 183 ms warm TTFT (~4.3 s)** for a small 4.1 GB model *with a hot page cache*; a 23 GB Qwen3.6 bundle measured ~1.9 s page-cache-hot and takes tens of seconds from cold storage. The 15-minute timer guarantees this cost after every idle gap.
- **Explicit-setting contract, live**: this machine has a persisted explicit policy (`{"mode": "after_seconds", "seconds": 900}` in `server.json`), and on the patched build `/health` correctly keeps showing `idle_unload_at` with the 900 s countdown — the tier default did **not** override it. Fresh installs (no persisted key) resolve through `ServerConfiguration.default`, which is where the tier default applies.
- **Tier mapping**: unit tests cover 8/16/24–96/128/512 GiB, the inclusive/exclusive boundaries, the decode round-trip of an explicit policy, and the hardware-independent rewrite of the previous `defaultsWarmForMultiTurnChat` test (which pinned 900 s and would otherwise fail on ≥128 GiB CI hosts).

## Risk & rollback

Higher idle RAM usage on big machines is the intended trade, bounded by the same eviction paths as before. The tier function keys on RAM only (always detectable); unknown chips are irrelevant to it. Users who prefer aggressive unloading pick any explicit policy in Settings, which wins unconditionally. Revert = two commits on this branch.

### DIFF
--- a/Packages/OsaurusCore/Models/Configuration/ServerConfiguration.swift
+++ b/Packages/OsaurusCore/Models/Configuration/ServerConfiguration.swift
@@ -328,8 +328,12 @@ public enum ModelIdleResidencyPolicy: Codable, Equatable, Hashable, Sendable {
         return .defaultWarm
     }
 
-    /// Settings picker presets.
+    /// Settings picker presets. Every value `tierDefault(physicalMemoryBytes:)`
+    /// can return must appear here, otherwise the Settings picker has no
+    /// matching tag on exactly the machines that tier targets (120 s is the
+    /// ≤ 16 GiB default).
     public static let presets: [ModelIdleResidencyPolicy] = [
+        .afterSeconds(120),
         .afterSeconds(300),
         .defaultWarm,
         .afterSeconds(1_800),
@@ -342,6 +346,8 @@ public enum ModelIdleResidencyPolicy: Codable, Equatable, Hashable, Sendable {
         switch self {
         case .immediately:
             return L("Immediately")
+        case .afterSeconds(120):
+            return L("2 minutes")
         case .afterSeconds(300):
             return L("5 minutes")
         case .afterSeconds(900):
--- a/Packages/OsaurusCore/Models/Configuration/ServerConfiguration.swift
+++ b/Packages/OsaurusCore/Models/Configuration/ServerConfiguration.swift
@@ -224,7 +224,8 @@ public struct ServerConfiguration: Codable, Equatable, Sendable {
             allowedOrigins: [],
             globalProxyURL: nil,
             modelEvictionPolicy: .strictSingleModel,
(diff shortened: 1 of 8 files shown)


tpae:
its implicit .never default on big-RAM machines undermines our critical-memory-pressure unload

mimeding:
Good catch — agreed that a *default* must never leave memory-pressure relief dependent on user action. Pushed 8092000: the ≥128 GiB tier now defaults to `.afterSeconds(3600)` instead of `.never`, so weights always leave on their own after an hour idle while the warm-window win (no reload between nearby sessions) is preserved. `.never` remains available as an explicit Settings choice only. Tests and the doc comment updated to record the rationale. PR description will be refreshed to match.

mimeding:
Adversarial-review fixes pushed (9a029690): the 2-minute tier default is now a Settings preset (the picker previously had no matching tag on exactly the machines it targets), and the legacy immediate-unload migration routes through the tier default.

PR #1905, opened 2026-07-07
Add measured per-model prefill step tuning (`osaurus bench --tune-prefill`)

mimeding:
## What

Adds **measured per-model prefill step tuning**:

- `osaurus bench --tune-prefill [--model <id>] [--candidates 512,1024,2048,4096] [--prompt-tokens 8192] [--runs 3]` measures the model's uncached TTFT at each candidate prefill step size and persists the winner (with provenance: chip, date, winning TTFT) to `~/.osaurus/config/prefill-tuning.json`.
- The runtime (`MLXBatchAdapter.generate`) applies the per-model value via a new `ModelPrefillTuningStore` with an mtime-checked read — one `stat(2)` per generation — so tuned values (and the sweep itself) take effect **without a server restart**.
- Precedence: measured per-model value → server-wide `concurrency.prefillStepSize` → vmlx default (512).

**Stacked on #1863** (extends the `osaurus bench` command).

## Why — a tier table was the original plan, and measurement rejected it

The original idea was to scale `prefillStepSize` by hardware tier (bigger chips → bigger chunks). The sweep says otherwise — on the *same* M5 Max / 128 GB:

| Model | 512 (default) | 1024 | 2048 | 4096 |
|---|---|---|---|---|
| gemma-4-E2B 4-bit — 8K warm TTFT | **592 ms** | 651 ms | 885 ms | 938 ms |
| gemma-4-E2B 4-bit — 32K TTFT | 2533 ms | **2470 ms** | 3167 ms | 4696 ms |
| Qwen3.6-35B-A3B MoE — 8K warm TTFT | 2620 ms | — | **2039 ms (−22%)** | — |
| Qwen3.6-35B-A3B MoE — 32K TTFT | 12366 ms | — | **9420 ms (−24%)** | — |

The optimum is **model-architecture-dependent, not chip-dependent**: the small dense model is fastest at vmlx's default 512 (bigger chunks cost up to +85% TTFT), while the 35B MoE gains 22–24% at 2048. Any global setting or hardware-tier table is wrong for someone. The only correct value is one measured for the (model, machine) pair — so that's what this PR ships.

## Measurement

- The table above (each cell = warm-model uncached-prefix TTFT via `osaurus bench` methodology, 2 runs; medians shown for gemma, both-run agreement within a few % for qwen).
- Live end-to-end: a real `--tune-prefill` run against a live server (Qwen3.6-35B MoE, 8K prompts, 2 runs/candidate) swept 512→4096 **without a single restart** — each candidate took effect per-request via the mtime re-read — and persisted the winner:

  ```
  step  512: median uncached TTFT 2967 ms
  step 1024: median uncached TTFT 3195 ms
  step 2048: median uncached TTFT 2685 ms
  step 4096: median uncached TTFT 2674 ms   ← persisted (~10% under the 512 default)
  ```

  Persisted record: `{"qwen3.6-35b-a3b-mxfp4-mtp": {"prefillStepSize": 4096, "chip": "Apple M5 Max", "measuredAt": "2026-07-07T01:39:13Z", "benchTTFTMs": 2674}}`. Run-to-run variance in this quick 2-run sweep was noticeable (the controlled 2-runs-each sweep in the table above showed 2048 and 4096 within noise of each other for this model); `--runs 3`+ is recommended for production tuning, and discarding each candidate's first run is a noted follow-up improvement.
- Unit tests cover the store: round-trip, case-insensitive model lookup (chat router lowercases names), mtime-triggered re-read after an external (CLI-process) write, and merge-write semantics preserving other models' records.

## Risk & rollback

The runtime change is a two-line precedence addition at the existing `prefillStepSize` plumbing point; with no tuning file present, behavior is byte-identical to today. A bad tuned value is bounded by the sweep's own measurement (it persists only what it measured as fastest) and is trivially cleared by deleting the model's entry from `prefill-tuning.json`. Revert = one commit.


mimeding:
Adversarial-review hardening pushed (3a2ddaf6): interrupted sweeps (Ctrl-C/SIGTERM/early-exit) now restore the pre-sweep record via signal handlers + a sidecar backup, and near-tie winners (within 3%) resolve toward the smaller step instead of persisting noise. 8 new tests.

PR #1911, opened 2026-07-07
Tier the MLX buffer-cache cap by RAM (RAM/10 small, RAM/6 capped 24 GiB big)

mimeding:
## What

Tiers the MLX freed-buffer cache cap (`Memory.cacheLimit`) by host RAM, replacing the single formula `min(max(weights/4, 1 GiB), min(RAM/8, 8 GiB))`:

| Host RAM | System bound | Rationale |
|---|---|---|
| ≤ 16 GiB | RAM/10 | tighter pool where jetsam is a real risk; a smaller reuse pool costs a little re-allocation, never correctness |
| 16–96 GiB | RAM/8 capped 8 GiB | **unchanged** — historical heuristic, verified bit-for-bit by test |
| ≥ 96 GiB | RAM/6 capped 24 GiB | the flat 8 GiB cap starved activation reuse big machines trivially afford; 24 GiB covers the largest shipped models' transient activations at long context, beyond which the cache mostly hoards |

The `weights/4` model-scaled bound still applies on every tier (a 4 GiB model never gets a 21 GiB pool). Policy is a pure, unit-tested function reading RAM from `ChipProfile`; escape hatch: `defaults write com.dinoki.osaurus ai.osaurus.perf.disableAdaptiveMemoryPolicy -bool true` restores the historical formula everywhere.

**Stacked on #1861.** Deliberately scoped down from the original plan: the resident-weights budget now follows the user-configurable `modelLoadRAMSoftThreshold` (not touched), and the prefix-cache `memoryPercent: 15` default is load-bearing for `shouldRepairLegacyCacheDefaults`' migration detection (also not touched).

## Why

Same thesis as #1902: one memory constant cannot be right for an 8 GB Air and a 512 GB Studio. The freed-buffer cache is the last purely-internal memory heuristic that was hardware-blind. On a 128 GB machine serving a 50 GB-class model, the cap moves 8 GiB → ~21.3 GiB of activation reuse; on an 8 GB machine it moves 1 GiB → 0.8 GiB of extra safety margin.

## Measurement

- Unit tests pin all three tiers, both boundaries, the mid-range historical-equivalence property, the weights/4 bound, and the 24 GiB ceiling (8/16/24–64/128/512 GiB).
- Mid-range machines are provably unaffected (historical-equivalence test), so the change is default-preserving for the majority of installs.
- The effect on big machines is a larger reuse pool for transient activations; the cap is advisory to MLX's allocator (freed-buffer retention), so the worst case is bounded by the cap itself and reversible via the defaults key.

## Risk & rollback

Buffer-cache caps affect allocator retention, not correctness. Escape hatch restores the historical formula; revert = one commit.

### DIFF
--- a/Packages/OsaurusCore/Networking/HTTPHandler.swift
+++ b/Packages/OsaurusCore/Networking/HTTPHandler.swift
@@ -8854,6 +8854,7 @@ final class HTTPHandler: ChannelInboundHandler, Sendable {
             let obj: [String: Any] = [
                 "status": "healthy",
                 "timestamp": Date().ISO8601Format(),
+                "hardware": ChipProfile.current.healthJSONObject(),
                 "loaded": loaded,
                 "current_model": current,
                 "inflight": inflightObj,
--- a/Packages/OsaurusCore/Services/ChipProfile.swift
+++ b/Packages/OsaurusCore/Services/ChipProfile.swift
@@ -1,0 +1,175 @@
+//
+//  ChipProfile.swift
+//  osaurus
+//
+//  Detects the Apple Silicon chip this process is running on (generation,
+//  tier, RAM, GPU core count, Metal working-set budget) so runtime policy can
+//  be derived from hardware capability instead of one-size-fits-all
+//  constants. Detection is read-only and resolved once per process; nothing
+//  in this file changes runtime behavior by itself.
+//
+//  Why not persist the result: every field is re-derivable in microseconds
+//  from sysctl/IOKit/Metal at launch, and a cached file would go stale when a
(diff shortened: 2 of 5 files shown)


mimeding:
Adversarial-review fixes pushed (2d059c34): the escape-hatch doc command now uses the correct defaults domain (com.dinoki.osaurus — the documented ai.osaurus form wrote a domain the app never reads), and the UserDefaults read is hoisted out of mlxCacheLimitBytes so the function is genuinely pure; new test pins the escape hatch restoring the historical formula.

PR #1913, opened 2026-07-07
Add ModelCapabilityLedger skeleton; migrate the production blocklist to it

mimeding:
## What

Adds `ModelCapabilityLedger` — measured, per-model capability records that feature gates consult before falling back to hardcoded name lists — and migrates its first consumer: the ZAYA1-VL JANGTQ_K production block in `MLXService.validateRuntimePolicy`.

- **Records** live at `~/.osaurus/config/model-ledger.json`, keyed by normalized model name, carrying verdicts (`pass`/`fail`/`untested`), block reasons, and provenance (`source`, weights `digest`, `chip`, `measuredAt`). Reads are mtime-checked (the `ModelPrefillTuningStore` pattern from #1905), so records written by the CLI apply without a server restart.
- **Seed rules** are compiled in and reproduce today's hardcoded behavior exactly — the ZAYA block decision is covered by tests across picker/router/repo name forms, so day-one behavior is bit-identical.
- **Measured beats static, both directions**: a `productionServing: pass` record clears a seed block without an app release (e.g. a future gauntlet run proving a fixed bundle); a measured `fail` blocks a bundle no seed knows (e.g. a broken finetune).

## Why

Every "does model X support Y" policy in the runtime is a hand-maintained name-substring list (hybrid families in `isKnownHybridModel`, production blocklists, per-family decode profiles). Names lie — renames and finetunes slip past them — and static claims lie too. Two measured examples from this PR series: an MTP tuning stamp promising 1.56× that the serving path doesn't deliver (vmlx-swift#120), and prefill step optima that turned out to be per-model rather than per-chip (#1905). The ledger gives those measurements a durable home that gates can read.

This is the skeleton PR: schema + store + seeds + one migrated consumer. Follow-ups per the Wave-2 plan add the `osaurus bench --gauntlet` probes (template-leak, tool-call, stop/EOS, degeneration canary, MTP byte-equivalence) that write these records, and migrate further name lists seed-first.

## Measurement

- 6 unit tests: seed reproduces the retired hardcoded check across all name forms (and does **not** block the recommended replacement bundles); measured pass clears a seed block; measured fail blocks unknown models; `untested` falls through to seeds; external writes picked up by mtime without restart.
- Zero behavior change by construction: with no ledger file present (every existing install), the only decision path is the seed rule, which is the old check verbatim.

## Risk & rollback

The one runtime change is swapping a hardcoded substring check for a function that consults an optional file then applies the same substring check. No file = identical behavior. Revert = one commit.


mimeding:
Rebased on current main: #1907 lifted the ZAYA JANGTQ_K block upstream, so the seed rules are now empty (with the mechanism kept as the extension point) and the tests assert main's actual behavior — no hardcoded blocks, measured ledger records as the primary path. CI should trigger now that the merge conflict is resolved.

mimeding:
Adversarial-review fix pushed (70198214): `save()` now merges at the raw-JSON level, preserving foreign records' fields (gauntlet `probes`/`evidence`, unknown future fields) byte-identically — previously a Core-side write would have stripped them file-wide. Test pins the preservation.

PR #1914, opened 2026-07-07
Add `osaurus bench --kv-matrix`: automated fp16-vs-TurboQuant KV codec evidence

mimeding:
## Summary

Adds `osaurus bench --kv-matrix` to generate reproducible per-model evidence comparing the engine-selected KV baseline with explicit TurboQuant 3/3 behavior across configured prompt sizes.

The workflow derives each probe from the canonical `/admin/runtime-settings` document, restarts Osaurus, verifies the effective persisted settings and loaded-model KV mode, runs matched uncached and cached samples, and restores the original settings byte for byte.

## Safety and evidence contract

- Changes no runtime defaults.
- Uses exact-byte compare-and-swap before every file mutation.
- Accepts formatting-only runtime rewrites but rejects semantic changes.
- Adopts verified persisted bytes as the next compare-and-swap baseline.
- Restores the original document, reloads and verifies it semantically, then re-pins the exact original bytes.
- Preserves a collision-safe backup for manual recovery after an uncatchable termination or failed restore.
- Requires runtime settings and cache telemetry to prove the requested codec before emitting a result.

## Validation

- 23 focused KV matrix tests passed.
- Full OsaurusCLI suite passed: 153 XCTest tests and 18 Swift Testing tests.
- Release CLI build passed.
- `git diff --check` passed.
- Scoped SwiftLint passed for the benchmark implementation. One unrelated pre-existing violation remains in an untouched section of `ServerController.swift`.

## Draft gate

This PR is intentionally draft. The implementation and safety contract are tested, but promotion still requires live evidence from a real local model matrix:

- real engine and TurboQuant cycles;
- `/admin/cache-stats` evidence;
- token/s and physical-footprint measurements;
- 8K and 32K prompt rows;
- normal, SIGINT, SIGTERM, and concurrent-settings restoration proof.

No performance claim should be inferred until those artifacts are attached.

### DIFF
--- a/Packages/OsaurusCLI/Sources/OsaurusCLICore/Commands/Bench.swift
+++ b/Packages/OsaurusCLI/Sources/OsaurusCLICore/Commands/Bench.swift
@@ -394,12 +394,18 @@ public struct BenchCommand: Command {
 
     enum KVMatrixError: LocalizedError {
         case malformedConfig(String)
+        case runtimeSettingsReadbackMismatch
+        case persistedSettingsReadbackMismatch
         case restoreFailed(String)
 
         var errorDescription: String? {
             switch self {
             case .malformedConfig(let detail):
                 return "server-runtime.json does not look like a runtime settings file (\(detail))"
+            case .runtimeSettingsReadbackMismatch:
+                return "runtime settings readback differs from the requested probe; Osaurus normalized or migrated additional fields, or another settings writer changed them"
+            case .persistedSettingsReadbackMismatch:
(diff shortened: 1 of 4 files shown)


mimeding:
Adversarial-review hardening pushed (bab27bd5): signal-safe config restore + sidecar backup (SIGKILL-recoverable), concurrent-writer detection before restore, and per-codec `codec_verified` in the JSON artifact so unverified codecs can never enter the evidence matrix as clean data. 6 new tests.

PR #1916, opened 2026-07-07
Add `osaurus bench --gauntlet`: measured model-capability probes writing the ledger

mimeding:
## What

Adds `osaurus bench --gauntlet`, a seven-probe model verification workflow:

1. Non-empty load/stream response
2. Template-marker leakage across content and reasoning channels
3. Forced tool call with exact function name and schema-valid arguments
4. Multi-character stop handling with an unstopped control run
5. Long-generation degeneration detection, including two-token loops and character runs
6. MTP-versus-AR comparison evidence
7. Prefill/decode performance snapshot

Results are field-merged into the capability ledger without deleting existing verdicts, block reasons, digests, or unknown fields.

## Safety contract

The gauntlet records per-probe evidence and a production-serving candidate result. It does not change `productionServing` in v1 because the runtime cannot yet bind that promotion to a verified weights digest. Failed or unproven critical probes exit nonzero. MTP equivalence remains `untested` unless the runtime can prove per-request MTP activation.

## Validation

- `BenchGauntletTests`: 20 passed
- `DegenerationDetectorTests`: 12 passed
- Scoped SwiftLint: passed
- `git diff --check`: passed
- Independent runtime review findings addressed: no vacuous stop/template passes, no stale ledger promotion, no destructive record replacement, and no two-token degeneration blind spot

## Risk and rollback

CLI and evidence-writing scope only. Runtime model behavior, prompts, templates, samplers, and generation defaults are unchanged. Reverting the single feature commit removes the workflow.

### DIFF
--- a/Packages/OsaurusCLI/Sources/OsaurusCLI/OsaurusCLI.swift
+++ b/Packages/OsaurusCLI/Sources/OsaurusCLI/OsaurusCLI.swift
@@ -130,6 +130,10 @@ struct OsaurusCLI {
               osaurus bench [--model <id>] [--prompt-tokens 1024,8192] [--runs 3] [--json <path>]
                                       Benchmark TTFT / prefill / decode against the
                                       running server; emits JSON tagged with hardware info
+              osaurus bench --gauntlet [--model <id>] [--json <path>]
+                                      Run the onboarding verification suite (measured
+                                      probes against the running server) and record
+                                      per-model results in the capability ledger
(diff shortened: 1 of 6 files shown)


mimeding:
Adversarial-review hardening pushed (e2753123): stop-sequence probe now requires positive evidence (counted to the boundary) and fails on leaked stop text; mtp-equivalence and the canary guard against vacuous empty-transcript passes via real usage token counts; --tune-prefill/--gauntlet mutually exclusive. 15 new tests.

PR #1917, opened 2026-07-07
Micro-batch local embeddings into one batched MLX forward

mimeding:
## What

Micro-batches the local embedding path into genuinely batched MLX forwards:

- **`EmbeddingBatcher`** (actor): the first request in an idle period opens a 25 ms window (later arrivals never extend it, so a lone caller's added latency is bounded by one window); up to 16 requests coalesce, with immediate flush when full; per-batch errors propagate to every waiter; cancelling one waiter removes only that caller. Clock and window are injectable for tests.
- **A true `[N, L]` batched forward** in `VMLXModel2VecEmbedder`: padded gather → masked mean → row L2-normalize → **one** `eval` per batch. This was necessary because vmlx-swift's `embed(texts:)` is secretly a sequential loop (`texts.map(embedOne)` — one gather+eval per text), so a coalescing front alone would have been a fake batch.
- All four VecturaKit call sites (memory/skill/tool/method search) pick up batching through the shared embedder with zero call-site changes; `/v1/embeddings` and the plugin `embed(texts:)` paths are pass-through.
- **MetalGate discipline preserved and improved**: one coalesced batch acquires the embedding gate once, instead of N sequential acquisitions each potentially queueing behind active generation.

## Why

Embeddings are the highest-frequency small inference in osaurus (memory search + distillation), and they serialize today — both at the gate and in vmlx's per-text loop. Batched execution amortizes the gate acquisition, kernel launches, and eval/`asArray` round-trips; the biggest practical win is under generation load, where each embedding previously re-queued behind the generation gate individually.

## Measurement

- **Reference parity against vmlx's original pipeline** (the implementation this replaces), on the real potion-base-4M bundle: max delta **1.49e-08** on the longest padded text, exactly **0.0** on the others including the empty string; unit norms verified. The reference dependency (`MLXEmbedders`) is test-target-only.
- One intentional convention difference, asserted explicitly on both sides so silent drift fails the test: empty *input array* returns `[]` here vs vmlx throwing `emptyBatch`.
- 9/9 batcher tests (batched==sequential via counting mock; N concurrent callers → exactly 1 backend call; lone-caller latency bound; error propagation; cancellation isolation; full-batch early flush) + 106 memory-suite regression tests green; build + swiftlint clean.
- The real-model parity test is opt-in (`OSAURUS_EMBEDDING_MODEL_DIR`) because MLX hard-aborts the test process when its metallib isn't locatable under plain `swift test` — the same reason no existing test runs MLX compute.

## Risk & rollback

Worst case for a lone embedding call is +25 ms (documented; negligible against gate-contention waits it replaces). The batched math is validated to 1e-8 against the prior implementation. A follow-up worth doing upstream: vectorize `Model2VecStaticEmbeddingPipeline.embed(texts:)` in vmlx-swift so the ~40 mirrored loader lines here can shrink back to a coalescing front. Revert = two commits.

### DIFF
--- a/Packages/OsaurusCore/Package.swift
+++ b/Packages/OsaurusCore/Package.swift
@@ -234,6 +234,10 @@ let package = Package(
                 "OsaurusCore",
                 "OsaurusSQLCipher",
                 .product(name: "VMLXJinja", package: "vmlx-swift"),
+                // Reference implementation for the embedding batch-parity
+                // test: the batched forward in VMLXModel2VecEmbedder must
+                // match vmlx's original sequential Model2Vec pipeline.
+                .product(name: "MLXEmbedders", package: "vmlx-swift"),
                 .product(name: "NIOEmbedded", package: "swift-nio"),
                 .product(name: "VecturaKit", package: "VecturaKit"),
             ],
--- a/Packages/OsaurusCore/Services/Memory/EmbeddingBatcher.swift
+++ b/Packages/OsaurusCore/Services/Memory/EmbeddingBatcher.swift
@@ -1,0 +1,216 @@
+//
+//  EmbeddingBatcher.swift
+//  osaurus
+//
+//  Micro-batches concurrent single-text embedding requests into one batched
+//  forward pass. Memory search and distillation fan out many single-text
+//  embeds (VecturaKit embeds each search query with `embed(text:)`), and
+//  without coalescing those callers serialize one-by-one through MetalGate.
+//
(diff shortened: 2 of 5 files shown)


mimeding:
Adversarial-review fix pushed (06f84bcc): the direct embed(texts:) pass-through now chunks client-controlled arrays to ≤16 per forward inside the single MetalGate acquisition — previously /v1/embeddings could materialize an unbounded [N, L_max, D] padded tensor (~8.6 GB for a hostile batch). Chunk-parity and mismatched-count tests added.

PR #1919, opened 2026-07-07
Add opt-in auto batch sizing driven by observed demand, tier-capped

mimeding:
## What

Opt-in **auto batch sizing** driven by observed demand, capped by hardware tier:

- **`BatchAutoscaler`** (actor + pure policy core): tracks a 60 s sliding window of observed concurrent demand per model; recommendation = 1 when no overlap, else `min(nextPowerOfTwo(windowMax), tierCap)`. Sticky during a busy window (mid-burst downsizing thrashes engine rebuilds); decays to 1 only after a full quiet window.
- **Tier caps from ChipProfile**: base/unknown → 1 (never escalate — the compiled-decode TTFT win outweighs concurrency that bandwidth can't feed), pro → 4, max → 8, ultra → 16.
- **Demand signal** recorded in `MLXBatchAdapter.generate` **before the solo lease** — deliberately: at `maxBatchSize == 1` the SoloGenerationGate serializes overlapping requests upstream of the engine, so post-lease counts can never exceed 1 and overlap would be structurally unobservable. One registry peek + one pending/active read pair, only when the flag is on.
- **Return-to-1 restores the compiled path correctly**: a hot-resize down does NOT recover compile eligibility (vmlx invariant), so the decay path shuts the engine down via the existing tombstoned `Registry.shutdownEngine` — only when idle (pending==0 && active==0), retrying while busy, off the request path — so the next request rebuilds fresh at 1 with compile eligibility.
- **Off by default**: `defaults write com.dinoki.osaurus ai.osaurus.scheduler.autoBatchSize -bool true`, and auto only engages when no explicit `maxConcurrentSequences`/legacy value is set (explicit settings always win). Flag off = behavior identical to today, covered by test.

**Stacked on #1861.**

## Why

`maxBatchSize` is the sharpest tradeoff in the runtime: 1 keeps vmlx's compiled decode (documented 9× TTFT on Mistral 3.5), >1 buys multi-client throughput but permanently loses compile for that engine. Today it's a static `defaults write` the user must reason about. The machine can decide instead: single-user chat stays on the compiled path; a machine that actually observes overlapping same-model requests escalates to what its memory bandwidth tier can feed, and returns to the compiled path when quiet.

## Measurement

- 18 autoscaler tests (tier caps incl. unknown; no-overlap → 1; overlap 3 → 4; cap respected; busy-window stickiness; quiet-window decay with injected clock) + MLXBatchAdapter/TaskCoalescer/ChipProfile regression suites: 90/90 green; build + lint clean on changed lines.
- Live multi-client throughput measurement deliberately deferred to the flag-flip evaluation: this PR ships the mechanism dark (flag off), so the measurable claim here is "no behavior change by default", which the tests pin.

## Risk & rollback

Flag-gated dark launch; explicit settings win; decay teardown reuses the existing tombstone-protected shutdown path and only fires on an idle engine (a straggler racing the shutdown gets vmlx's clean `.cancelled` + registry rebuild — the documented fail-closed behavior). Revert = one commit.

### DIFF
--- a/Packages/OsaurusCore/Networking/HTTPHandler.swift
+++ b/Packages/OsaurusCore/Networking/HTTPHandler.swift
@@ -8854,6 +8854,7 @@ final class HTTPHandler: ChannelInboundHandler, Sendable {
             let obj: [String: Any] = [
                 "status": "healthy",
                 "timestamp": Date().ISO8601Format(),
+                "hardware": ChipProfile.current.healthJSONObject(),
                 "loaded": loaded,
                 "current_model": current,
                 "inflight": inflightObj,
--- a/Packages/OsaurusCore/Services/ChipProfile.swift
+++ b/Packages/OsaurusCore/Services/ChipProfile.swift
@@ -1,0 +1,175 @@
+//
+//  ChipProfile.swift
+//  osaurus
+//
+//  Detects the Apple Silicon chip this process is running on (generation,
+//  tier, RAM, GPU core count, Metal working-set budget) so runtime policy can
+//  be derived from hardware capability instead of one-size-fits-all
+//  constants. Detection is read-only and resolved once per process; nothing
+//  in this file changes runtime behavior by itself.
+//
+//  Why not persist the result: every field is re-derivable in microseconds
+//  from sysctl/IOKit/Metal at launch, and a cached file would go stale when a
+//  Time Machine / Migration Assistant restore moves the install to different
+//  hardware. Persistence becomes worthwhile only for *measured* values
+//  (micro-benchmarked bandwidth/FLOPS), which are out of scope here.
+//
+
+import Foundation
+import IOKit
+import Metal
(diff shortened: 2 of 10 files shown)


mimeding:
Adversarial-review hardening pushed (f6fc60b3): idle-shutdown now requires a zero in-flight-handle count in the same actor slice (closing the TOCTOU that could turn a live request into an empty 200), and decay no longer hot-resizes an active engine down — shrink-to-1 happens only via the sweep's tombstoned rebuild, so compile recovery actually works as claimed.

mimeding:
Follow-up on f6fc60b3: the shutdown claim/park handshake goes beyond the original finding — a final same-slice re-check alone still left a window during the coalescer suspension where a racing engine(for:) could receive the doomed handle; the tombstone claim closes it. Real-Registry tests cover held/released/overlap counting.

PR #1921, opened 2026-07-07
Session-freeze composer inputs on /agents/{id}/run for KV prefix reuse

mimeding:
## What

When a `/agents/{id}/run` caller supplies a `session_id`, `enrichWithAgentContext` now runs the same session-freeze contract as the chat and plugin-host paths: validate the `(executionMode, toolMode)` fingerprint via `SessionToolStateStore`, echo the frozen snapshot (always-loaded tool names, enabled-capabilities manifest, SOUL content, mid-session loaded tools) into `composeChatContext`, and record the first compose's snapshot. Without a `session_id` the endpoint stays stateless, byte-for-byte unchanged.

50-line diff mirroring the existing, proven `PluginHostAPI.resolveAgentContext` pattern — the store, the freeze fields, and the fingerprint invalidation all already exist; this path was simply never wired to them (the code comment even said so).

## Why

The disk-L2 KV cache matches on an **exact hash of the token prefix** (`DiskCache.hashTokens`: SHA-256 over modelKey + token ids, probed at stored prompt boundaries) — one drifting byte in the system prompt or `<tools>` block moves the divergence point to token 0 and forfeits all reuse. The chat path freezes its composer inputs per session for exactly this reason; the plugin path gained the same in its store consolidation. The HTTP agent-run path recomposed everything fresh per request, leaving it exposed to every variance class the freeze exists to close: per-query preflight tool picks, tools registering mid-session (e.g. sandbox coming online late), mid-session `SOUL.md` edits (whose own contract says "applies next session"), and DB-schema/secrets drift.

## Measurement — honest framing

A live 5-turn A/B on the dev build (same binary; the freeze only engages with `session_id`) using topically diverse, preflight-eligible queries showed **no regression and healthy disk-L2 reuse on both paths** (`+4–5 hits / 5 turns` each; TTFTs equivalent):

```
WITHOUT session_id: ttfts=[287,160,145,133,154] ms | disk hits +4
WITH    session_id: ttfts=[ 90,156,143,210,151] ms | disk hits +5
```

No TTFT *win* is claimed from this run, and the reason is instructive: the test machine has **no dynamic plugins installed**, so preflight had nothing to vary, and upstream's prompt-determinism work already keeps a plugin-less agent's recomposed prompt byte-stable. The variance classes this PR closes require dynamic tools / late registration / SOUL edits to manifest — states that exist in the field but not on a clean bench box. The change is therefore justified as **contract parity with the two already-frozen paths** (same store, same fingerprint rule, same semantics), with the A/B as no-regression evidence. Prompt/session regression suites (68 tests: SessionPreflightCache, PromptSectionOrdering, ToolResolution, PromptWhitespace) green.

## Risk & rollback

Engages only when `session_id` is present; explicit fingerprint invalidation preserves the mode-flip safety rule; stateless behavior untouched. Revert = one commit.

### DIFF
--- a/Packages/OsaurusCore/Networking/HTTPHandler.swift
+++ b/Packages/OsaurusCore/Networking/HTTPHandler.swift
@@ -2823,6 +2823,18 @@ final class HTTPHandler: ChannelInboundHandler, Sendable {
 
     // MARK: - Chat handlers
 
+    /// `SessionToolStateStore` key for the `/agents/{id}/run` path.
+    ///
+    /// `session_id` is a CLIENT-CHOSEN string; keying the store on it raw
+    /// would let two agents given the same value cross-contaminate each
+    /// other's frozen SOUL / manifest / tool snapshots. Namespacing by the
+    /// agent UUID makes collisions structurally impossible. The chat and
+    /// plugin surfaces own their keys structurally (chat window UUIDs,
+    /// per-plugin session scoping) and don't need this.
+    static func agentSessionStoreKey(agentId: UUID, sessionId: String) -> String {
+        "agent:\(agentId.uuidString):\(sessionId)"
+    }
+
     /// Enrich an agent-loop request with the agent's system prompt and memory context.
     ///
     /// Goes through `composeChatContext` and injects the rendered prompt +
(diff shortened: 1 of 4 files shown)


mimeding:
Adversarial-review fixes pushed (6c93d0a5): session store keys are now agent-namespaced (client-chosen session_ids could cross-contaminate frozen SOUL/manifest/tools across agents — the same leak existed for frozen memory prefixes and is fixed there too), and the post-compose delegation tool surface is frozen into the session snapshot so the byte-stability contract actually covers it. New AgentRunSessionFreezeTests (this path previously had zero direct tests).

PR #1922, opened 2026-07-07
Add live stream guardrails: degeneration abort + template-leak detection

mimeding:
## What

Live output guardrails in `GenerationEventMapper` — every local stream flows through it:

- **Degeneration detector** (incremental port of the `bench --gauntlet` detector from #1916, identical thresholds kept in sync by cross-referenced comments): any 3–12-token n-gram repeating ≥8× consecutively, or any ≥64-char single-character run — the documented `!!!!!` / `idea idea idea` failure shapes. Fed by both content and reasoning deltas via a bounded tail buffer (4096 chars ≈ 10× the worst-case 96-token detection window; trim realigns to whitespace so truncation can neither fabricate nor mask a match; per-delta cost O(4 KB)). On trigger: the stream finishes with a typed `StreamGuardrailError.degeneration(fragment:)`, one log line, the garbage delta never yielded, upstream generation cancelled via iterator drop. **Abort-only in v1** — retry-with-safe-settings needs caller-side design and is a noted follow-up.
- **Template-leak detector** (log-only, never mutates output): scans content deltas for `<|im_start|>`, `<|im_end|>`, `〈|EOS|〉`, `<tool_call>`, MiniMax markers, and the U+FFFE sentinel, with an 11-char cross-delta carry for split tokens; one log line per stream. `<think>` is deliberately excluded (some families legitimately emit it on the fallback template path) — divergence from the CLI list documented and test-guarded. A leaked marker means template-fallback mismatch: data for the capability ledger, not something to hide by filtering.
- Flags `ai.osaurus.guardrails.degenerationDetection` / `.templateLeakDetection`, both default **true** — detect-and-abort is strictly better than streaming unbounded garbage, and leak logging is free. Settings are injected into `map()` (default `.resolve()`) so production callers are unchanged and tests don't race process-global UserDefaults.

## Why

The gauntlet (#1916) catches degeneration at onboarding, but static verification can't cover every prompt — the historical TurboQuant repetition failures surfaced only on long thinking-mode generations in real use, and template-fallback mismatches (573-line fallback loader) recur with new bundles. This converts the worst user experience (an endless `!!!!!` stream) into an immediate, typed, logged failure, and makes template drift observable in the field.

## Measurement

- 50 tests in 4 suites green: all CLI boundary cases ported plus incremental-specific ones (trigger fragment split across 3-char deltas; `!!!!!` split across deltas; 64 KB clean stream with per-step buffer-bound assertion; loop after long clean prefix; latching; split leak token; reasoning-channel leak non-trigger; per-flag disable; **clean-stream byte-for-byte pass-through** asserted per-delta). Existing `GenerationEventMapperTests` untouched and green.
- Detector cost is bounded string work per delta, negligible next to decode.

## Risk & rollback

False-positive aborts are the risk; bounded by thresholds proven in the gauntlet's live runs (both installed model families pass clean), per-flag kill switches, and the typed error making any misfire diagnosable. Leak detection cannot alter output by construction. Revert = one commit.

### DIFF
--- a/Packages/OsaurusCore/Services/ModelRuntime/GenerationEventMapper.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime/GenerationEventMapper.swift
@@ -19,6 +19,11 @@
 //  stream with `.info(stopReason: .stop)`. Osaurus never inspects chunk
 //  text for stop-sequence matches.
 //
+//  The one exception to "purely a translation step" is the live guardrails
+//  (StreamDegenerationDetector.swift): text deltas are additionally fed to a
+//  per-stream degeneration detector (abort with a typed error) and template
+//  leak detector (log-only) — these observe, they never rewrite deltas.
+//
 
 import Foundation
 @preconcurrency import MLXLMCommon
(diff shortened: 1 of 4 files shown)


mimeding:
Adversarial-review hardening pushed (0ab8b42d): the char-run rule now exempts whitespace, uses per-class thresholds (64 letters/digits, 256 punctuation), and only triggers on runs that grow across ≥3 deltas — so markdown dividers, requested repetition ('print 100 !'), base64 runs, and table padding no longer abort streams; n-gram repeats raised 8→12. Legit-shape non-abort tests added alongside the true-positive pins; thresholds now intentionally diverge from the gauntlet's batch detector (documented).

PR #1926, opened 2026-07-07
Add cross-model bandwidth budgeter for multi-model big-RAM serving (flag-off)

mimeding:
## What

A **cross-model** decode admission budgeter for big-RAM machines running multiple models concurrently (`manualMultiModel`), dark-launched behind `ai.osaurus.perf.crossModelBudgeter` (default false):

- Admission: immediate when it's the only active model or the model is already decoding (same-model concurrency belongs to BatchEngine/#1919, not here); otherwise a new cross-model generation waits (FIFO, head-blocks so big models can't be starved, cancellation-safe removal) until the aggregate weight-streaming demand of distinct active models + incoming fits `bandwidth × 0.7` bytes/s — a coarse proportional guard, documented as such, not a scheduler.
- Wiring in `generateEventStream`: acquire right after `ModelLease.acquire`, release on both exit paths that release the lease (adapter-throw and producer-completion); one-shot idempotent lease token mirroring `HTTPInferenceAdmission.Token`.
- Gates (all must hold): flag on, `manualMultiModel` eviction, known bandwidth (local spec table pending #1923's measured record — cross-referenced), RAM ≥ 96 GiB. Strict-eviction single-model machines: provably zero change (tested).

**Stacked on #1861.**

## Why

Ultra/Max-class machines have RAM for several resident models but one memory bus; N concurrent cross-model decodes each run at ~bandwidth/N, which is strictly worse for interactive use than a short FIFO wait behind a fast turn. The same admission philosophy as `InferenceLoadCoordinator` (chat-vs-distillation), generalized across models and sized by the machine's actual bandwidth tier.

## Measurement

16/16 budgeter tests (admission math, FIFO no-skip, release-wakes-waiter, cancellation of a blocked head unblocks the next, same-model bypass, per-model-not-per-stream aggregation, idempotent double-release, all four gate legs + flag-default-false) and 200 tests across 8 adjacent runtime suites green. Live multi-model throughput measurement deferred to flag-flip evaluation — this ships dark, and the measurable claim is "no behavior change by default", which the tests pin.

## Risk & rollback

Flag-gated; head-of-line waiting bounded by generation lifetimes with cancellation-safe queue removal; no timeout by design (generations always finish or cancel). Revert = one commit.

### DIFF
--- a/Packages/OsaurusCore/Networking/HTTPHandler.swift
+++ b/Packages/OsaurusCore/Networking/HTTPHandler.swift
@@ -8857,6 +8857,7 @@ final class HTTPHandler: ChannelInboundHandler, Sendable {
             let obj: [String: Any] = [
                 "status": "healthy",
                 "timestamp": Date().ISO8601Format(),
+                "hardware": ChipProfile.current.healthJSONObject(),
                 "loaded": loaded,
                 "current_model": current,
                 "inflight": inflightObj,
--- a/Packages/OsaurusCore/Networking/ServerController.swift
+++ b/Packages/OsaurusCore/Networking/ServerController.swift
@@ -296,6 +296,14 @@ final class ServerController: ObservableObject {
     /// Saves the current configuration to disk
     func saveConfiguration() {
         ServerConfigurationStore.save(configuration)
+        // The runtime caches slices of this legacy configuration (the
+        // RuntimeConfig snapshot's genTopP fallback, and the model
+        // eviction policy consulted by the cross-model budgeter). Legacy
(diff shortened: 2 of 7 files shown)


mimeding:
Adversarial-review fixes pushed (a9d0c862): the same-model fast path is disallowed once the FIFO head has waited >60 s (the boundedness claim was false under continuous same-model traffic — now honest and aged); the per-request MainActor disk read is replaced by a cached eviction-policy read invalidated on config saves; holding logs go through os.Logger.

PR #1927, opened 2026-07-07
MoE-aware decode estimates + model-picker wiring

mimeding:
## What

Extends #1923's decode estimates with MoE awareness and wires them into the model picker:

- **MoE-aware estimator** (Core): active-fraction derived from the `A<k>B` token the metadata parser already understands (35B-A3B → 3/35 of total weights), applied as `bandwidth × 0.7 ÷ active-weights`. Dense models unchanged; config-declared MoE **without** a name-derivable fraction stays suppressed (a misleading dense floor is worse than nothing). Boundary-safe parsing (won't bite on "llama"/"qat"), memoized, documented as name-derived display-only.
- **`osaurus show`** un-suppressed for name-parseable MoE: `Estimated decode: ~145 tok/s on this Mac (measured 411 GB/s × 0.7 ÷ 2.0 GB active weights of 23.1 GB total)`.
- **Picker wiring landed fully**: `ModelPickerItem.estimatedDecodeTps` computed in the cache's existing off-main warm-up loop (measured-or-spec bandwidth, `ModelSizeCache` bytes, same MoE rule); the view change is one line appending `· ~145 tok/s` to the row's existing subtitle label — no layout changes.
- Ledger integration deliberately deferred: the gauntlet's measured decode currently lives in a prose evidence string, and parsing prose is how estimates become lies — follow-up is a structured decode field on the ledger record, after which measured beats estimated in the picker.

**Stacked on #1923.**

## Why

#1923 shipped honest estimates for dense models and suppressed MoE entirely; this closes the gap for the dominant modern bundles (MoE) using data osaurus already extracts from bundle names, so users see a defensible number before a 20 GB download.

## Measurement

Live validation on the dev machine: the name-derived estimate for Qwen3.6-35B-A3B (**~145 tok/s**) sits within 20% of the gauntlet's *measured* 121.3 tok/s @1K for the same bundle — right order, slightly optimistic (0.7 efficiency doesn't model MoE routing overhead; per-family tuning noted as follow-up). 52 Core tests across 5 suites + 13 CLI tests green, including an on-disk MoE-config fixture; builds and changed-line lint clean.

## Risk & rollback

Display-only numbers, clearly labeled; suppression rule prevents the known-misleading case; view diff is one line. Revert = one commit.


mimeding:
Closing without merge because this approach should not become capability truth. The branch derives MoE decode behavior from model names and inherited calibration code, while current main now has a canonical measured capability ledger. Model selection should use bundle configuration, active-parameter/runtime telemetry, or measured ledger evidence rather than marketing-name parsing. The valid calibration work is being retained in #1923.

PR #1928, opened 2026-07-07
Auto-run capability verification on model install (flag-off)

mimeding:
## What

Background capability verification on model install — the automatic complement to `bench --gauntlet` (#1916), dark-launched behind `ai.osaurus.verification.autoRunOnInstall` (default **false**; intended end-state is true after review):

- **`ModelVerificationScheduler`** (actor): on install completion, waits for true runtime idleness (chat coordinator idle AND no load or tracked generation in flight — deliberately stricter than chat-only, since HTTP/plugin streams are GPU work too), polling with 5s→60s backoff and giving up quietly after 30 min. Runs serialize; probes execute at `.utility`.
- **Probes v1 (deliberately minimal)**: a hidden 8-token greedy load probe (the #1903 micro-generation pattern, progress-UI suppressed) + a template-leak scan of its output (vendored 22-token mirror of #1922's detector, cross-referenced for later unification). After the probe, the model is unloaded **only if** the probe cold-loaded it and no leases exist — an already-resident model is untouched.
- **Ledger writes** to `~/.osaurus/config/model-ledger.json` (self-contained mirror of #1913's contract): per-probe verdict/evidence with `source:"autorun"` stamped per-entry so gauntlet and autorun writes interleave without clobbering provenance; foreign records and unknown fields preserved byte-for-byte; **`productionServing` is never written** — a background probe must never block (or bless) a model.
- **Trigger**: wired in `ModelDownloadService`'s completion block after finalization, completeness verification, and compatibility preflight — failed/paused/cancelled/preflight-blocked downloads can never fire it.

Self-contained on main (no dependency on the open ledger/guardrails PRs — they meet at the file contract).

## Why

The gauntlet turns "this bundle claims X" into "this bundle measured X on this machine" — but only when someone runs it. Wiring a minimal probe to install-time means every new model arrives with at least load + template-integrity facts in the ledger before its first real use, which is when template-fallback mismatches actually bite.

## Measurement

12/12 new tests (scheduler: runs-when-idle, skips-past-deadline, flag-off no-op, serialization; ledger: merge preservation incl. foreign gauntlet entries and unknown fields; leak-scan token list) + 79 neighboring suite tests green. No real model loads in verification — probes exercised through injection seams; the live behavior composes two already-proven mechanisms (#1903's micro-generation, #1916's ledger writes). Builds, `swift-format --strict`, and changed-line swiftlint all clean.

## Risk & rollback

Default-off dark launch; idle-gated with a give-up deadline; residency-respecting unload; conservative ledger semantics. Open review questions listed in-code: ledger field casing vs #1913, whether skipped-busy runs deserve a ledger record, Settings toggle placement. Revert = one commit.

### DIFF
--- a/Packages/OsaurusCore/Services/ModelDownloadService.swift
+++ b/Packages/OsaurusCore/Services/ModelDownloadService.swift
@@ -509,6 +509,14 @@ final class ModelDownloadService: ObservableObject {
                         )
                         ModelManager.invalidateLocalModelsCache()
                         NotificationCenter.default.post(name: .localModelsChanged, object: nil)
+                        // Background capability autorun (ledger probes). No-op
+                        // unless `ai.osaurus.verification.autoRunOnInstall` is
+                        // enabled; the scheduler waits for a fully idle runtime
+                        // before touching the GPU.
+                        ModelVerificationScheduler.shared.modelInstalled(
+                            name: model.name,
+                            localDirectory: model.localDirectory
+                        )
                     }
                 }
             } catch let pauseInfo as DirectDownloader.PauseInfo {
--- a/Packages/OsaurusCore/Services/ModelRuntime.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime.swift
@@ -205,6 +205,17 @@ public actor ModelRuntime {
         return modelCache[name] != nil
     }
 
+    /// True while any container load or tracked generation wrapper task is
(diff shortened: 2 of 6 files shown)


mimeding:
Closing without merge as superseded by the canonical capability-ledger direction. This branch writes a nested ledger schema that is incompatible with the flat store now on main and duplicates probe/detector ownership without retained live-model proof. A future install-time verification flow should reuse the measured writer and runtime guardrail contracts after they are stable.

PR #1945, opened 2026-07-09
Add secure chat document intake and recovery

mimeding:
## Summary
- adds a guided chat document intake flow with preview, progress, attach, cancel, and explicit unsupported-format states
- preserves structured document provenance across chat persistence and encrypted image/document spillover
- verifies attachment integrity before model input, historical replay, and ZIP export
- exports extracted document text with truthful filenames and supports safe document conversion without overwriting unless the user explicitly allows it
- restores clipboard retry behavior after intake failures or cancellation

## Security and reliability
- captures selected files once with no-follow bounded descriptor reads, then hashes and parses the same private immutable copy
- rejects symlinks, special files, oversized inputs, source races, provenance mismatches, and tampered persisted blobs
- uses private temporary permissions, atomic publication, bounded reads, path-free diagnostics, and cancellation generation fences
- blocks send explicitly when a provenance-bearing attachment cannot be verified
- preserves memory and screen context when corrupt media fails to hydrate

## Validation
- 83 focused document, chat, export, spillover, and security tests passed
- 48 document-intake, attachment-hardening, and spillover tests passed after the CI repair
- unsigned Release CLI and app builds passed
- localization validation now covers the page action, page-number interpolation, integrity status, security inspection, and source identity labels
- SwiftLint and `git diff --check` passed

## Review state
This PR remains draft only until the native UI walkthrough is completed. All required GitHub checks are green and the merge state is clean.

## Known limits
- scanned PDFs use the image-page fallback when structured extraction is unavailable
- unsupported emitters are reported explicitly rather than silently producing a different format


RaajeevChandran:
- describe() has cosmetic redundancy — it computes name but the default (video) branch re-reads att.redactedFilename ?? "attachment" instead of using name
  - exportFilename forces documentRefs to .txt (since the recovered blob is extracted text, so budget.xlsx → budget.txt). Intentional and tested, just worth being aware the exported extension won't match the original document type

PR #1958, opened 2026-07-09
Size model fit against the GPU working set, not physical RAM

jjang-ai:
## Why

Two user reports that look opposite but share one root cause: model fit was measured against **physical RAM** instead of the **GPU working set**.

**Report 1 — a model that "fits" but crawls.** A 48 GB M5 Pro installed Ornith-1.0-35B-MXFP8. Osaurus said "41 GB required, 63% of memory loaded, close other applications", then generated about one character every ten seconds with nothing else running.

That's not RAM pressure. MLX holds weights in Metal buffers, and its allocator only refuses a request once it passes `min(1.5 × recommendedMaxWorkingSetSize, 0.95 × RAM)` (`mlx/backend/metal/allocator.cpp:56`); `wired_limit_` defaults to `0`, so nothing is pinned in the residency set. Anything between the recommended working set and that hard limit **allocates successfully** and is then paged by macOS on every decode step. The model loads without a single error and is unusable.

The bundle is 34.17 GiB on disk → `estimatedMemoryGB` = ×1.25 = **42.7 GB**. Against physical RAM that's 42.7/48 = 0.890 → `.tight`, so we offered it. Against the ~36 GB a 48 GB Mac budgets to the GPU it's **1.19** → it can never run well.

**Report 2 — the RAM popup is too aggressive.** An M1 Max with 64 GB running Qwen3.6-35B-A3B-MXFP4 (~25.8 GB to load) got "This model needs ~25.8 GB to load and memory is at 77% of 64 GB. Close other apps for best performance." on every launch, with plenty of RAM free.

That model needs 25.8 GB against a 48 GB budget — a 0.54 ratio, comfortable by any measure. The banner fired only because of `lowAvailable`: `buildRAMFeasibility` sets `verdict = .tight` when free pages are short, and `loadPressureSeverity` warned on `verdict != .ok`. But on macOS free pages are *always* scarce — the compressor and file cache hand memory back on demand.

And "close other apps" is actively wrong for report 1: the GPU working set is a fixed fraction of installed RAM, so closing apps cannot make room.

## What changed

**New `GPUMemoryBudget`** — how much unified memory a model may actually occupy:

```
budget(RAM) = RAM × (RAM ≤ 36.5 ? 2/3 : 0.75)
budget      = min(budget(RAM), metalAdvertised)   // host only; only ever LOWERS
```

Metal's `recommendedMaxWorkingSetSize` is consulted only for the machine we're running on, and only to *lower* the budget — a user who pinned `iogpu.wired_limit_mb` below the default split gets the tighter number, but nobody gets a more optimistic one. A hypothetical RAM size (catalog math, tests) always resolves to the pure default, so a verdict never depends on the machine that computed it. This matters: an M5 Max on macOS 26 advertises **84%** of RAM (107.52 GB of 128 GB), not the classic 75%, so the fraction is not a constant worth hardcoding as truth.

**`MLXModel.compatibility`** now divides by the budget, not RAM:

| ratio | verdict |
|---|---|
| ≤ 0.85 | `.compatible` |
| ≤ 1.10 | `.tight` — runs, close other apps |
| > 1.10 | `.tooLarge` — macOS would page it |

The 1.10 band absorbs the conservative split (the recommended working set is a recommendation, not a cliff). `.tooLarge` stays advisory: it badges, deprioritises in sort order, excludes the model from `recommendedLocalPick`, and disables the onboarding radio row — it **never blocks a download**.

**`RAMFeasibility`** gains `gpuBudgetBytes` and `exceedsGPUBudget`, and `loadPressureSeverity` now warns on `exceedsGPUBudget` or `projected > softLimit` — no longer on `verdict != .ok`, which folded in the noisy `lowAvailable` signal. `verdict` still records that pressure for `/health` and logs, and the load path stays advisory exactly as before.

**Banner copy** tells the truth when over budget instead of suggesting something that cannot help:

> This model needs ~42.7 GB, but this Mac can only keep ~36 GB on the GPU. macOS will page the weights and generation will be extremely slow — pick a smaller model or a lower-precision build.

## Verdicts against real bundles

Measured on disk, `need = disk × 1.25`:

| bundle | disk | need | 16 | 24 | 32 | 48 | 64 | 96 | 128 |
|---|---|---|---|---|---|---|---|---|---|
| Ornith-1.0-35B-MXFP8 | 34.2 G | 42.7 G | too | too | too | **too** | tight | ok | ok |
| Ornith-1.0-9B-MXFP8 | 9.5 G | 11.8 G | too | ok | ok | ok | ok | ok | ok |
| Qwen3.6-35B-A3B-MXFP4-MTP | 18.5 G | 23.1 G | too | too | tight | ok | **ok** | ok | ok |
| gemma-3n-E2B | 7.6 G | 9.5 G | tight | ok | ok | ok | ok | ok | ok |

Row 1 column 48 is report 1 (now correctly refused as a recommendation). Row 3 column 64 is report 2 (comfortable — the new rule never touches it).

## Tests

- New `GPUMemoryBudgetTests` (9): the budget split, that Metal can only tighten and never loosen, that hypothetical RAM sizes are host-independent, and the 35B-MXFP8-on-48 GB regression.
- New in `ModelRuntimeRAMFeasibilityTests`: `comfortableModelOnBusyMacStaysQuiet()` pins report 2's exact byte counts; `exceedingGPUBudgetWarnsOnIdleMac()` pins report 1's.
- `softThresholdCrossingWarns` updated: the soft threshold (0.80 of RAM) sits *above* the GPU budget (0.75), so a single model at the soft limit is already past what Metal keeps resident. The physical soft/hard thresholds now only bind for multi-model residency, where `projected` includes `resident + inflightOther`.
- `lowAvailableWarnsWithoutBlocking` → `lowAvailableIsAdvisoryOnly`: encodes the fixed behaviour from report 2.
- Fixtures re-based where the old ones were only "comfortable" under the RAM ratio: `ModelFilterPerformanceTests` (`mid` 11 → 8 GB, since 11 GB is genuinely `.tooLarge` on a 16 GB Mac now) and `ConfigureAIStateResourceTests` (4/8 GB → 2/6 GB).

## Verified live

Built the dev app from this branch and ran against it on a 128 GB M5 Max (`/health` now reports the budget):

```
model                            ornith-1.0-9b-mxfp8
verdict                          ok
incoming_load_footprint_bytes        9.45 GB
kv_headroom_bytes                    1.89 GB
required_available_bytes            11.34 GB
gpu_budget_bytes                    96.00 GB    <- min(0.75 × 128, 107.52 advertised)
exceeds_gpu_budget               False
```

`required_available` of 11.34 GB against the catalog's ×1.25 estimate of 11.8 GB for the same bundle is an independent check that the heuristic tracks the runtime's config-derived KV sizing.

Cross-matrix over the architecture families (multiturn recall, tool call, reasoning toggle on/off, control-marker leak scan) — all pass, no leakage:

| model | arch | multiturn | tool call | think on/off | leaks |
|---|---|---|---|---|---|
| ornith-1.0-9b-mxfp8 | hybrid GDN | pass | `{"city":"Paris"}` | pass (rc 1093 ch) | none |
| gemma-4-e2b-it-4bit | dense sliding | pass | `{"city":"Paris"}` | pass (rc 881 ch) | none |
| qwen3.6-35b-a3b-mxfp8-crack-mtp | MoE + MTP | pass | `{"city":"Paris"}` | pass (rc 1218 ch) | none |
| lfm2.5-8b-a1b-mxfp8 | hybrid SSM | pass¹ | `{"city":"Paris"}` | pass (rc 895 ch) | none |

¹ needs `max_tokens ≥ 400`: at 80 it spends the whole budget on reasoning and returns `finish_reason=length` with empty content. Pre-existing, unrelated to this change.

The over-budget banner itself can't be exercised on a 128 GB host (no catalog model exceeds 96 GB); it is pinned by `exceedingGPUBudgetWarnsOnIdleMac()` and `gpuBudgetForFortyEightGigMac()` instead.

## Test note

`swift test` on `origin/main` already fails ~7 tests in a full run (tool-registry counts, gemma4 tokenizer surface, path resolver, chat-store flush, plus order-dependent flakes) and hangs partway. That is pre-existing and unrelated: the suites covering this change — `GPUMemoryBudget`, `ModelRuntimeRAMFeasibility`, `ChatWarmup`, `ModelFilterPerformance`, `ConfigureAIState`, `ModelManagerSuggested`, `FamilyGrouping`, `RuntimePolicySource`, `ImageGenerationBridge` — pass **218 tests / 21 suites**.

### DIFF
--- a/Packages/OsaurusCore/Managers/Model/ModelManager.swift
+++ b/Packages/OsaurusCore/Managers/Model/ModelManager.swift
@@ -116,15 +116,16 @@ final class ModelManager: NSObject, ObservableObject {
         /// hardware info and we let everything through until we know.
         enum PerformanceFilter: String, CaseIterable, Identifiable {
             /// Only include models whose `compatibility` is `.compatible`
-            /// (memory usage below the 75 % ratio threshold).
+            /// (working set at or below 85 % of the GPU memory budget).
             case runsWell = "Runs Well"
             /// Only include models whose `compatibility` is `.tight`
-            /// (memory usage between 75 % and 95 % of total RAM)
+            /// (working set between 85 % and 110 % of the GPU memory budget).
             case tightFit = "Tight Fit"
             /// Exclude models whose advisory `compatibility` is `.tooLarge`
-            /// (memory usage above the 95 % ratio threshold). This filter is
-            /// user-selected catalog triage only; runtime load/download does
-            /// not block RAM pressure from this estimate.
+            /// (working set above 110 % of the GPU memory budget — macOS would
+            /// page the weights). This filter is user-selected catalog triage
+            /// only; runtime load/download does not block RAM pressure from
+            /// this estimate.
             case hideTooLarge = "Hide Too Large"
 
             var id: String { rawValue }
--- a/Packages/OsaurusCore/Models/Configuration/GPUMemoryBudget.swift
+++ b/Packages/OsaurusCore/Models/Configuration/GPUMemoryBudget.swift
@@ -1,0 +1,78 @@
+//
+//  GPUMemoryBudget.swift
+//  OsaurusCore
+//
+//  The unified-memory ceiling a model's working set has to respect.
+//
+
+import Foundation
+
+#if canImport(Metal)
+    import Metal
+#endif
+
+/// How much unified memory a model may actually occupy before macOS starts
+/// paging it.
+///
+/// MLX holds weights in Metal buffers, and its allocator only refuses a
+/// request once it passes `min(1.5 × recommendedMaxWorkingSetSize, 0.95 × RAM)`.
+/// Everything between the recommended working set and that hard limit
+/// *allocates successfully* and is then paged in and out by the OS on every
+/// decode step. So a model sized against raw RAM can pass the check, load
+/// without a single error, and then emit roughly one character every ten
+/// seconds. Fit has to be judged against the GPU working set, not against
+/// `physicalMemory`.
+enum GPUMemoryBudget {
+
+    private static let bytesPerGB: Double = 1024 * 1024 * 1024
+
+    /// Apple's default split between the GPU working set and everything else.
+    /// Machines at or below 36 GB hold back proportionally more for the OS.
+    ///
+    /// This is deliberately the long-standing documented split rather than the
+    /// (larger) figure recent macOS releases advertise — an M5 Max on macOS 26
+    /// reports 84% of RAM. Pinning the catalog verdict to the conservative
+    /// number keeps it stable across OS releases and never optimistic, and the
+    /// `.tight` band below absorbs the difference.
+    static func defaultBudgetGB(physicalMemoryGB: Double) -> Double {
+        guard physicalMemoryGB > 0 else { return 0 }
+        return physicalMemoryGB * (physicalMemoryGB <= 36.5 ? (2.0 / 3.0) : 0.75)
+    }
+
+    /// Physical memory of the machine we're running on, in GB.
+    static let hostPhysicalMemoryGB: Double =
+        Double(ProcessInfo.processInfo.physicalMemory) / bytesPerGB
+
+    /// What Metal advertises as this machine's working-set budget, when that
+    /// is usable. `nil` on a paravirtual device, when the value exceeds
+    /// installed RAM, or where Metal is unavailable.
+    static let hostAdvertisedBudgetGB: Double? = {
+        #if canImport(Metal)
+            guard let device = MTLCreateSystemDefaultDevice() else { return nil }
+            let advertised = Double(device.recommendedMaxWorkingSetSize) / bytesPerGB
+            guard advertised > 0, advertised <= hostPhysicalMemoryGB else { return nil }
+            return advertised
+        #else
+            return nil
+        #endif
+    }()
+
+    /// Working-set budget for a Mac with `physicalMemoryGB` of unified memory.
+    ///
+    /// Metal is consulted only for the machine we're actually running on, and
+    /// only ever to *lower* the budget — a user who has pinned
+    /// `iogpu.wired_limit_mb` below the default split gets the tighter number,
+    /// but nobody gets a more optimistic one than `defaultBudgetGB`. Callers
+    /// passing a hypothetical RAM size (the catalog reasoning about another
+    /// machine, or a test) always land on the pure default, so the verdict
+    /// never depends on the host it was computed on.
+    static func budgetGB(physicalMemoryGB: Double) -> Double {
+        let base = defaultBudgetGB(physicalMemoryGB: physicalMemoryGB)
+        guard
+            base > 0,
+            abs(physicalMemoryGB - hostPhysicalMemoryGB) < 1.0,
+            let advertised = hostAdvertisedBudgetGB
+        else { return base }
+        return min(base, advertised)
+    }
+}
--- a/Packages/OsaurusCore/Models/Configuration/MLXModel.swift
+++ b/Packages/OsaurusCore/Models/Configuration/MLXModel.swift
@@ -565,12 +565,34 @@ struct MLXModel: Identifiable, Codable {
             : String(format: "~%.1f GB", gb)
     }
 
-    /// Assess whether this model can run on the given hardware.
+    /// Comfortably inside the GPU working set, with room for a long-context KV
+    /// cache to grow.
+    private static let comfortableBudgetRatio: Double = 0.85
+    /// The most a working set may overshoot the budget and still be offered.
+    /// `GPUMemoryBudget.defaultBudgetGB` is the conservative split rather than
+    /// the larger one modern macOS advertises, and the recommended working set
+    /// is a recommendation rather than a cliff, so a small overshoot is
+    /// survivable — it just leaves nothing for other apps. Past this the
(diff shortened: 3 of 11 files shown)


mimeding:
Ran the branch's fit logic against our measured rows and traced the verdict through its consumers. Data first, then two questions, then an offer.

## Data consistent with the mechanism GPUMemoryBudget.swift documents

Our 2026-07-07 KV-codec sweep (M5 Max 128 GB, macOS 26, greedy, fixed 128-token outputs) has a context-scaling row that fits the paging description in the file doc:

| MiniMax-M2.7 fp16 (63 GB weights, full-KV) | 8K | 32K | 64K |
|---|---|---|---|
| decode tok/s | ~92 | ~61 | ~9.6 |
| TTFT (uncached) | — | 60.4 / 63.8 s | 161.5 / 182.4 s |

The 64K decode collapse is an order of magnitude, not the smooth 1/(weights+KV) bandwidth decay the 8K→32K step follows, and the TurboQuant arm at 32K showed the run-to-run TTFT spread you flagged in vmlx#124 (62.7 vs 114.0 s on identical requests). 63 GB of weights plus fp16 KV at 64K plausibly crosses this machine's working set, so your paging read of F4 is consistent with what we logged. Two honest caveats: n=2 per cell (below your ≥5 bar), and we logged no `vm_stat` pagein/compressor counters, so bandwidth decay vs paging is not yet discriminated in our data. Offer to fix that below.

For the host-clamp premise: confirmed on this machine — `iogpu.wired_limit_mb=0` (kernel default), `hw.memsize=137438953472`, and macOS 26 advertising ~84% of RAM as the working set (~107.5 GiB), matching the note in GPUMemoryBudget.swift.

## Two questions

**1. Is `requiredAvailableBytes` the intended comparand for `exceedsGPUBudget`?** (ModelRuntime.swift:970, with `requiredAvailable = incomingLoadFootprintBytes + kvHeadroom` at :1212.) `kvHeadroom` sizes KV at `max_position_embeddings` plus 25% slack. Concrete case on a 128 GB machine: Llama-3.3-70B-8bit is 75 GB of weights, and its config-derived headroom is 80×2×8×128×131072×2 B ×1.25 ≈ 50 GiB, so `requiredAvailable` ≈ 125 GB > the 96 GiB budget → the banner asserts "macOS will page the weights and generation will be extremely slow." But KV grows with use; actual resident at ≤8K context is ~78 GB, inside even the conservative budget, and we've run this model here with steady bandwidth-bound decode and no collapse. Weights-only against the budget would make the message true whenever it fires; KV growth seems like a different (long-context) advisory. Related: since the budget (0.75) sits below both block triggers on >36.5 GB machines, the `exceedsGPUBudget` branch in FloatingInputCard.swift:3484–3494 shadows the blocked copy, so a `.block`-severity load now shows a banner that never says sending is paused.

**2. On the host, should Metal's advertisement be trusted upward too?** `budgetGB` only ever lowers (GPUMemoryBudget.swift:76), but the mechanism in the file doc pins paging onset to `recommendedMaxWorkingSetSize` — which on this M5 Max is ~107.5 GiB, not 96. By the doc's own model, nothing in the 96–107.5 GiB band pages, yet `exceedsGPUBudget` (ratio 1.0) warns there. The catalog's 1.10 tight band absorbs most of it (0.825 ≈ the 0.84 advertisement — nice), so the tension is specifically the runtime banner. Same shape for users who raise `iogpu.wired_limit_mb` deliberately: the banner tells them no configuration can help, which is the one case where it's false. (Smaller note: `hostAdvertisedBudgetGB` is a cached `static let`, so the pinned-below case in the doc only holds for the value at launch.)

One interaction to flag rather than ask: `requiredAvailableBytes` excludes `resident`/`inflightOther`, so under flexible multi-model residency two models each inside the budget can jointly exceed the working set with no warn until `projected > 0.80×RAM` — a silent 0.75–0.80 band. Our #1902 (longer idle-residency windows on big RAM) widens exposure to exactly that case, and #1911's larger MLX buffer cache sits in the same working set uncounted. Happy to reconcile from our side whichever ledger you prefer.

## Offer: live validation on 128 GB / 75 GB weights

This machine can bracket both candidate ceilings (96 GiB default vs ~107.5 GiB advertised) with the installed Llama-3.3-70B-8bit — dense KVCacheSimple, so KV is a clean throttle at exactly 320 KiB/token (1 GiB per ~3,277 tokens). Protocol:

- Context ladder 8K → 120K with ≥2 rungs whose *measured* resident (`footprint -p`, logged before/during/after decode) lands strictly between 96 and 107.5 GiB, ≥1 rung past 107.5.
- ≥5 repeats per rung (cold load per rung, plus a fresh-cold repeat block for order effects), reporting spread.
- Per repeat: `osaurus bench` TTFT/prefill/decode; `vm_stat` pagein/pageout/compressor deltas across the decode window; `vm.swapusage`; and the `/health` `ram_feasibility` snapshot (`gpu_budget_bytes`, `exceeds_gpu_budget`) so each rung records what this PR would have said.
- Read-out: cliff + pageins + TTFT spread starting at ~96 GiB validates `defaultBudgetGB`; smooth tight-spread decay through 96–107.5 with the cliff only past ~107.5 says the advertised value is the host's real ceiling and the band is being over-warned. Optional last run with `iogpu.wired_limit_mb` raised to test whether the cliff moves.

Say the word and we'll run it and attach the raw table.

PR #1962, opened 2026-07-10
Surface engine-path attribution and predicted cache topology through diagnostics

mimeding:
## What

Surface **engine-path attribution** — which vmlx code path a request was submitted on and why native MTP did or did not engage — plus **predicted per-layer cache topology**, through the existing diagnostics surfaces:

1. **Per-model** (`/admin/generation-settings`, `/admin/cache-stats`): `EffectiveGenerationSettings` gains `engine_path` (`solo`/`batched`), `native_mtp_requested`, `native_mtp_engaged`, `native_mtp_fallback_reason`. The load-time MTP warmup (#1903) now records a distinct `stage: "load_warmup"` so its row can't be mistaken for user traffic.
2. **Per-request** (OpenAI-compatible `usage`, streaming and non-streaming, tool-call and plugin paths): additive optional keys `engine` (`solo_mtp` | `solo_ar` | `solo_diffusion` | `batched`) and `native_mtp_fallback` (`cold_warmup` | `explicit_sampling` | `tiny_prompt` | `engine_sampling_gate` | `engine_iterator_gate`), carried on the existing stats sentinel's forward-compatible flag list.
3. **`/admin/cache-stats`**: `predicted_cache_topology` — per resident model, the per-layer cache classes from `model.newCache(parameters:)` type inspection (`CacheList` rendered one level deep), as **two rows**: default parameters, and `long_prompt(maxKVSize=N)` using the live coordinator's cap, with `applies_when` thresholds matching the engine's `resolveKVPolicy` rule. Rows are labeled `source: newCache` — predicted construction, not the live in-flight array (which is engine-internal).

Attribution is **submitted-path attribution**: built once in `MLXBatchAdapter.generate` after the final `GenerateParameters` exist, using the engine's own public gates — `mlxParams.canUseNativeMTP(for: input)` for eligibility, `model is any BlockDiffusionModel` (the exact `BatchEngine.generate` dispatch predicate, evaluated on the loaded instance) for diffusion routing. Known residuals where the engine's internal choice can still differ (post-disconnect drain window; `nativeMTPAvailable` iterator veto) are named in the doc comments rather than papered over.

Two defects found during verification are fixed in-branch, both pre-existing on main:
- Non-finite `tokens_per_second` reached the wire as invalid JSON (`"tokens_per_second":nan`) on cancelled zero-token streams; now normalized at the sentinel encode and omitted from `usage` JSON.
- Admin snapshots retained model-holder references across container awaits, so an unload racing the endpoint deferred the multi-GB weight release outside the teardown gate (the documented `Fence::wait`/`MetalAllocator::free` SIGSEGV class); releases now route through a gated, synchronized section with holder-identity detection.

## Why

Boundary attribution is the engine maintainer's stated top cost: osaurus computes which path it submits on (solo lease, effective draft strategy, fallback reason) but only logged it, so "did MTP actually engage, and if not why" was invisible to clients and to anyone debugging across the osaurus↔vmlx boundary. This is the osaurus-side counterpart of vmlx-swift#120 §7 ("surface the existing [NativeMTP] counters through osaurus's diagnostics endpoint… small plumbing task in your own repo"): the request-level `engine`/`mtp_off` attribution answers *whether and why* speculation was submitted, and the topology block answers *what cache classes the engine will build* — observed from the engine's own public API, not inferred from config.

A concrete example of what this makes visible: MTP-tuned bundles ship `generation_config.json` sampling defaults (e.g. `top_p: 0.95`), so a plain `temperature: 0` request silently loses native MTP through the engine's greedy-eligibility gate. Today that is undetectable without stderr access; with this PR the usage block says `"engine":"solo_ar","native_mtp_fallback":"engine_sampling_gate"`.

## Measurement

Live on M5 Max 128 GB, isolated instance, `qwen3.6-35b-a3b-mxfp4-mtp` (hybrid, `native_mtp:d3`) — all observed values, raw transcripts retained:

- **Load**: `/admin/generation-settings` shows `{"stage":"load_warmup","engine_path":"solo","native_mtp_requested":true,"native_mtp_engaged":false,"native_mtp_fallback_reason":"cold_warmup"}` during the #1903 warmup window.
- **Explicit greedy** (`top_p:1, top_k:0, min_p:0`): usage `{"completion_tokens":79,"tokens_per_second":68.6,"engine":"solo_mtp"}`, agreeing with the engine's `[NativeMTP] … outputTokens=79` stderr line for the same generation.
- **Greedy with bundle defaults** (`top_p:0.95` merges in): `{"engine":"solo_ar","native_mtp_fallback":"engine_sampling_gate"}` — the silent-strip case above, observed.
- **Sampled** (`temperature:0.7`): `{"engine":"solo_ar","native_mtp_fallback":"explicit_sampling"}`; no `[NativeMTP]` line emitted.
- **`max_tokens:1` greedy, ≥24-token prompt**: `finish_reason:"cancelled"`, `{"completion_tokens":0,"engine":"solo_mtp","native_mtp_fallback":"engine_iterator_gate"}`, no invalid JSON (with a shorter prompt, osaurus's own `tiny_prompt` gate correctly fires first).
- **Topology**: `{"cache_entries":40,"classes":{"MambaCache":30,"KVCacheSimple":10},"layout":"3xMambaCache,1xKVCacheSimple,…"}` plus the `long_prompt(maxKVSize=65536)` row flipping full-attention slots to `RotatingKVCache`, `applies_when: prompt_tokens>131072` (= cap × 2.0 multiplier, matching the live coordinator config).

Zero behavior change to generation: no parameter, gate, or submit-path semantics are touched (the only wire-visible delta beyond the new optional keys is the invalid-JSON fix above). The attribution taxonomy was verified cell-by-cell against `BatchEngine.generate`'s branch structure at the pinned revision, including the demote-vs-cancel asymmetry (`engine_sampling_gate` → plain AR runs; `engine_iterator_gate` → cancelled stream). Targeted suites: 290 tests across 8 suites, green; mid-tier wire behavior pinned byte-identical by legacy-decode tests.

## Risk & rollback

Additive only: new optional struct fields with defaulted params, optional stats-sentinel flags (the flag list is forward-compatible by design; unknown-flag preservation is test-pinned), optional `usage` keys omitted when absent, new namespaced `/admin` JSON blocks. Both in-branch fixes tighten existing failure paths (invalid JSON → omitted key; ungated release → gated). Rollback is a clean revert of one commit.


mimeding:
This PR is being returned to draft because it is conflict-blocked against current main and its engine-attribution contract must be revalidated against the currently pinned runtime. Please do not review or merge it in its current state. The next step is either a current-main refresh with updated live attribution and cache evidence, or closure as superseded if the current diagnostics already cover the same contract.

PR #1986, opened 2026-07-12
Background work must never evict the model the user is using (+ vmlx repin)

jjang-ai:
## Background work could throw away the user's model

The MLX runtime is strictly single-model: loading model B **evicts** resident model A, and a new load **cancels** one already in flight. #1984 taught warm-ups and the greeting pool the right rule — fill an empty slot, never disturb an occupied one — but three callers still went straight through `CoreModelService` into the runtime with no residency check at all:

- **memory distillation**, which runs unprompted on ingest and at launch
- **generative greetings**
- **voice-transcript cleanup** — whose local fallback picked `installed.first`, an **arbitrary model chosen by install order**

So the app could evict the model you were mid-conversation with, or cancel the load you were watching a spinner for, in order to summarise your history or tidy a voice transcript.

### The fix

`CoreModelService.generate` takes a `CoreModelIntent` (`.interactive` / `.background`). A `.background` call that resolves to the local MLX route declines when a load is in flight or another model is resident. Remote and Foundation routes don't touch GPU residency, so only the local route is gated.

**Declining is not a failure.** It deliberately does *not* count toward the circuit breaker — otherwise a long chat, where a model is legitimately resident the whole time, would trip the breaker and lock the user out of their own *interactive* calls. Distillation logs it as `model_in_use` and leaves the signals pending for a later pass; voice cleanup keeps the raw transcript; the greeting pool already breaks its refill cycle.

Voice cleanup's fallback now prefers a model that is **already resident**, accepts any installed model only when the runtime is **idle**, and otherwise declines. A tidier transcript is not worth the user's model.

## vmlx repin → `e487f3f` (vmlx-swift#139)

- **The prefix-cache store could kill the machine.** It materialised the KV cache several times over at the memory high-water mark with nothing measuring first — reported as a kernel panic on a 128 GB Mac at 64K context. It now measures before it copies and skips the store when the copies won't fit. A skipped store costs one slower request; an unchecked store costs the machine. Guarded in all four store paths (solo, batched, diffusion, MTP).
- **Hunyuan v3 packs with a bare marker lost their entire answer.** HY3's template writes markers through a variable; open-source packs use `:opensource`, preview conversions leave it empty. Our parser knew only the suffixed spelling *and starts inside reasoning*, so a bare-marker pack had its whole reply — tool calls included — routed into the thinking block, leaving the visible answer **empty**. Both spellings are now accepted; every other model family is untouched.
- **The "thinking didn't close" chip** no longer fires on a clean parse.

## Verification

- `CoreModelService` suite green (7/7); OsaurusCore builds clean.
- vmlx side: verified against a clean worktree of vmlx `main` — the branch introduces **zero new test failures**, fixes two pre-existing ones, and removes a hard crash in the focused suite.
- Codex reviewed both diffs adversarially and confirmed: the decline cannot wedge any caller, the breaker is correctly bypassed, resident-name shapes match between `MLXService.getAvailableModels()` and the runtime's cache keys, and actor isolation on the new `residentModelNames()` is sound.
- Live in-app multiturn verification across HY3 / Gemma-4 / Qwen-MoE / a dense model is the final gate and runs against this branch's dev build.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "31b165cb0fa3ca49a8495971f8d906157906aeb6"
+        "revision" : "e487f3fff06883ab71341a3c550246c48a8ef222"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "e487f3fff06883ab71341a3c550246c48a8ef222"
+        "revision" : "31b165cb0fa3ca49a8495971f8d906157906aeb6"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "65e3237144419ba6e862058b4a18440e70ba90bc4e73db274e72fab957c62145",
+  "originHash" : "4472c2a6a993478f2106174982f1cf4e2912105336c34c2c9e710d9516aae144",
   "pins" : [
     {
       "identity" : "aachartkit-swift",
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "31b165cb0fa3ca49a8495971f8d906157906aeb6"
+        "revision" : "e487f3fff06883ab71341a3c550246c48a8ef222"
       }
     },
     {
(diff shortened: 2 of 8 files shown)


jjang-ai:
## Live visual gate — dev-built PR app, driven through the UI (codex computer-use)

No curl, no API, no unit tests were used as proof. Every result below is what rendered on screen.

**HY3 (`hy3-jang_2k`, 94 GB) — PASS, all 5 turns.**
- Loads and reports `Model warm — ready for a fast first response`.
- Reasoning renders as a proper collapsible block (`Thought for 1m 10s`, `218 chars`) — **no `:opensource` marker leakage, no empty reply, no split error**. This is the exact failure the parser fix targets.
- Tool card fired once (`Get current time · 3.3s`) and the result was used verbatim in the answer.
- Turn 4 recalled Paris + the three landmarks + the tool result across the tool turn.
- Reasoning toggle High → Off: no thinking block, correct answer.
- Prefix-cache reuse visible: TTFT 13.28s → 2.56s, 4.7 → 17.3 tok/s between turn 1 and 2.

**Gemma-4 12B — PASS, all 5 turns.** Unaffected by the parser change, as designed: tool card fired, harmony reasoning intact, web-search card on the thinking toggle, no protocol leakage.

**Qwen3.6-35B CRACK MTP — Turn 2 derails with tools ON, and it is PRE-EXISTING.**
Rather than assume, I rebuilt this same branch against the **old vmlx pin** and ran the identical script four times. The baseline breaks Turn 2 in **2 of 4** runs too (artifact detours, a failed `share_artifact`, a mid-word truncation). With tools **off** the model answers correctly at 87 tok/s. The engine parses faithfully — when the model invented tools `scope` and `function`, those names reached the app verbatim and were correctly rejected with `tool_not_found`. Tracked separately as an agent-loop/model issue; not a regression from this PR.

**Qwen3-0.6B dense** — weak-model errors (duplicated "Eiffel Tower"). Not an engine defect.

Also filed from this run: short outputs render nonsense tok/s (`117597.3 tok/s` on a 7-token reply), and image gen/edit has **no RAM feasibility gate at all** — both separate from this PR.

PR #1995, opened 2026-07-12
Headless runs must not claim to be the user; the stats card must not invent a tok/s

jjang-ai:
Stacked on #1994 (needs `ModelLoadIntent` from it). Two independent fixes; both live-gated on a dev build via computer-use.

---

## 1. Headless runs claimed to be the user

A cron schedule firing, a file watcher tripping, an agent waking itself — nobody is waiting on any of these, so none of them should be able to evict the model someone is actively chatting with. All three could.

**The information was never missing — it was thrown away.** Every trigger already tags its `DispatchRequest` with a truthful `source` (`.schedule`, `.watcher`, `.selfSchedule`), and `ExecutionContext` copies it onto the `ChatSession`. Then `ChatSession` built its engine as `ChatEngine(source: .chatUI)`, unconditionally, ignoring its own `source` — so every headless run reached the model claiming to be the user typing in the chat window.

That mislabel had two teeth:
- **Load intent.** A headless run got the same right to evict a resident model as a human pressing send.
- **Residency.** `accelerateIdleUnloadAfterChatClose` only shortens residency for `.chatUI`-sourced models — so closing an unrelated chat window could accelerate the unload of the model a scheduled job was mid-run with.

### The subtlety, and why intent is NOT derived from `source`

A cron fire and the user pressing **Run Now** on that same schedule both arrive as `.schedule`. They call the same `executeSchedule`. Inferring intent from `source` would make the Run Now button refuse to load its own model whenever another model happened to be resident — breaking the feature in the name of protecting it. Same trap in the Next Run panel, where the automatic wake and the panel's own button share `makeDispatchRequest`.

So `loadIntent` is set at the **trigger boundary**, the one place that knows whether anyone is waiting. Timer fires, missed-schedule catch-up, watcher fires and agent self-wakes pass `.background`; every hand-pressed button and every waiting API client keeps the default `.interactive`. A caller who forgets fails safe — toward loading, not toward refusing.

`RequestSource` gains a `.scheduled` case rather than being made to lie with `.chatUI` or `.httpAPI`; the Insights log now says "Scheduled", which is what actually happened.

Tests pin the **trap itself**: Run Now must stay interactive, the Next Run panel button must not be background, and no engine may be constructed with a hardcoded `.chatUI` again.

---

## 2. The chat stats card invented a tok/s it never measured

It showed **`2397.3 tok/s • 7 tokens`**. No local model decodes at 2000+ tok/s. That number was not a measurement of anything.

`RollingTokenRate.finalRate()` fell back to `totalTokens / (lastAt - firstAt)` when the rolling window never converged. But `firstAt` is when the first token **arrived**, not when decoding began — so the span omits all the decode time that produced it. When a short reply is delivered in one coalesced burst, first and last arrival land ~3 ms apart: `7 / 0.003` = 2333 tok/s. **It measured the delivery, not the decode.**

`currentRate` already documents this exact hazard and floors its denominator against it ("would collapse an oldest-to-now span toward zero and report a physically impossible rate"). The fallback did the very thing that comment warns about. The existing test missed it because it only exercised the benign case — 3 tokens spread evenly over 0.2 s reads as a plausible 15 tok/s — and never a burst. **That plausibility is why the bug survived.**

There is no honest denominator available to that type: it only sees arrival timestamps. So it no longer guesses. `finalRate()` returns `nil` when the window never converged, and the caller falls back to the **engine's real decode rate** — tokens over the decode wall-clock, timed from the end of prefill — which the UI was already receiving in the stats sentinel and deliberately discarding.

Every rung of the ladder is a real measurement, and the last rung is *nothing*. A blank cell is honest; `2397.3 tok/s` was not.

(Engine-side companion, for the `NaN`/`+inf` the accessors could also emit: osaurus-ai/vmlx-swift#143.)

---

## Live proof (dev build, GUI only, computer-use)

**Stats card** — the exact case that used to fabricate thousands:

| reply | stats line (verbatim) |
|---|---|
| `yes` | `TTFT 0.21s • 25.5 tok/s • 1 tokens` |
| `Hello.` | `TTFT 0.23s • 24.7 tok/s • 2 tokens` |
| 200-word paragraph | `TTFT 0.57s • 38.0 tok/s • 183 tokens` |

A 1-token reply — the worst case for the old divide — now reports a rate in the same neighbourhood as the long-form one on the same binary, instead of 60× higher. **No value above 1000.** (These are Debug-build magnitudes and say nothing about actual decode speed; the point is the *shape*.)

That pass also surfaced `1 tokens`, fixed in the last commit.

**Interactive loading unaffected:**
- Explicit picker switch `Gemma 4 e2b it 4bit` → `Qwen3 0.6B 8bit` still evicts and loads — `Warming up…` → `Model warm — ready for a fast first response`.
- 3-turn context held (Paris → Seine/Aube/Marne → Seine). No mojibake, looping, truncation, unclosed reasoning, or stray template tags.
- No chat hung. **No error banner, no `Skipped background load`, no `would evict`.**

## Found, but NOT from this change

Schedule **"Run Now" never produces output** — sometimes stuck at `Running`, sometimes "Completed" with an empty result. **Reproduced identically on unmodified `main`** (baseline dev build), so it is pre-existing. App stdout shows the generation runs and finishes but emits nothing:

```
[Osaurus][Stream] Starting stream wrapper for model: mlx-community/gemma-4-e2b-it-4bit
[Osaurus][Stream] Stream completed: 0 content/reasoning deltas in 0.85s
                  classification=sentinel-only reasoning=0 stats=1 toolHints=0
```

Filed separately; it is the reason Run Now could not itself serve as the live gate here.

### DIFF
--- a/Packages/OsaurusCore/Managers/BackgroundTaskManager.swift
+++ b/Packages/OsaurusCore/Managers/BackgroundTaskManager.swift
@@ -666,7 +666,8 @@ public final class BackgroundTaskManager: ObservableObject {
             folderBookmark: request.folderBookmark,
             source: request.source,
             sourcePluginId: request.sourcePluginId,
-            externalSessionKey: request.externalSessionKey
+            externalSessionKey: request.externalSessionKey,
+            loadIntent: request.loadIntent
         )
     }
 
--- a/Packages/OsaurusCore/Managers/ExecutionContext.swift
+++ b/Packages/OsaurusCore/Managers/ExecutionContext.swift
@@ -41,7 +41,8 @@ public final class ExecutionContext: ObservableObject {
         folderBookmark: Data? = nil,
         source: SessionSource = .chat,
         sourcePluginId: String? = nil,
-        externalSessionKey: String? = nil
+        externalSessionKey: String? = nil,
+        loadIntent: ModelLoadIntent = .interactive
     ) {
         self.id = id
         self.agentId = agentId
@@ -56,6 +57,7 @@ public final class ExecutionContext: ObservableObject {
         // actual saved session.
         session.sessionId = id
         session.source = source
+        session.loadIntent = loadIntent
         session.sourcePluginId = sourcePluginId
         session.externalSessionKey = externalSessionKey
         session.dispatchTaskId = id
--- a/Packages/OsaurusCore/Managers/NextRunScheduler.swift
+++ b/Packages/OsaurusCore/Managers/NextRunScheduler.swift
@@ -266,7 +266,8 @@ public final class NextRunScheduler {
             return
         }
 
-        let request = await Self.makeDispatchRequest(for: entry)
+        // The agent woke itself on a timer. Nobody is waiting.
+        let request = await Self.makeDispatchRequest(for: entry, loadIntent: .background)
         guard let handle = await TaskDispatcher.shared.dispatch(request) else {
             print(
                 "[NextRunScheduler] dispatch failed for agent \(entry.agentId.uuidString.prefix(8))"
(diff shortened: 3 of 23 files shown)


jjang-ai:
Superseded — all three commits were folded into the checkpoint (#2015) and are on main:
- Stop headless runs from claiming to be the user
- Stop inventing a tok/s for replies too short to measure
- Say '1 token', not '1 tokens'

The tok/s fixes are live-verified on the merged build (no NaN, no Inf, no absurd rate).

Issue #148, opened 2025-11-26
Random output with some models.


aptonline:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest `main` or release

### Describe the bug

With the latest version (0.5.7) some models are generating random output.

<img width="724" height="722" alt="Image" src="https://github.com/user-attachments/assets/b304fcfb-5fcd-4680-8ffe-c73c4ffbe128" />

this is with deepseek 1

### Steps to reproduce

Entered a question.

### Osaurus version / commit

0.5.7

### macOS version

Sequoia 15.7.2

### Apple Silicon chip

M2 Max

### Logs

```shell

```

### Screenshots

as above

tpae:
Hey @aptonline , thanks for trying out the latest version. Does the specific model work for you in other apps and/or has worked well in the past?

aptonline:
It's the first time trying this model in Osaurus but I've used it elsewhere with no issue.  I just tried another model (gemma-3-270m-it-MLX-8bit) and that worked fine.

<img width="726" height="720" alt="Image" src="https://github.com/user-attachments/assets/87521d3c-1cc8-4505-9fa1-3aba44568562" />

tpae:
This should be fixed, but let me know if the issue persists @aptonline

PR #2029, opened 2026-07-13
Computer Use AX bounding + AccessibilityBoundedElement test fix + vmlx repin (SpecDec tail flush #148)

jjang-ai:
Consolidated PR.

## Changes
1. **Computer Use: bound the two AX calls on the open-app path** — fixes the first-GUI-action beachball (unbounded `AXUIElementCopyAttributeValue`/messaging on the open-app path). Live-proven via Computer Use (TextEdit opens, no beachball).
2. **Fix AccessibilityBoundedElementSource test path** — the AX source-guard tests resolved `packageRoot` to `Packages/OsaurusCore/Tests` (three `deletingLastPathComponent`) and appended `ComputerUse/Driver/Mac/Accessibility.swift`, producing a nonexistent `Tests/ComputerUse/...` path → both tests errored "no such file". One more deletion reaches `Packages/OsaurusCore`, where the source actually lives. This was the sole `test-core` failure.
3. **Repin vmlx-swift → a9b10f60 (#148)** — picks up the SpecDec fix that flushes the detokenizer's held-back tail at end of stream, so the last ≤24 characters of speculative-decode answers are no longer dropped. Updates Package.swift, all three Package.resolved, and the two revision tripwire tests.

## Verification
- vmlx@a9b10f60 resolves cleanly (`swift package resolve` OK).
- test-core previously failed ONLY on the two AX path tests (now fixed); test-cli / test-evals / swiftlint / shellcheck were green.
- AX bounding live-proven on the dev app.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "8f731c2388daeab1c257303bf653a00218ee51dd"
+        "revision" : "a9b10f60e330337a9de2d8ebe3ca74a7370525e4"
       }
     },
     {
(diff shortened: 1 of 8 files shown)


jjang-ai:
Folded the vmlx repin (→ a9b10f60, SpecDec detokenizer tail flush #148) and the AccessibilityBoundedElementSource test-path fix into this PR. The test-path off-by-one was the only `test-core` failure — the guard test looked under `Tests/ComputerUse/Driver/Mac/` instead of the real `ComputerUse/Driver/Mac/`. Re-running CI to confirm green before merge.

PR #2056, opened 2026-07-17
Reliability audit: transactional installs, web assets, key pinning, strict verify

tpae:
## Summary

Host/harness fixes from the July 2026 plugin reliability audit (install-integrity theme). Part of a 5-branch series; each branch is independent and separately mergeable. Full campaign notes cover all five themes.

### 1. `audit/install-integrity` (3 commits)

Commits:
- `773df291` Enforce host/macOS version checks, TOFU key pinning, single dylib, and web asset install in registry installer
- `5fb17ae0` Fail tools verify on missing or malformed receipts and dylibs
- `686608bb` Use bounded extraction, atomic publication, and explicit consent for manual tools install

Confirmed defects fixed:
- **High** `Packages/OsaurusRepository/PluginInstallManager.swift` — registry installs copied only the dylib (+ SKILL.md/docs); the `web/` directory promised by `docs/plugins/PACKAGING.md` was silently dropped, so installed plugins' static UIs 404'd. `installVerifiedArtifact` now locates and copies the artifact's `web/` directory (preferring the one next to the dylib).
- **High** `PluginInstallManager.swift` + `PluginSpec.swift` — `resolveBestVersion` was called with `minimumOsaurusVersion: nil` (so `osaurus_min_version` never filtered) and `PluginArtifact.min_macos` was never consulted. Resolution now receives the real host version (`CFBundleShortVersionString`) and `ProcessInfo.operatingSystemVersion`, and filters artifacts by `min_macos`.
- **High** `PluginInstallManager.swift` — `PluginInstallError.authorKeyMismatch` existed but was never thrown; a changed registry signing key was silently trusted (TOFU without the "trust" part) and the previous version was deleted before the new install proved good. Now: pinned-key mismatch fails closed with `authorKeyMismatch` (the existing UI alert directs users through explicit uninstall/reinstall approval), and the rollback version is only deleted after the install fully succeeds.
- **Medium** `PluginInstallManager.swift` — artifact containing zero or multiple dylibs nondeterministically picked the first; now `locateSingleDylib` requires exactly one and errors otherwise.
- **High** `Packages/OsaurusCLI/Sources/OsaurusCLICore/Commands/Tools/ToolsVerify.swift` — missing/unreadable `receipt.json`, missing dylibs, and dangling `current` symlinks were silently `continue`d, so `osaurus tools verify` exited 0 for broken installs. Every missing/malformed/mismatched file is now a reported failure and the command exits non-zero.
- **High** `Packages/OsaurusCLI/Sources/OsaurusCLICore/Commands/Tools/ToolsInstall.swift` — manual remote install buffered the whole archive in memory (`URLSession.data`), shelled out to `/usr/bin/unzip` (no entry/size/ratio/path-escape limits), replaced an existing version non-transactionally (delete-then-copy), and granted consent by default. Now: streaming download (`URLSession.download`), extraction through the same `BoundedArchiveExtractor` the registry installer uses (via a new `PluginInstallManager.extractPluginArchive` facade), staging + atomic publication (`replaceItemAt`), consent only with an explicit `--consent` flag, and failure when the payload produces no dylib (receipt creation now throws instead of silently skipping).

Tests added:
- `Packages/OsaurusRepository/Tests/OsaurusRepositoryTests/PluginInstallIntegrityTests.swift` — min_osaurus/min_macos filtering, pinned-key mismatch, single-dylib enforcement, web directory discovery/copy, verified-artifact install shape.
- `Packages/OsaurusCLI/Tests/OsaurusCLITests/ToolsVerifyTests.swift` — missing receipt, malformed receipt, missing dylib, SHA mismatch, dangling `current` symlink all fail; healthy install passes.
- `Packages/OsaurusCLI/Tests/OsaurusCLITests/ToolsManualInstallPublishTests.swift` — atomic publish over existing version, no-dylib failure, consent-only-when-granted.

Verification (all green):
- `swift test --package-path Packages/OsaurusRepository --parallel` — 77 tests passed.
- `swift test --package-path Packages/OsaurusCLI --parallel` — 130 tests passed.
- `swift build -c release --package-path Packages/OsaurusRepository` and `--package-path Packages/OsaurusCLI` — build complete.

### 2. `audit/loader-abi` (1 commit)

Commit:
- `82135263` Validate plugin ABI table, v1 prefix decode, and manifest/tool/route identity at load

Confirmed defects fixed (all in `Packages/OsaurusCore/Managers/Plugin/PluginManager.swift` and `Models/Plugin/ExternalPlugin.swift`):
- **High** — loader only required `init` and `get_manifest`; nil `free_string` (leaks every returned string), `destroy` (no teardown), or `invoke` (broken tools) were accepted. `abiTableValidationFailure` now rejects incomplete ABI tables before `init` is called.
- **High** — the v1 entry path read `apiPtr.pointee` as the full current `osr_plugin_api` struct, reading past the end of a historical v1 plugin's static struct (out-of-bounds read of whatever follows it in the binary). New `osr_plugin_api_v1` prefix struct decodes exactly the five required members and widens into the full layout with v2+ slots zeroed.
- **Medium** — no load-time identity validation: manifest `plugin_id` could differ from the receipt/install-directory id (previously "re-keyed" silently), two plugins could load under the same id, and duplicate/empty tool or route ids were accepted. Load now fails on manifest/directory id mismatch, rejects duplicate loaded plugin ids (newcomer is shut down), and rejects empty or duplicate tool/route ids.

Tests added:
- `Packages/OsaurusCore/Tests/Plugin/PluginLoaderAbiValidationTests.swift` — ABI completeness matrix, v1 prefix widening (zeroed v2+ slots, version 0), id-mismatch failure, duplicate/empty tool and route id failures.

Verification (all green):
- `swift build --package-path Packages/OsaurusCore` — build complete.
- `swift test --package-path Packages/OsaurusCore --filter 'PluginLoaderAbiValidationTests|PluginAbiHandshakeTests'` — 28 tests passed.

### 3. `audit/route-reliability` (1 commit)

Commit:
- `22b84a1f` Fix route handler timeout race and segment-boundary web mount matching

Confirmed defects fixed:
- **High** `Packages/OsaurusCore/Models/Plugin/ExternalPlugin.swift` — `handleRoute` implemented its 30s timeout with `withThrowingTaskGroup`; task groups drain children before returning, so a blocking C route callback (which never observes Swift cancellation) held the HTTP caller for the full duration of the call — the timeout never fired from the request's perspective. Replaced with the unstructured-body-task + GCD-timer race already used by `ToolRegistry.runToolBody` (new `RouteHandlerRaceState`), so the caller is resumed the moment the timer wins while the abandoned callback finishes on the plugin's invoke queue.
- **Medium** `Networking/HTTPHandler.swift` + `Managers/Plugin/PluginManager.swift` — static web mount dispatch used bare `subpath.hasPrefix(mountPrefix)`, so mount `/ui` captured `/ui-other`, disagreeing with the load-time web/route overlap validation (which already used segment-boundary logic). Both now call one shared helper, `PluginManifest.WebSpec.mountCaptures` / `relativeSubpath` / `normalizedMount`, so validation and dispatch cannot diverge.

Tests added:
- `Packages/OsaurusCore/Tests/Plugin/PluginRouteReliabilityTests.swift` — elapsed wall-time assertion with a deliberately blocking C route callback (semaphore-held; asserts the caller returns at the 0.5s timeout instead of blocking until release), fast-path control, and mount matching for `/ui`, `/ui/x`, `/ui-other`, `/`, plus normalization (`ui`, `/ui/`) and relative-subpath expectations.

Verification (all green):
- `swift build --package-path Packages/OsaurusCore` — build complete.
- `swift test --package-path Packages/OsaurusCore --filter 'PluginRouteReliabilityTests|PluginRoutingTests'` — 22 tests in 6 suites passed (includes existing routing/static-file containment suites).

### 4. `audit/lifecycle-config` (1 commit)

Commit:
- `c601dd68` Order plugin teardown before secret sweep on agent delete; unify config default resolution

Confirmed defects fixed:
- **High** `Packages/OsaurusCore/Managers/AgentManager.swift` + `Managers/Plugin/PluginManager.swift` — `AgentManager.delete(id:)` swept `ToolSecretsKeychain` for the agent BEFORE posting `.agentRemoved`; the `.agentRemoved` handler was additionally fire-and-forget (Task + detached push), so plugins reading config (e.g. Telegram's `bot_token`) during webhook deregistration found nothing — the webhook stayed registered upstream. New awaitable `PluginManager.tearDownPluginsForRemovedAgent(agentId:)` delivers `tunnel_url=""` to every routed plugin via the synchronous config path and returns only after each plugin's `on_config_changed` completes; `delete` awaits it and only then sweeps. The notification-driven handler remains as an idempotent backstop (deduped plugin-side).
- **Medium** `Services/Keychain/ToolSecretsKeychain.swift` + `Services/Plugin/PluginHostAPI.swift` + `Managers/Plugin/PluginManager.swift` — inconsistent default semantics: tool payload injection merged `Agent.defaultId` values as global defaults (`resolvedSecretsWithDefaults`), but `config_get`, initial config delivery, and required-secret checks (`hasAllRequiredSecrets` / `getMissingRequiredSecrets`) read only the exact agent namespace — a key saved once on the Plugins tab was "configured" for injection but invisible to `config_get`. Centralized one policy, `ToolSecretsKeychain.resolvedSecret(id:for:agentId:)` (exact agent first, then default-agent fallback), and pointed all three paths at it. The anonymous-context guard (no agent → nil) is unchanged; UI editors (`PluginConfigView`, `ToolSecretsSheet`) intentionally keep exact-namespace reads since they edit a specific agent's values.

Tests added:
- `Packages/OsaurusCore/Tests/Plugin/AgentLifecycleConfigResolutionTests.swift` — ordering test driving a real `ExternalPlugin` whose C `on_config_changed` reads `bot_token` mid-teardown (asserts the token is readable during deregistration and swept afterwards, all before `delete` returns); resolution-policy tests: exact-wins, default fallback, missing-everywhere, required-secret checks honoring the fallback, and `config_get` through a TLS-scoped `PluginHostContext` (fallback + exact + anonymous-refusal).

Verification (all green):
- `swift build --package-path Packages/OsaurusCore` — build complete.
- `swift test --package-path Packages/OsaurusCore --filter 'AgentDeletionPluginTeardownOrderingTests|ToolSecretsResolutionPolicyTests|AgentManagerLifecycleNotificationTests|ResolvedSecretsMergingTests'` — 18 tests in 4 suites passed (includes the pre-existing lifecycle-notification and secrets-merge suites, confirming no ordering regressions).

### 5. `audit/ci-coverage` (1 commit)

Commit:
- `a6e1737b` Add CI lanes for OsaurusRepository, OsaurusPluginTestKit, and StatsPack tests

Changes (`.github/workflows/ci.yml`):

### DIFF
--- a/Packages/OsaurusCLI/Sources/OsaurusCLICore/Commands/Tools/ToolsInstall.swift
+++ b/Packages/OsaurusCLI/Sources/OsaurusCLICore/Commands/Tools/ToolsInstall.swift
@@ -13,24 +13,19 @@ public struct ToolsInstall {
     public static func execute(args: [String]) async {
         guard let src = args.first, !src.isEmpty else {
             fputs(
-                "Usage: osaurus tools install <plugin_id|url-or-path> [--version <semver>] [--no-consent]\n",
+                "Usage: osaurus tools install <plugin_id|url-or-path> [--version <semver>] [--consent]\n",
                 stderr
             )
             exit(EXIT_FAILURE)
         }
 
-        // The release-build loader requires a `.user_consent` marker
-        // alongside `receipt.json` before it will dlopen a plugin. The
-        // registry-install path (`PluginInstallManager.install`) writes
-        // it automatically; manual sideload via `tools install` should
-        // be symmetric so authors iterating with `tools install
-        // ./my-plugin-1.0.0` don't get a silent `consent_required:`
-        // load failure they then have to clear in the UI. `--no-consent`
-        // opts out for users who want the explicit Approve gate.
-        let grantConsent = !args.contains("--no-consent")
-
         // Check if argument is a local path or URL
         if isManualInstallSource(src) {
+            // Manual sideloads bypass the registry's checksum/signature
+            // verification, so the release loader's `.user_consent` gate is
+            // NOT waived implicitly: pass `--consent` to grant it at install
+            // time, otherwise approve the plugin in Osaurus settings.
+            let grantConsent = args.contains("--consent")
             await installManual(src: src, grantConsent: grantConsent)
         } else {
             await installFromRegistry(pluginId: src, args: args)
@@ -93,7 +88,11 @@ public struct ToolsInstall {
         var sourceName: String = ""
         var preferManifestIdentity = false
 
-        // 1. Unpack/Copy to staging
+        // 1. Unpack/Copy to staging. Archives go through the same bounded
+        // in-process extractor as registry installs (path-safety, resource
+        // limits) instead of shelling out to /usr/bin/unzip, and remote
+        // archives are streamed to disk instead of buffered in memory.
+        let stageRoot = tmpDir.appendingPathComponent("extracted", isDirectory: true)
         if src.hasPrefix("http://") || src.hasPrefix("https://") {
             guard let url = URL(string: src) else {
                 fputs("Invalid URL: \(src)\n", stderr)
@@ -103,13 +102,14 @@ public struct ToolsInstall {
             sourceName = url.lastPathComponent
             let zipFile = tmpDir.appendingPathComponent("download.zip")
             do {
-                let (data, resp) = try await URLSession.shared.data(from: url)
+                let (downloadedURL, resp) = try await URLSession.shared.download(from: url)
                 guard let http = resp as? HTTPURLResponse, (200 ... 299).contains(http.statusCode) else {
+                    try? fm.removeItem(at: downloadedURL)
                     fputs("Download failed (status \((resp as? HTTPURLResponse)?.statusCode ?? -1))\n", stderr)
                     exit(EXIT_FAILURE)
                 }
-                try data.write(to: zipFile)
-                try unzip(zipURL: zipFile, to: tmpDir)
+                try fm.moveItem(at: downloadedURL, to: zipFile)
+                try PluginInstallManager.extractPluginArchive(at: zipFile, to: stageRoot)
             } catch {
                 fputs("Download/Unzip error: \(error)\n", stderr)
                 exit(EXIT_FAILURE)
@@ -121,11 +121,12 @@ public struct ToolsInstall {
             if fm.fileExists(atPath: pathURL.path, isDirectory: &isDir) {
                 if isDir.boolValue {
                     preferManifestIdentity = true
-                    // Copy contents to tmpDir
+                    // Copy contents to the staging root
                     do {
+                        try fm.createDirectory(at: stageRoot, withIntermediateDirectories: true)
                         let contents = try fm.contentsOfDirectory(at: pathURL, includingPropertiesForKeys: nil)
                         for item in contents {
-                            try fm.copyItem(at: item, to: tmpDir.appendingPathComponent(item.lastPathComponent))
+                            try fm.copyItem(at: item, to: stageRoot.appendingPathComponent(item.lastPathComponent))
                         }
                     } catch {
                         fputs("Failed to copy directory: \(error)\n", stderr)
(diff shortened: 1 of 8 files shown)


tpae:
Consolidated into #2061 (one PR per repo).

PR #2101, opened 2026-07-20
Preserve explicit Thinking mode across agent tool turns

jjang-ai:
## Scope

This is a four-file emergency hardening for the Chat UI agent loop only.

- Freeze prompt-affecting model options once at user send time.
- Apply the same options to every reconstructed agent request and the iteration-cap finalizer.
- Emit the official Codable `enable_thinking` field alongside local `modelOptions`.
- Preserve `nil` when the user has not made an explicit choice, so the bundle template remains authoritative.

No Gemma, cache, Sentry, AppleScript/Zaya, routing, sampler, template, or model-bundle changes are included.

## Source finding and claim boundary

Exact model bundle: `/Users/eric/models/JANGQ-AI/Ornith-1.0-9B-JANG_4M`.

Its template gates non-thinking mode on an explicit false value:

`{% if enable_thinking is defined and enable_thinking is false %}`

Therefore an absent kwarg means Thinking ON for Ornith. Osaurus 0.22.5 and current main already detect that default for the UI, and the pre-change ChatView copied local `activeModelOptions` into each ordinary loop request and finalizer. I did not reproduce the reported current-main condition where explicit OFF is lost only after a tool call. This PR is a contract hardening: it freezes the explicit per-turn state and also carries it in the official wire-safe field. It is not presented as proof that the existing local path dropped the toggle.

This evidence points away from content-delta streaming or tool JSON schema as the cause of the hidden-reasoning slowdown. The observed speed difference follows the template/UI Thinking state.

## Tests

Current result bundles from the candidate source:

- `/private/tmp/osaurus-ornith-thinking-focused-6.xcresult`: 4 passed, 0 failed.
- `/private/tmp/osaurus-ornith-thinking-source-6.xcresult`: 92 passed, 0 failed.
- Release app build completed from base `5202bfe95a8905043251a3a58f1ec4a9a026a305` plus commit `73cedec3f`.
- Candidate app is ad-hoc signed and `codesign --verify --deep --strict` succeeds.
- Bundle ID: `com.dinoki.osaurus.orniththinkingproof`.
- CDHash: `4b582bf516c29bab2215b9557a65aaba4543ee54`.

## Live Computer Use proof

Candidate: `/private/tmp/Osaurus-Ornith-Thinking-Propagation.app`

Fresh isolated root: `/private/tmp/osaurus-ornith-thinking-live-root-20260720`

The exact Ornith JANG_4M model was selected through Settings after adding `/Users/eric/models` as an External Model Source. The toggle was operated in the actual model picker. The isolated preference file independently recorded `disableThinking=true` for OFF and `disableThinking=false` for ON.

Repeated identical multi-call Todo prompt:

| State | Session | Model steps | Persisted thinking | Visible result |
|---|---|---:|---:|---|
| explicit OFF | `EDB961F5-D04E-4478-8BF1-8887F2249177` | 4 | 0 chars on all 4 | no Thinking blocks |
| explicit ON | `B1AA50BA-C431-42EF-88E1-D362646F26E3` | 4 | 348, 255, 292, 628 chars | four Thinking blocks; final row 0.27s TTFT, 58.0 tok/s, 190 tokens |
| explicit OFF again | `BB5CF0A9-8A34-4115-8972-BDA3680F7821` | 4 | 0 chars on all 4 | no Thinking blocks; final text row 0.40s TTFT, 90.2 tok/s, 57 tokens |

This proves the candidate honors OFF -> ON -> OFF on every observed model/tool iteration through the real UI. It does not prove a before/after repair of current main because current-main source already propagated the local option.

## Adjacent failure kept visible

The model did not satisfy the Todo semantics in any run: final state was 1/3, 0/3, and 1/3 respectively instead of completing alpha and beta. One run also falsely summarized completion. That is a separate tool-selection/argument correctness defect; this PR does not mask it with a prompt, sampler, parser, or close-token guard.

The first OFF run ended tool-only and exposed no tok/s metric. The repeated OFF run did expose 90.2 tok/s. No throughput claim is made beyond those visible rows.

### DIFF
--- a/Packages/OsaurusCore/Services/Chat/ChatTurnGenerationControls.swift
+++ b/Packages/OsaurusCore/Services/Chat/ChatTurnGenerationControls.swift
@@ -1,0 +1,42 @@
+//
+//  ChatTurnGenerationControls.swift
+//  osaurus
+//
+//  Freezes prompt-affecting UI model controls for one logical chat turn.
+//  Agent loops rebuild ChatCompletionRequest for every model step, so the
+//  explicit Thinking choice must be copied to every reconstruction in both
+//  its local runtime representation (`modelOptions`) and its wire-safe API
+//  representation (`enable_thinking`).
+//
+
+import Foundation
+
+struct ChatTurnGenerationControls: Sendable, Equatable {
+    let modelOptions: [String: ModelOptionValue]?
+    let enableThinking: Bool?
+
+    static func capture(
+        activeModelOptions: [String: ModelOptionValue]
+    ) -> ChatTurnGenerationControls {
+        let options = activeModelOptions.isEmpty ? nil : activeModelOptions
+
+        // Only an explicit toggle becomes an API override. An absent value
+        // remains nil so the bundle's chat-template default still owns fresh
+        // models. `disableThinking` is the canonical stored option for every
+        // current Thinking profile, including dynamically discovered local
+        // profiles such as Ornith. Translate it directly instead of resolving
+        // the model profile again on the send path; dynamic profile discovery
+        // can touch the local model catalog and must not add synchronous TTFT.
+        let enableThinking = options?["disableThinking"]?.boolValue.map { !$0 }
+
+        return ChatTurnGenerationControls(
+            modelOptions: options,
+            enableThinking: enableThinking
+        )
+    }
+
+    func apply(to request: inout ChatCompletionRequest) {
+        request.modelOptions = modelOptions
+        request.enable_thinking = enableThinking
+    }
+}
--- a/Packages/OsaurusCore/Tests/Chat/ChatTurnGenerationControlsTests.swift
+++ b/Packages/OsaurusCore/Tests/Chat/ChatTurnGenerationControlsTests.swift
@@ -1,0 +1,101 @@
+//
+//  ChatTurnGenerationControlsTests.swift
+//  osaurusTests
+//
+
+import Foundation
+import Testing
+
+@testable import OsaurusCore
+
+@Suite("Chat turn generation controls")
+struct ChatTurnGenerationControlsTests {
+    private func request() -> ChatCompletionRequest {
+        ChatCompletionRequest(
+            model: "JANGQ-AI/Ornith-1.0-9B-JANG_4M",
+            messages: [ChatMessage(role: "user", content: "Use three tools")],
+            temperature: nil,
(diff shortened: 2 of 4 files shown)


jjang-ai:
Final pre-merge evidence at head `73cedec3fabd9fc2b2f87f0fbae51a96a0fb7d54`:

- Scope re-audited through the GitHub diff: exactly four files (`ChatTurnGenerationControls.swift`, its focused tests, `ChatView.swift`, and the source-policy test). No Gemma, cache, Sentry, AppleScript/Zaya, routing, sampler, or bundle changes.
- Source trace: the helper captures explicit `disableThinking` once per user turn, translates it once to official `enable_thinking`, and applies both to every normal ChatView request plus the iteration-cap finalizer. ChatEngine makes the official field authoritative and vMLX maps the resulting option to Jinja. HTTP and plugin continuation builders already preserve both fields.
- Claim boundary: Osaurus 0.22.5/current source already copied local `activeModelOptions` during its ChatView loop. The reported current-main mid-loop loss was not reproduced. Ornith defaults ON when the kwarg is absent; this patch hardens immutable per-turn and wire behavior rather than claiming a proven repair of the prior local loop.
- Live isolated Release candidate: `/private/tmp/Osaurus-Ornith-Thinking-Propagation.app`, bundle `com.dinoki.osaurus.orniththinkingproof`, exact model `/Users/eric/models/JANGQ-AI/Ornith-1.0-9B-JANG_4M`. Computer Use plus persisted DB showed OFF -> ON -> OFF across four model steps each: thinking chars `0/0/0/0`, then `348/255/292/628`, then `0/0/0/0`. Visible ON final row was 0.27s TTFT / 58.0 tok/s; repeated OFF final row was 0.40s TTFT / 90.2 tok/s.
- Local current-source results: focused 4 passed / 0 failed; source-policy 92 passed / 0 failed; ad-hoc signature verified, CDHash `4b582bf516c29bab2215b9557a65aaba4543ee54`.
- GitHub checks all succeeded: shellcheck, swiftlint, test-cli, test-core, test-packages, test-statspack, test-evals, and release drafter. PR state is CLEAN and MERGEABLE.
- Adjacent limitation remains explicit: Ornith failed the Todo semantics in all three live runs (1/3, 0/3, 1/3). This PR does not hide that separate model/tool-argument correctness defect with parser, sampler, prompt, or token guards.

PR #2155, opened 2026-07-24
nudge model to continue when run ends with unchecked todo items

### LINKED ISSUE #2133: Osaurus fails to follow "Todo" list of actions
### Checks

### Describe the bug

When using any model (Gemma 4, Qwen3.6-35B-A3B) and given a multi-step task the model will create a Todo action list and this is visible in the UI. Osaurus *might* execute the first task, but then fails to continue.

### Steps to reproduce

1. Give the model a multi-step task. Example task attached.

[Example-Task.md](https://github.com/user-attachments/files/30329116/Example-Task.md)

### Provider type

Local model (MLX)

### Model name

Qwen3.6-35B-A3B

### Osaurus version / commit

0.22.8

### macOS version

26.5.2

### Apple Silicon chip

M5 Max 128gb

### Xcode version

26.6

### Logs

```shell

```

### Screenshots

_No response_

RaajeevChandran:
## Summary

closes #2133

  - the agent loop ended on any text-only final response even when the session todo still had unchecked items 
  - now the loop bounces a premature stop back with a transient system notice telling the model to continue with the next unchecked item or check the boxes and finish
  - bounded at 2 consecutive nudges so a model that insists on stopping still ends the run, and a productive tool turn resets the allowance
  - adds tests for the nudge-then-accept bound, allowance reset, and unchanged behavior with no pending items

## Changes

- [x] Behavior change
- [ ] Refactor / chore
- [ ] Tests
- [ ] Docs

## Checklist

## eval verification

- ran `AgentLoop`, `AgentLoopFrontier`, and `CacheProof` suites on main (baseline) and this branch with `mlx-community/Qwen3.5-9B-MLX-4bit`
- `CacheProof` passed 8/8 on both branches, so the KV prefix cache concern is cleared by the live cache gates, not just theory
- measured worst case for the new user role nudge on a Qwen template: partial re-prefill of about 500 tokens (13 percent of the prompt, roughly 30 to 40 ms at logged prefill speeds), never a reset to 0
- the continuation nudge fired 24 times across the fix branch suites and flipped `no-clarify-when-default-obvious` from fail to pass
- 4 diff regressions were inspected and are sampling variance: in 3 of them no todo list was ever created so the changed code path never executed, and the agent loop runner samples at model default temperature
- the hard todo cases that still fail (`todo-discipline-multistep`, `long-horizon-project`, `multi-file-refactor-with-todo`) fail identically on both branches for pre-existing model quality reasons

### DIFF
--- a/Packages/OsaurusCore/Services/Chat/AgentToolLoop.swift
+++ b/Packages/OsaurusCore/Services/Chat/AgentToolLoop.swift
@@ -1062,6 +1062,20 @@ enum AgentToolLoop {
         "[System Notice] Your todo list still has \(pending) unchecked item\(pending == 1 ? "" : "s") and has not been updated recently. If you finished any, re-send the full list with those boxes checked now; if the plan changed, rewrite the list."
     }
 
+    /// Bounded number of times a text-only final response is bounced back
+    /// to the model while the session todo list still has unchecked items.
+    /// Small local models routinely narrate the next step in plain text
+    /// instead of calling a tool; without this guard the run ends with the
+    /// checklist abandoned (issue #2133).
+    static let maxPendingTodoContinuations = 2
+
+    /// Transient nudge staged when the model tried to end the run with
+    /// unchecked todo items remaining. Rides the same notice channel as the
+    /// staleness nudge — never persisted into history.
+    static func pendingTodoContinuationNotice(pending: Int) -> String {
+        "[System Notice] Your todo list still has \(pending) unchecked item\(pending == 1 ? "" : "s"). Do not stop here: continue with the next unchecked item using tool calls. If everything is actually done, re-send the full list with all boxes checked, then finish."
+    }
+
     // Capability schemas loaded mid-run are delivered append-only in the
     // `capabilities_load` tool *result* (see
     // `CapabilitiesLoadTool.loadedSchemaBlock`), never by rewriting the frozen
@@ -1151,6 +1165,10 @@ enum AgentToolLoop {
         // productive turn; bounds the nudge-and-retry recovery so the loop
         // can never spin on a deterministically-empty model.
         var consecutiveEmptyTurns = 0
+        // Text-only final responses bounced back because unchecked todo
+        // items remained. Bounded so a model that insists on stopping can
+        // still end the run.
+        var pendingTodoContinuations = 0
         // Last iteration that carried a `todo` call (0 = run start), for
         // the staleness check below.
         var lastTodoIteration = 0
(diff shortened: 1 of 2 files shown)


jjang-ai:
I reproduced the opposite behavior in the real app and this continuation policy is unsafe as written.

1. `pendingTodoContinuations` is not actually a run-wide bound. It resets to `0` after every tool turn. An alternating sequence such as `final -> nudge -> tool -> final -> nudge -> tool ...` can therefore continue until the outer iteration limit, including repeating a successful side effect. The reported Reminders trace followed this shape: the reminder succeeded, then the run emitted multiple success finals, added new unchecked Todo items, shared an unnecessary artifact, and continued repeating.

2. The pending Todo belongs to session state, so this rejects an otherwise valid final on a later or unrelated run whenever stale unchecked items remain. In exact Release-app proof, a stale `Todo 0/1` followed by “What is two plus two?” must terminate after the single answer; treating Todo as a run lock instead reopens it.

3. A successful `todo` call with an unchecked advisory item followed by a normal final must also terminate. The model may intentionally leave follow-up work unchecked, or the checklist may simply be stale. The final response is the authoritative turn boundary; Todo is progress metadata, not execution authority.

4. The scripted tests here encode the problematic behavior (three accepted final attempts) and omit the critical rows: successful consequential tool -> final exactly once, stale Todo across a new run, unchanged Todo update/no timestamp churn, Stop/cancel, delayed no-repeat UI finalization, and a follow-up turn.

The cache measurements in the PR body are useful but do not validate termination correctness. PR #2175 addresses this at the owning contract instead: ordinary finals remain authoritative, Todo is advisory, semantically unchanged Todo writes are no-ops, and the exact stale/current-run Todo cases were exercised in the rebased Release app plus Bonsai/Ornith/Gemma agent-loop evals. I recommend closing this PR as superseded rather than rebasing the forced-nudge behavior.

jjang-ai:
Current-source review: this approach should not be reopened as written.

The core problem is scope. `pendingTodoCount` reads the session-persistent
checklist on every text final, but a pinned Todo may belong to an earlier user
turn. Therefore this patch can bounce a valid unrelated final solely because
old UI state remains unchecked. It also resets `pendingTodoContinuations` after
every tool turn, so a model alternating one tool call with premature prose can
re-arm the two-nudge budget repeatedly rather than having one bounded logical
run budget. The tests cover a current pending list but not stale prior-turn
Todo, batched tool execution, follow-up finalization, Stop/input unlock, or
post-final idle behavior.

#2192 instead gives each logical `AgentToolLoop.run` a TaskLocal Todo ownership
scope while leaving the visible checklist session-persistent. Only a valid Todo
call in that same run can affect `complete`; an old checklist gets typed reason
`stale_session_todo` and the loop recovers to one normal final. It has
current-source Release UI proof for the stale-follow-up and direct-complete
paths, plus the following scores.

OsaurusEval (Osaurus 78e12abe, vMLX d2f6f982, 3 trials/model, disk-L2 with paged RAM off):
- AgentLoop — Gemma 4 26B A4B JANG_4M: 3/3 PASS — finalResponse, 4 model steps, 35.3 tok/s, TTFT 211 ms, L2 +3 hits/+10 stores.
- AgentLoop — Bonsai 27B Ternary JANG: 2/3 FLAKY — two strict passes; failed trial looped on Todo/tool choice before writing. Passing trial: finalResponse, 4 steps, 18.4 tok/s, TTFT 281 ms.
- AgentLoop — Ornith 9B JANG_4M: 1/3 FLAKY — strict failures were incorrect/extra final prose despite correct file_write, not repeated terminal confirmations. Passing trial: finalResponse, 4 steps, 31.3 tok/s, TTFT 318 ms.
- CacheProof — Gemma 4 26B A4B JANG_4M: 3/3 PASS — SSD restore 293/705 -> 704/705 (+411), L2 +2 hits/+13 stores, 33.5 tok/s.
- CacheProof — Bonsai 27B Ternary JANG: 3/3 PASS — hybrid SSD restore 295/716 -> 709/716 (+414), SSM +2 hits/+0 rederives, L2 +2 hits/+4 stores, 18.5 tok/s.
- CacheProof — Laguna S 2.1 JANG_4M: 3/3 PASS — SSD restore 291/713 -> 712/713 (+421), L2 +2 hits/+11 stores, 27.8 tok/s.
- Focused Xcode suites: 115/115 PASS, 0 failed, 0 skipped.
No external judge key was configured; model rows above use deterministic exit/tool/file/cache assertions. Bonsai and Ornith quality rows remain explicitly PARTIAL.

Issue #1091, opened 2026-05-14
Add global HTTP/SOCKS5 proxy support (for all network access)


fortinmike:
### Checks

- [x] I have searched existing issues and discussions

### Area

Other

### Problem statement

There are many reasons to proxy traffic but a very good one in my opinion is to prevent your home IP from being blacklisted due to your machine performing scraping using the Search plugin repeatedly. Making the searches that the agent performs more private is another good reason.

Using your VPN's proxy URL allows you to proxy all of the agent's traffic through that private proxy without needing to pipe your whole machine's traffic through a VPN connection, which has many downsides. Mullvad, Windscribe, NordVPN, IVPN, Proton VPN and other providers offer this service.

Related to #232 but not limited to specific areas of the app. This is the way most apps do it; they provide a global field to set a proxy URL that will be used for all network requests. In terms of UX at least it would also be way simpler than trying to scope it to specific features of the app.

### Proposed solution

Add a global field in settings to allow specifying a proxy URL to use for all network connections including model downloads, external provider requests, fetch and search tool requests, etc. Some care must be taken to avoid DNS leaks, etc.

### Contribution

- [ ] I would be willing to submit a PR for this feature

mimeding:
Triage `main` @ `2df5d3a029ffae4bf7ff0ac29687c378b9a0051b` (2026-05-26 bugfix sweep). Please retest on 0.18.38+; reply with fresh repro if still broken.

mimeding:
**Triage (BASE `2df5d3a029ffae4bf7ff0ac29687c378b9a0051b`)**

Likely **fixed by #1214** (global proxy). Please **retest on 0.18.38+** and comment if the issue persists.

PR #2168, opened 2026-07-26
Wire global proxy into search and fetch tool requests

peterlodri-sec:
## Summary
Closes the remaining gap in #1091. PR #1214 (and follow-ups #1565, #1693) already added global HTTP/SOCKS5 proxy support and wired it into most network call sites — model downloads, provider requests, MCP transport, Discord/Telegram/Slack clients, etc. Two surfaces #1091 names explicitly were never migrated:

- **Search** — `SearchHTTPClient.request()` (`Services/Search/SearchTypes.swift`), used by every search backend, called `URLSession.shared` directly.
- **Fetch/extract** — `SearchReadability.extract()` (`Services/Search/SearchReadability.swift`), backing the `search_and_extract` tool, built its own `URLSession(configuration:delegate:delegateQueue:)` per call with no proxy configuration.

This closes both, following the existing pattern used by `GitHubSkillService`, `ThemesAPIClient`, and others.

## Changes
- `SearchHTTPClient`: added a `makeSession()` factory backed by `GlobalProxySettings.sharedSession()`, exposed separately from `request()` (which now calls it) for testability.
- `SearchReadability`: added a `makeSession(configuration:delegate:)` factory backed by `GlobalProxySettings.makeSession(base:delegate:delegateQueue:)`. Delegate has a default value so tests don't need access to the private `SearchReadabilityRedirectDelegate` type; `extract()` still passes its own instance explicitly so it can read `.blockedReason` afterward — that behavior is unchanged.
- Added `SearchHTTPClientGlobalProxyTests` and `SearchReadabilityGlobalProxyTests`, following the `<Subject>GlobalProxyTests.swift` convention already established for this feature (proxy-configured session picks up the right `connectionProxyDictionary`; unconfigured session has none).

## Test plan
- [x] `swift test --package-path Packages/OsaurusCore --filter GlobalProxy` — 29/29 tests pass across all 13 proxy suites (11 pre-existing + 2 new)
- [x] `swift test --package-path Packages/OsaurusCore --filter Search` — 279/279 pass on a clean run. (One earlier run of this same broad filter reported 3 failures elsewhere in the 279; an immediate re-run passed clean, so that looks like pre-existing flakiness unrelated to these two files — flagging for visibility, not something I chased down further since it reproduces outside the files this PR touches.)
- [x] `swift build --package-path Packages/OsaurusCore` — clean, no new warnings

## Out of scope (real gaps, left for follow-up PRs)
While tracing every proxy call site I found a few more real, unproxied outbound-HTTP surfaces that #1091 doesn't name explicitly, so I left them out of this PR to keep it reviewable:
- `Managers/OpenAICompatibleTTSClient.swift` — external TTS provider streaming
- `PrivacyFilter/Rampart/RampartModelManager.swift` — a second NER-model downloader, inconsistent with its sibling `PrivacyFilterModelDownloader.swift` which *is* wired
- `Services/AgentChannel/AgentChannelCustomJSONRunner.swift` — the generic/custom agent-channel HTTP client
- `OsaurusCLI` package — `URLSession.shared` throughout, even though `OsaurusCLICore` transitively depends on the already-wired `OsaurusRepository`

Happy to open separate PRs for these if useful.

### DIFF
--- a/Packages/OsaurusCore/Services/Search/SearchReadability.swift
+++ b/Packages/OsaurusCore/Services/Search/SearchReadability.swift
@@ -51,6 +51,20 @@ enum SearchReadability {
         var extracted: Bool { status == .ok }
     }
 
+    /// Exposed separately from `extract` so tests can inspect the session's
+    /// proxy configuration directly, matching the pattern used by the other
+    /// `GlobalProxySettings`-backed call sites. The default delegate value
+    /// (rather than requiring callers to pass one) means tests don't need
+    /// access to the private redirect delegate type at all; `extract` still
+    /// passes its own instance explicitly so it can read `.blockedReason`
+    /// afterward.
+    static func makeSession(
+        configuration: URLSessionConfiguration = .ephemeral,
+        delegate: URLSessionTaskDelegate = SearchReadabilityRedirectDelegate()
+    ) -> URLSession {
+        GlobalProxySettings.makeSession(base: configuration, delegate: delegate, delegateQueue: nil)
+    }
+
     /// Fetch `url` and extract the main content. Always returns a typed status
     /// so tool payloads can explain failures without raw network errors.
     static func extract(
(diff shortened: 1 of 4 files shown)


tpae:
@peterlodri-sec can you fix the build

PR #2034, opened 2026-07-15
Update Simplified Chinese and Traditional Chinese translations

ftzahao:
## Summary

Update Simplified Chinese and Traditional Chinese translations

## Changes

Update Simplified Chinese and Traditional Chinese translations

## Checklist

- [x] I have read `CONTRIBUTING.md`
- [ ] I added/updated tests where reasonable
- [ ] I updated docs/README as needed
- [x] I verified build on macOS with Xcode 16.4+

PR #1947, opened 2026-07-09
Update Simplified Chinese and Traditional Chinese translations

ftzahao:
## Summary

Update Simplified Chinese and Traditional Chinese translations

## Changes

Update Simplified Chinese and Traditional Chinese translations

## Checklist

- [x] I have read `CONTRIBUTING.md`
- [ ] I added/updated tests where reasonable
- [ ] I updated docs/README as needed
- [x] I verified build on macOS with Xcode 16.4+

Issue #1857, opened 2026-07-04
Use "早安" instead of "早上好" for Traditional Chinese (zh-Hant)


Flyingcat10235:
### Checks

- [x] I have searched existing issues and discussions

### Area

Chat / conversation

### Summary

The Traditional Chinese (zh-Hant) interface currently displays the greeting:

> 早上好

While this is understandable, it is not the natural greeting used by most Traditional Chinese speakers in Taiwan.

The greeting that users naturally expect is:

> 早安

"早上好" is commonly used in Mainland China, while "早安" is the standard and more natural morning greeting for Traditional Chinese users in Taiwan.

I noticed that Traditional Chinese (zh-Hant) is currently marked as **Help wanted** in the localization documentation, so I hope this suggestion can help improve the localization quality.

Thank you for your great work on Osaurus!

### Proposed solution

Use locale-specific translations instead of sharing the same Chinese wording.

Suggested localization:

- zh-Hant → 早安
- zh-Hans → 早上好

If the greeting changes according to the time of day, the Traditional Chinese localization should use wording that is natural for Traditional Chinese users.

### Alternatives considered

N/A

This is a localization improvement rather than an alternative feature.

### Additional context

<img width="912" height="753" alt="Image" src="https://github.com/user-attachments/assets/de4f7fd8-508a-4817-8ac8-160404920df8" />

### Contribution

- [ ] I would be willing to submit a PR for this feature

PR #2174, opened 2026-07-27
i18n(zh-Hant): restore Traditional character forms in 325 strings marked translated

YspritanHyzygy:
TRANSLATORS.md asks contributors to claim a language on the localization
tracking issue or via a draft PR. I couldn't find an open tracking issue, and
since the catalog moves quickly I thought a finished patch would be more useful
to you than a proposal — TRANSLATORS.md also notes that multiple contributors
can share a language, which is what I'm hoping to do here.

@ftzahao — you've done essentially all of the zh-Hant work (#1909, #1947,
#2034, #2076, #2141, #2172), so please read this as a heads-up rather than
anyone stepping on your toes. Happy to close it if you'd sooner fold the same
change into your next pass.

**Upfront about my qualifications, because it determines what I'm proposing:** I
read and write Chinese natively, but I'm a Simplified/mainland user, not a
Traditional Chinese one. So I am deliberately **not** proposing to make any
style, vocabulary, or variant-form judgment. This PR is scoped to the one slice
where there is nothing to judge — characters that exist only in Simplified and
have exactly one Traditional form under every convention.

## What I found

All 5,869 `zh-Hant` entries are marked `"state": "translated"` (zero
`needs_review`, unlike `de` or `ko`). Running OpenCC `s2tw` over every value
flags 758 that would change. Split by how much judgment each one actually needs:

| | Count | Who should decide |
|---|---:|---|
| **① Simplified-only characters with a single Traditional form** | **325** | nobody — mechanical |
| ② Variant-form normalization (爲/為, 啓/啟) | 302 | you / the project |
| ③ One-to-many mappings needing context (后→後/后, 发→發/髮) | 131 | a Traditional Chinese reader |

**This PR is group ① only.** ② and ③ are noted at the bottom, not included.

Group ① spans 174 distinct Simplified characters — 内→內 ×29, 选→選 ×28,
时→時 ×26, 运→運 ×21, 从→從 ×21, 个→個 ×20, 会→會 ×18, 无→無 ×17.
**73 of the 325 are byte-identical to their `zh-Hans` value**, i.e. the
conversion never ran on them at all.

Examples, all currently `"state": "translated"`:

| Key | Current `zh-Hant` | Proposed |
|---|---|---|
| `Sources` | 来源 | 來源 |
| `%lld found` | 找到 %lld 个 | 找到 %lld 個 |
| `%lld external candidates skipped` | 跳过 %lld 個外部候选 | 跳過 %lld 個外部候選 |
| `Search credits` | 搜尋积分 | 搜尋積分 |
| `Built-in sources` | 内置源 | 內置源 |

`搜尋积分` and `跳过 %lld 個外部候选` are the clearest tells: Traditional
搜尋/個 sitting directly beside Simplified 积/过/选 inside one short string.
That isn't a regional style choice, it's an incomplete conversion.

To be explicit about what this does **not** touch: `內存` stays `內存` (I'm
fixing 内→內, not proposing 記憶體), `智能體` stays `智能體`. Vocabulary is
group ②/③ territory and yours to call.

## Why it hasn't been caught

`scripts/i18n/check.sh` hardcodes `LOCALES="de,zh-Hans,ko,ru"`, so `zh-Hant`
values are never validated — this is invisible to CI and to review. Probably
worth fixing independently of this PR.

## Reproducing the numbers

```bash
pip install opencc-python-reimplemented
python3 - <<'PY'
import json, opencc
cc = opencc.OpenCC('s2tw')
s = json.load(open('Packages/OsaurusCore/Resources/Localizable.xcstrings'))['strings']
print(sum(1 for e in s.values()
          if (u := e.get('localizations', {}).get('zh-Hant', {}).get('stringUnit'))
          and cc.convert(u['value']) != u['value']))
PY
```

## Notes

- Same class of fix as #1857 (早上好 → 早安), accepted as an enhancement.
- `bash scripts/i18n/check.sh` passes, emitting the same pre-existing stale-key
  warnings (604 + 1) that `main` already produces. The catalog is written back
  with Xcode-exact formatting, so the diff is 325 value lines and nothing else —
  no keys, no `state` fields, no other locale, no placeholder changes.
- If 325 strings is too much for one review, say so and I'll split by area.
- Group ② is a real project-level question rather than a defect — the catalog is
  internally split (爲 156× vs 為 47×, 啓 174× vs 啟 108×). I'd rather open a
  separate issue than pick a side inside this PR.
- Unrelated typo spotted in passing: `Restore an agent bundle exported from this
  or another Mac.` reads `邊原从此 Mac…`, where `邊原` should be `還原`. This PR
  does touch that string (从→從) but deliberately leaves `邊原` alone, since both
  characters are already valid Traditional and fixing it is a wording change
  rather than a character-form one. Fold it in or file separately, your call.

Does this scoping sound useful?

### DIFF
--- a/Packages/OsaurusCore/Resources/Localizable.xcstrings
+++ b/Packages/OsaurusCore/Resources/Localizable.xcstrings
@@ -2505,7 +2505,7 @@
         "zh-Hant" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "%@ / %@ 剩余"
+            "value" : "%@ / %@ 剩餘"
           }
         }
       }
@@ -3589,7 +3589,7 @@
         "zh-Hant" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "%lld 個活跃"
+            "value" : "%lld 個活躍"
           }
         }
       }
@@ -3665,7 +3665,7 @@
         "zh-Hant" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "允许 %1%lld 个 · %2$lld 人"
+            "value" : "允許 %1%lld 個 · %2$lld 人"
           }
         }
       }
@@ -3838,7 +3838,7 @@
         "zh-Hant" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "%lld 个頻道"
+            "value" : "%lld 個頻道"
           }
         }
       }
@@ -4426,7 +4426,7 @@
         "zh-Hant" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "跳过 %lld 個外部候选"
+            "value" : "跳過 %lld 個外部候選"
           }
         }
       }
@@ -4568,7 +4568,7 @@
         "zh-Hant" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "找到 %lld 个"
+            "value" : "找到 %lld 個"
           }
         }
       }
@@ -4754,7 +4754,7 @@
         "zh-Hant" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "%lld 個项目将被編輯"
+            "value" : "%lld 個項目將被編輯"
           }
         }
       }
@@ -5376,7 +5376,7 @@
         "zh-Hant" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "还有 %lld 个匹配 — 继续输入以缩小列表。"
+            "value" : "還有 %lld 個匹配 — 繼續輸入以縮小列表。"
           }
         }
       }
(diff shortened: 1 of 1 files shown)


ftzahao:
I will check it carefully.

RaajeevChandran:
@ftzahao is this safe to merge?

ftzahao:
> @ftzahao is this safe to merge?

No problem. Some traditional Chinese have not been done well. I will check the traditional Chinese carefully.

ftzahao:
@YspritanHyzygy Thank you for pointing out the issue.

@RaajeevChandran This PR looks really great.

YspritanHyzygy:
Status note now that #2178 has landed, to keep the history readable:

- Timeline: this PR was opened at 07:10 UTC with 325 fixed strings. #2178 was opened at 08:32 UTC the same morning and merged at 10:33, without a cross-reference to this PR.
- Outcome: 268 of the 325 keys fixed here now carry byte-identical values on `main`; the other 57 either received a different conversion or still hold the old defective value (e.g. the 泄→洩 leak-class fixes are still missing on `main` today).
- The remaining 57 I'll carry into the phase-2 zh-Hant pass (section 4 of #2189 documents the full scope), in the same consolidated format as #2188.

PR #2201, opened 2026-07-28
Match assistant and user chat text sizing

hkay-dev:
**AI disclaimer: I used GPT-5.6 Sol Extra High in Codex to help investigate this, write the code, and prepare the PR. I reviewed the changes myself, and all functionality testing was done manually.**

I noticed that assistant responses looked quite a bit larger than user messages and the text in the input bar. I'm a bit of a typography snob :) so I started digging through the code and history to see why.

It looks like this came from a few changes that made sense individually but interacted in a way that probably wasn't intended. Chat typography scales with the width it is given, assistant responses get almost the full chat width, and user messages are capped at 65% width. That meant the same theme Body size rendered differently depending on who sent the message.

This PR keeps the existing message widths, but uses the typography system's existing baseline width when rendering chat text. That makes user messages, assistant responses, and the input bar use the same configured Body size. It changes two renderer call sites and adds a focused regression test.

## Testing

I tested the functionality manually in a separate self-signed `Osaurus-dev` build. I tried custom fonts, different Body sizes, live theme changes, and relaunching the app. User text, assistant text, and input text stayed matched. I also added a focused regression test for a narrow user message and a wide assistant response.

## Screenshots

**Before:**
CleanShot 2026-07-27 at 16 13 25

**After:**

 CleanShot 2026-07-27 at 16 26 59

### DIFF
--- a/Packages/OsaurusCore/Tests/Chat/ResponsiveChatTypographyTests.swift
+++ b/Packages/OsaurusCore/Tests/Chat/ResponsiveChatTypographyTests.swift
@@ -1,0 +1,33 @@
+import AppKit
+import Testing
+
+@testable import OsaurusCore
+
+@Suite
+struct ResponsiveChatTypographyTests {
+    @Test @MainActor func chatBodyTypographyMatchesInputSize() throws {
+        let blocks: [SelectableTextBlock] = [.paragraph("Hello")]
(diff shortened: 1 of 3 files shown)


RaajeevChandran:
forensic attention to detail 🫡

PR #199, opened 2025-12-12
added anthropic API support

tpae:
## Summary

for use in Claude Code

## Checklist

- [x] I have read `CONTRIBUTING.md`
- [x] I added/updated tests where reasonable
- [x] I updated docs/README as needed
- [x] I verified build on macOS with Xcode 16.4+

PR #2264, opened 2026-08-02
Pin DSV4 disk-prefix fix and improve eval proof

jjang-ai:
## What changed

- pin vMLX PR #199 at `205720c63a166772564ea662c88f791b40581bad`, which captures DSV4's complete SWA/CSA/HSA prompt-minus-one disk seed during cold prefill so exact replay can restore the longest valid prefix and re-feed only the final prompt token
- preserve DSV4's stable leading system preface while mapping later caller-supplied `system` turns to the model's trained `<｜latest_reminder｜>` role; this keeps prior tokens prefix-stable instead of inserting bare unmarked system text mid-transcript
- make long AgentLoop model steps observable with privacy-clean count-only progress and explicit measured/unavailable decode-throughput attribution
- persist failed/error eval transcripts with trial-distinct forensics
- make the absolute CacheProof resident-budget gate explicitly not applicable when the required resident safetensor bytes already exceed that nominal budget, while retaining the independent growth/leak gate

## Source and focused verification

- Osaurus head: `2bc0ca7a829a3a718dd8816692b901902e794e4d`
- pinned vMLX head: `205720c63a166772564ea662c88f791b40581bad` (#199)
- tested model bundle: `/Users/eric/models/DeepSeek-V4-Flash-0731-JANG`
- `SwiftTransformersTokenizerLoaderTests.dsv4MidConversationSystemUsesLatestReminderWithoutBreakingPrefix`: PASS against that exact local tokenizer; old history remains an exact token prefix, the later reminder renders with `<｜latest_reminder｜>`, and the leading system preface remains at BOS
- core pin/template/progress contract selection: 7/7 PASS
- `AgentLoopCacheProofCampaignTests` + `EvalTranscriptStoreTests`: 23/23 PASS
- resolved-manifest audit: all three tracked Osaurus manifests changed only vMLX `0beef203...` to `205720c63...`
- `git diff --check`: PASS

## Live release gate

Status is **PARTIAL** until the fresh isolated Release app completes the exact DSV4 live rows on this head:

- zero-baseline long disk-L2 store, app restart, exact replay restoring prompt-minus-one with one remaining prefill token
- recorded restored/prefilled token counts, disk hits/misses/stores/bytes, TTFT, and decode token/s
- coherent visible output, terminal Stop disappearance/input unlock, and a follow-up turn
- one reasoning/tool loop showing visible progress while active and a settled tool/reasoning/final lifecycle

No broad model claims or release-ready claim are made before those artifacts and current CI are attached.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "205720c63a166772564ea662c88f791b40581bad"
+        "revision" : "28e71a51b22c2fd51ff01acc3b4d64b44047592c"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "0beef2031e963cfd2ba5ddd362fdd6cb072a809f"
+        "revision" : "205720c63a166772564ea662c88f791b40581bad"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "7a63d960844dc3cb1b145f8150485c73978637f4"
+        "revision" : "6c93611062bd3e42cc7d4b31323467677e44a340"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "28e71a51b22c2fd51ff01acc3b4d64b44047592c"
+        "revision" : "7a63d960844dc3cb1b145f8150485c73978637f4"
       }
     },
     {
--- a/Packages/OsaurusCore/Managers/Model/ModelManager.swift
+++ b/Packages/OsaurusCore/Managers/Model/ModelManager.swift
@@ -1977,6 +1977,15 @@ extension ModelManager {
             localModelsCacheCondition.broadcast()
             localModelsCacheCondition.unlock()
 
+            // A cold scan may outlive discoverLocalModels()'s bounded wait.
+            // In that case launch-time picker builders have already published
+            // their partial snapshot, so merely filling cachedLocalModels is
+            // invisible until some unrelated download/provider event happens.
+            // Publish completion after the cache is ready so every local-model
+            // consumer rebuilds from the finished scan. The next discovery is
+            // cache-only, so this cannot recursively start another scan.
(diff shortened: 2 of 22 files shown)


jjang-ai:
## LIVE VISUAL PROOF — merging on this evidence

**Tested source:** Osaurus `298c4934`, vMLX pin `6c936110` (now on vMLX main via osaurus-ai/vmlx-swift#200)
**App:** Release, binary sha256 `c1424a22d9be244f902b19de396200446cd75179e1ca8b98b44980caa2835b66`, isolated bundle id `com.dinoki.osaurus.dsv4prefillseed20260802`
**Embedded checkout verified:** `SourcePackages/checkouts/vmlx-swift` HEAD = `6c93611062bd3e42cc7d4b31323467677e44a340`
**Bundle:** `~/models/DeepSeek-V4-Flash-0731-JANG`; fresh isolated root, empty cache; paged RAM cache OFF, disk L2/prefix ON; Sandbox + folder tools enabled

All rows were driven through the real chat UI over Accessibility and observed
with window captures. No CLI or API request was used as proof.

### What this PR carries

1. **Repin to the DSV4 cache fixes** — #199 (warm-up seed persisted, so the
   visible send stops re-prefilling the warmed prefix) and #200 (history cache
   boundary survives an assistant continuation turn).
2. **DSV4 pre-send warm-up no longer re-prefills an already-warm prompt.**
   `requireDSV4PreSendWarmupIfNeeded` routed Send through
   `handleContextShapeChange`, which unconditionally calls
   `invalidateWarmState()` and clears `warmedFingerprint`. That path is for a
   real prompt-shape edit; Send does not change the shape, and
   `selectedModelResident` lags the runtime by a residency snapshot, so the
   guard fires routinely on a prompt that is already warm. Clearing the
   fingerprint defeated `performWarmup`'s identical-payload check and prefilled
   the same tokens twice.
   Sabotage-verified: with `invalidatingShape: true` the regression test records
   `engine.requestCount == 2`; with the fix it is `1`.
3. **Repin completeness.** `test-core` was red because the vMLX pin lives in
   SEVEN places and the previous bump updated four.
   `App/osaurus.xcodeproj/.../Package.resolved` and both tripwire tests still
   asserted `7a63d960`, so `ImageGenerationBridgeContractTests` failed on
   `packageSwift`/`packageResolved`/`workspaceResolved`. All seven now agree.

### Live rows

**Duplicate prefill — fixed.**
```
[vmlx][cache/fetch] MISS all tiers tokens=1129                      <- warm-up (cold, stores seed)
[vmlx][cache/fetch] HIT disk boundary=1127 remaining=96 tokens=1223 <- visible send restores it
```
Previously the visible send logged `MISS all tiers` and prefilled the prefix again.
Turn: "Thought for 3.6s" (reasoning opened and closed, nothing leaked inline),
coherent answer, `TTFT 1.73s · 31.3 tok/s · 144 tokens`, Stop cleared, input unlocked.

**Tool call — full lifecycle.** "Thought for 4.5s" → tool card **"Wrote a file ·
2.7s"** started and finished → `dsv4proof.txt +1` diff card → coherent final
answer → Stop cleared, input unlocked. Boundaries `all=[208, 1830, 2391]`,
fetch `HIT 2286 remaining=107`, finalized 41s.

**Follow-up after the tool turn.** Coherent ~300-word answer, no protocol debris,
no repetition, no truncation. `TTFT 1.91s · 23.3 tok/s · 681 tokens`, fetch
`HIT 2756 remaining=110`, stores advancing 2865 → 3547, Stop cleared.

**Model loads once.** Each `[Load]` line appears exactly once; `footprint` 96–101 GB,
single mapping. The doubled `[vmlx][cache/boundaries]` lines are the two probe
renders in `MLXBatchAdapter.warmupSendInvariantPrefixTokens`, not a second prefill.

### Throughput

The 4.7–5.3 tok/s rows are short-generation artifacts (18 and 108 tokens) where
fixed per-turn cost dominates the honest denominator. A real 681-token
generation measures **23.3 tok/s**, consistent with the 21.0 tok/s native rate
recorded in vmlx `ad059136`. Compiled decode is unchanged and must stay off for
DSV4: that commit records an exact-bundle A/B of 21.0 native vs 12.7 with the
generic compiled request.

### CI

All 8 checks green on `298c4934`: test-core, test-cli, test-packages,
test-statspack, test-evals, swiftlint, shellcheck, update_release_draft.

### Known open items, tracked NOT fixed here

- Post-turn background re-warm can still miss with signature `stored+2`
  (2084→2086, 2286→2288, 2756→2758). Fixed in osaurus-ai/vmlx-swift#201, which is
  deliberately unmerged until it has its own live proof. Visible turns hit.
- End-of-turn finalization: `BatchEngine.swift:3264-3296` runs the cache store
  and `Stream().synchronize()` before `continuation.finish()`. Measured 20.2s and
  13.9s unaccounted between last token and Stop clearing, with 11 blobs / 462 MB
  written per 4-turn session.

PR #2267, opened 2026-08-02
fix(make): derive EVALS_DETERMINISTIC_SUITES from Config/floors.json

### LINKED ISSUE #2266: Make evals-deterministic derive from Config/floors.json to prevent drift between CI lane and floor gate
## Problem

`Makefile`'s `EVALS_DETERMINISTIC_SUITES` (line 317) is a hand-maintained duplicate of the `suitePassRates` keys in `Packages/OsaurusEvals/Config/floors.json` (lines 3-12). Both currently list the same eight token-free deterministic suites:

- `Schema`
- `ToolEnvelope`
- `PrefixHash`
- `ArgumentCoercion`
- `ToolResultGrounding`
- `AgentChannels`
- `ComputerUse`
- `ScreenContext`

Today they happen to match, but nothing in the repo enforces that. Adding a new deterministic suite requires editing two files in two places. A contributor who adds an entry to `floors.json` and forgets the Makefile silently ships a thinner `evals-deterministic` lane in CI than the floor gate enforces, and the regression signal is lost.

## Current behavior

1. `make evals-deterministic` iterates only over the hardcoded `EVALS_DETERMINISTIC_SUITES` list.
2. The OsaurusEvals harness's `--fail-on-floor` flag uses the `suitePassRates` keys from `floors.json`.
3. The two lists can drift. A new deterministic suite added to `floors.json` is not picked up by the Makefile lane until a maintainer also edits line 317.

## Expected behavior

A single source of truth: `Config/floors.json#suitePassRates`. Editing that file is the only place a maintainer touches to add or remove a deterministic suite, and `make evals-deterministic` immediately reflects the change on the next build.

## Proposed solution

1. Replace the hardcoded `EVALS_DETERMINISTIC_SUITES` list in the Makefile with a `jq` derivation from `Packages/OsaurusEvals/Config/floors.json` so the two never drift.
2. Add `scripts/live-proof/assert-eval-floors-makefile-sync.sh` to flag the moment a contributor hand-edits the Makefile, breaks the contract, or ships a `floors.json` key without a matching `Suites/<name>/` directory.
3. Document the contract in the existing comment block above `EVALS_DETERMINISTIC_SUITES`.

## Scope

- `Makefile` (one variable, one comment block update)
- `scripts/live-proof/assert-eval-floors-makefile-sync.sh` (new file, follows the same shape as the other `assert-*.sh` guards in the directory)

No runtime code touched. No eval fixture touched. No CI wiring change (the existing `assert-*.sh` pattern is local-pre-PR, not CI).

## Acceptance criteria

- `make evals-deterministic` runs exactly the suites listed in `Config/floors.json#suitePassRates`, in the order they appear in the JSON.
- Adding a new key to `suitePassRates` automatically includes it in the next `make evals-deterministic` run with no second edit.
- `bash scripts/live-proof/assert-eval-floors-makefile-sync.sh` exits 0 on a clean tree, exits non-zero on drift.
- The new script lints clean under `shellcheck --severity=warning` (matches the existing CI shell-lint step).
- Existing `make evals-deterministic` runs on the eight current suites produce the same set as before this change.

## Backward compatibility

`evals-deterministic` is a CI lane. The only behavioral change is that adding a new deterministic suite to `floors.json` is now a one-edit change. A maintainer who today hand-edits the Makefile to add a new suite will need to add the suite to `floors.json` instead — that is the intended migration.

## Risks

- **`jq` missing on a contributor's machine.** `$(shell ...)` would expand to an empty string and the deterministic lane would silently no-op. Mitigation: the new assert script runs `make -n` and `make -p` to verify the resolved list is non-empty, and the loop in `evals-deterministic` already echoes the suite name before running it, so a zero-iteration loop is visibly wrong. `jq` is on the default macOS image used by GitHub Actions and is already a transitive dep of the OsaurusCore build graph (Aptabase pulls it in), so this is a contributor-side risk, not a CI risk.
- **`floors.json` malformed.** Same mitigation: the assert script catches malformed JSON at parse time.

Harsh23Kashyap:
## Summary

Fixes a drift hazard between `Makefile`'s `EVALS_DETERMINISTIC_SUITES` and the `suitePassRates` keys in `Packages/OsaurusEvals/Config/floors.json`. Both listed the same eight token-free deterministic suites; the two could drift because nothing enforced the contract.

Closes #2266.

## Changes

- `Makefile` derives `EVALS_DETERMINISTIC_SUITES` from `Config/floors.json#suitePassRates` with `jq`. Adding a new deterministic suite to `floors.json` is now a one-file edit. The `jq` call returns empty if the file is missing or malformed, which the new assert script catches.
- `scripts/live-proof/assert-eval-floors-makefile-sync.sh` (new): three checks
  1. `Config/floors.json` parses and has a non-empty `suitePassRates`
  2. `make -n -p` resolves `EVALS_DETERMINISTIC_SUITES` to a non-empty value
  3. The two lists are equal AND every key has a matching `Suites/<name>/` directory
- Catches a contributor who adds a key to `floors.json` but forgets the `Suites/<name>/` directory (the Makefile lane would then fail mid-loop with a confusing "no such path" error instead of a clear drift report here), a contributor who hand-edits the Makefile list back to a hand-rolled set, or a broken `jq` derivation.

## Test Plan

Verified locally on the change set:

```
$ bash scripts/live-proof/assert-eval-floors-makefile-sync.sh
PASS Config/floors.json suitePassRates has 8 keys
PASS Makefile resolved EVALS_DETERMINISTIC_SUITES = AgentChannels ArgumentCoercion ComputerUse PrefixHash Schema ScreenContext ToolEnvelope ToolResultGrounding
PASS every suitePassRates key has a matching Suites/<name>/ directory
exit=0
```

Verified the failure path: a temp `DoesNotExist` key in `suitePassRates` (with no matching directory) makes the assert fail with the right message:

```
FAIL floors.json suitePassRates key 'DoesNotExist' has no matching .../Suites/DoesNotExist directory
exit=1
```

Verified `make -n evals-deterministic` resolves to the same eight suites as before the change (alphabetical sort order via `jq` is stable across `apt`/Homebrew `jq` versions).

Existing `scripts/live-proof/assert-*.sh` scripts were unaffected: ran `assert-osaurus-pr-hygiene.sh`, `assert-keychain-disabled-source-coverage.sh`, and `assert-no-hidden-local-sampler-defaults.sh` against the change; all three pass.

The full `make evals-deterministic` lane runs `swift run` per suite (~minutes per suite on a cold build). I did not run the full lane locally because:
- The negative test on the Makefile variable + assert script already proves the contract
- A full run would consume ~30 minutes of CI-equivalent time and add no new evidence
- The CI lane will exercise the full path on this PR

## Design Decisions

- **jq is already a transitive dep** of the OsaurusCore build graph (Aptabase pulls it in), so this is a zero-new-dependency change for the project. It's also on the default GitHub Actions `macos-26` image.
- **Temp file + awk** for the `make -n -p` read instead of a pipe. The pipe would SIGPIPE on a multi-MB make database under `set -o pipefail` (awk exits after the first match → make closes the pipe → the whole substitution fails). Writing to `mktemp` first sidesteps that.
- **No CI wiring.** The new script follows the existing `scripts/live-proof/assert-*.sh` pattern: a pre-PR local guard, not a CI lane. The CI `Lint shell scripts` step will still pick it up for shellcheck.
- **No new eval lane.** The change is plumbing; the existing `make evals-deterministic` target is the same shape as before, just with the variable derived instead of hand-rolled.

## Backward Compatibility

A maintainer who today hand-edits the Makefile to add a deterministic suite will need to add the suite to `floors.json` instead. That is the intended migration: `Config/floors.json` becomes the single source of truth for both the CI lane and the eval floor gate.

## Risks

- **`jq` missing on a contributor's machine.** The Makefile's `$(shell ...)` would expand to an empty string and `evals-deterministic` would silently no-op. The new assert script catches this at PR-prep time. `jq` is on the default macOS image used by GitHub Actions and is a transitive dep of the build graph, so this is a contributor-side risk, not a CI risk.
- **`Config/floors.json` malformed.** Same mitigation. The assert script catches JSON parse failures with a clear error message.

## Future Work

None required for this change. The next deterministic suite a maintainer adds will flow into `evals-deterministic` automatically.

## Checklist

- [x] I added/updated tests where reasonable (`assert-eval-floors-makefile-sync.sh` follows the existing `assert-*.sh` pattern in the same directory)
- [x] I updated docs/README as needed (the Makefile comment block above `EVALS_DETERMINISTIC_SUITES` was updated; no README changes needed)
- [x] I verified build on macOS with Xcode 16.4+ (the change is Makefile + shell only; no Swift touched)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -346,8 +346,14 @@ evals-all-report: evals-prep
 # regression (floors.json pins their pass rate at 1.0). Safe on hosted CI
 # runners; the CI `test-evals` job runs this after the harness unit tests.
 # No `evals-prep` dependency: these lanes never touch MLX or the embedder.
-EVALS_DETERMINISTIC_SUITES := Schema ToolEnvelope PrefixHash ArgumentCoercion \
-	ToolResultGrounding AgentChannels ComputerUse ScreenContext
+#
+# Derived from Packages/OsaurusEvals/Config/floors.json (the suitePassRates
+# keys) so adding a new deterministic suite to floors.json is a one-file
+# edit. `scripts/live-proof/assert-eval-floors-makefile-sync.sh` flags
+# drift if a contributor hand-edits this list or if the file is missing.
+EVALS_DETERMINISTIC_SUITES := $(shell \
+	jq -r '.suitePassRates | keys | join(" ")' \
+	Packages/OsaurusEvals/Config/floors.json 2>/dev/null)
 
 evals-deterministic:
 	@rc=0; for name in $(EVALS_DETERMINISTIC_SUITES); do \
--- a/scripts/live-proof/assert-eval-floors-makefile-sync.sh
+++ b/scripts/live-proof/assert-eval-floors-makefile-sync.sh
@@ -1,0 +1,73 @@
+#!/usr/bin/env bash
+# Verifies the contract between Makefile's EVALS_DETERMINISTIC_SUITES and
+# Packages/OsaurusEvals/Config/floors.json#suitePassRates. Both list the
+# same token-free deterministic suites; this guard catches a contributor
+# who adds a key to floors.json but forgets a Suites/<name>/ directory,
+# who hand-edits the Makefile list, or who breaks the jq derivation.
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+fail=0
+
+pass() { echo "PASS $1"; }
+fail_msg() { echo "FAIL $1" >&2; fail=1; }
+
+FLOORS="$ROOT/Packages/OsaurusEvals/Config/floors.json"
+SUITES_DIR="$ROOT/Packages/OsaurusEvals/Suites"
+
+if [[ ! -f "$FLOORS" ]]; then
+  fail_msg "Config/floors.json missing at $FLOORS"
+  exit "$fail"
+fi
+
+if ! command -v jq >/dev/null 2>&1; then
+  fail_msg "jq not on PATH (Makefile derivation needs it)"
+  exit "$fail"
+fi
+
+# Read the JSON keys. Tr/normalize to a single space-separated line.
+json_keys="$(jq -r '.suitePassRates | keys[]' "$FLOORS" 2>/dev/null | tr '\n' ' ' | sed 's/ $//')"
+if [[ -z "$json_keys" ]]; then
+  fail_msg "Config/floors.json suitePassRates has no keys (or jq parse failed)"
+  exit "$fail"
+fi
+pass "Config/floors.json suitePassRates has $(echo "$json_keys" | wc -w | tr -d ' ') keys"
+
+# Read the Makefile's resolved value. `make -p` prints the database; filter
+# for the exact variable and take the right-hand side. The pipe to awk
+# would SIGPIPE on a multi-MB database (awk exits after the first match);
+# write the db to a temp file first and read it from there.
+make_db="$(mktemp)"
+trap 'rm -f "$make_db"' EXIT
+make -n -p 2>/dev/null > "$make_db" || true
+make_keys="$(awk -F' := ' '
+  /^EVALS_DETERMINISTIC_SUITES := / { sub(/\r$/, ""); print $2; exit }
+' "$make_db")"
+if [[ -z "$make_keys" ]]; then
+  fail_msg "Makefile did not resolve EVALS_DETERMINISTIC_SUITES (jq derivation produced empty value?)"
+  exit "$fail"
+fi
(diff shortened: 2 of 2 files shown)


Harsh23Kashyap:
PR is 11 days old with no review yet. Rebased onto current main (the branch had fallen 74 commits behind — none of those touched Makefile, Config/floors.json, or scripts/live-proof/, so the rebase was clean). The 8 deterministic eval suites still resolve via the jq derivation; the new assert-eval-floors-makefile-sync.sh passes 3/3 and catches both the drift case and the missing-directory case. If anything needs changing, let me know.

PR #2270, opened 2026-08-02
Repin vMLX to the boundary-digest diagnostic

jjang-ai:
Repins vMLX to `087e43ef` (osaurus-ai/vmlx-swift#204, merged to vMLX main as
`e16738c5`).

Diagnostic only — the vMLX change adds `head=` and `stable0=` prefix digests to
the `[vmlx][cache/boundaries]` trace behind the existing `VMLX_CACHE_FETCH_TRACE`
flag. No runtime behaviour changes, nothing new on the production path.

## Why

The `stored+2` re-warm miss has now produced two wrong root causes because the
trace could not distinguish two prompts of equal length. The digest makes that
distinguishable.

## What the first capture showed

App built from this exact pin, same isolated root, real chat UI:

```
[vmlx][cache/boundaries] prompt=1141 all=[72,1128,1139] head=22bca74a stable0=e7fcebd8
[vmlx][cache/fetch] HIT disk boundary=1127 remaining=12  tokens=1139
[vmlx][cache/boundaries] prompt=1221 all=[72,1128,1219] head=22bca74a stable0=e7fcebd8
[vmlx][cache/fetch] HIT disk boundary=1127 remaining=94  tokens=1221
[vmlx][cache/disk-store] count=1333
[vmlx][cache/boundaries] prompt=1338 all=[72,1128,1336] head=22bca74a stable0=e7fcebd8
[vmlx][cache/fetch] MISS all tiers tokens=1335
```

Digests identical on every line including the failing re-warm, so the system
region is byte-stable and the divergence is inside the conversation-history
span. That span is exactly what `head=` (first 256 tokens) and `stable0=`
(first 72) do not cover, so a third digest at the history boundary is the
follow-up.

Turn itself was clean: finalized in 37s, Stop cleared, input unlocked, two
cache HITs (`remaining=12`, `remaining=94`).

## Pin completeness

All seven pin sites updated together, including
`App/osaurus.xcodeproj/.../Package.resolved` and both tripwire tests.


jjang-ai:
## `test-core` failure is a pre-existing test race, not this change

`test-core` failed once on this PR in `SpawnBatchToolTests`:

```
Expectation failed: results.allSatisfy { $0.envelope.contains(#""ok":false"#) }
Suite SpawnBatchToolTests failed after 3.024 seconds with 3 issue(s)
```

### Why it cannot be this PR

The entire diff is **6 files, one line each, all SHA strings** in pin files —
`Package.swift`, three `Package.resolved`, and the two pin tripwire tests.
Nothing on the spawn/batch path is touched, and the vMLX commit being pinned
(`087e43ef`) only adds a field to a trace line gated behind
`VMLX_CACHE_FETCH_TRACE`, which CI does not set.

`main` was green on its three preceding CI runs with zero `SpawnBatchToolTests`
mentions.

### The actual race

`Packages/OsaurusCore/Tests/Subagent/SpawnBatchToolTests.swift:2388` uses a
fixed sleep as a synchronisation barrier:

```swift
let task = Task { await SpawnBatchTool.runPreparedJobs(jobs, maxParallel: 2, ...) }

try? await Task.sleep(for: .milliseconds(20))
interrupt.interrupt()
let results = await task.value...

#expect(results.allSatisfy { $0.envelope.contains(#""ok":false"#) })
#expect(results.allSatisfy { ...["kind"] as? String == "user_denied" })
```

If the two jobs complete inside that 20 ms window on a loaded runner, the
interrupt arrives after they have already succeeded and the envelopes are
`ok:true`, failing all three assertions together — which is exactly the observed
signature.

The same file already has the right idiom at `:1164-1168` (a 2-second deadline
with a 1 ms poll), so the fix is to wait on the jobs having actually started
rather than on a fixed duration. Tracked separately; not folded into this repin
so the diff stays a pure pin bump.

Re-running `test-core` to confirm.

PR #2271, opened 2026-08-02
Stop cancellationDuringWave racing a fixed sleep against job completion

jjang-ai:
`SpawnBatchToolTests.cancellationDuringWave` is timing-dependent and has been
failing `test-core` on CI.

## The race

The test starts the batch in a `Task`, sleeps a fixed 20 ms, then interrupts and
asserts every result came back `"ok":false` with `kind == "user_denied"`:

```swift
let task = Task { await SpawnBatchTool.runPreparedJobs(jobs, maxParallel: 2, ...) }

try? await Task.sleep(for: .milliseconds(20))
interrupt.interrupt()
...
#expect(results.allSatisfy { $0.envelope.contains(#""ok":false"#) })
```

The scripted jobs sleep 80 ms (`prepared`'s `delayMilliseconds` default), so the
barrier leaves only a **60 ms margin**. On a contended runner
`Task.sleep(20 ms)` oversleeps, both jobs complete before `interrupt()` lands,
and all three assertions invert to `ok:true`.

## Evidence it is the race and not a product change

- Failed twice in a row on osaurus#2270, whose **entire diff is six one-line SHA
  strings** in pin files. The vMLX commit being pinned adds one field to a trace
  line gated behind `VMLX_CACHE_FETCH_TRACE`, which CI does not set — verified by
  diffing the two vMLX revisions directly.
- The issue count **varied 3 → 4** between the two runs. A deterministic code
  fault does not change how many assertions it trips.
- The same test passes **6/6 locally** on a fast machine, where 20 ms comfortably
  precedes the 80 ms job.
- `main` was green on its three preceding CI runs.

## The fix

Wait for both jobs to actually **enter execution** via the existing
`BatchExecutionProbe`, using the deadline+poll idiom already used elsewhere in
this file (`:1164-1168`), instead of assuming a fixed duration:

```swift
let startDeadline = Date().addingTimeInterval(2)
while await probe.snapshot().total < jobs.count, Date() < startDeadline {
    try? await Task.sleep(for: .milliseconds(1))
}
interrupt.interrupt()
```

and widen those two jobs to 400 ms so the post-start window is not marginal
either. The test's intent — interrupt lands mid-wave — is now expressed as a
condition on observable state rather than a duration.

## Validation

`swift test --filter SpawnBatchToolTests` → **38/38 passed**.

Kept as its own PR rather than folded into the pin bump in #2270, so that repin
stays a pure one-line-per-file change.

### DIFF
--- a/Packages/OsaurusCore/Tests/Subagent/SpawnBatchToolTests.swift
+++ b/Packages/OsaurusCore/Tests/Subagent/SpawnBatchToolTests.swift
@@ -2356,7 +2356,8 @@ struct SpawnBatchToolTests {
                 model: "cloud/a",
                 isLocal: false,
                 admission: .remote,
-                probe: probe
+                probe: probe,
+                delayMilliseconds: 400
             ),
             prepared(
                 index: 1,
@@ -2366,7 +2367,8 @@ struct SpawnBatchToolTests {
                 model: "cloud/b",
                 isLocal: false,
                 admission: .remote,
-                probe: probe
+                probe: probe,
+                delayMilliseconds: 400
             ),
         ]
         let interrupt = InterruptToken()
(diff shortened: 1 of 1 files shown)


jjang-ai:
## CI on this PR: the target test passed, a *different* load-sensitive test failed

First CI run on this branch: `SpawnBatchToolTests` is **absent from the failure
list**, so the fix did what it was meant to. `test-core` still went red, but on
an unrelated suite:

```
Recorded an issue (Expectation failed: (completionNotifications.count → 0) == 1)
Suite ModelManagerTests failed after 10.604 seconds with 1 issue(s)
```

That is the same class of defect, not a consequence of this change.
`ModelManagerTests.swift:512` asserts a **NotificationCenter** delivery count
straight after a poll loop built on `try await Task.sleep(nanoseconds: 25_000_000)`:

```swift
try await Task.sleep(nanoseconds: 25_000_000)
...
#expect(completionNotifications.count == 1)
```

Notification delivery is asynchronous, so under runner load the assertion runs
before the observer fires and the count is 0 — a fixed-duration barrier standing
in for a condition, exactly like the 20 ms sleep this PR removes.

That makes three distinct load-sensitive tests observed on these runners in one
session (`SpawnBatchToolTests`, now `ModelManagerTests`, and the earlier
`ImageGenerationBridgeContractTests` pin mismatch which was a genuine bug and is
fixed). Tracked separately rather than bundled here; this PR stays scoped to the
one race it fixes.

Re-running `test-core` to confirm this branch is otherwise clean.

PR #2273, opened 2026-08-02
Report why the disk cache was disabled instead of silently downgrading

jjang-ai:
## The failure this makes visible

DSV4 agent loops hung on every tool call, re-prefilling the entire growing
transcript each round.

Root cause: `~/.osaurus/cache` had been created **root-owned** (Jul 29). The app
runs as `eric`, so every probe write failed with EACCES, `diskDirUsable` went
false, and `buildCacheCoordinatorConfig` forced `enableDiskCache = false`.

DSV4 marks itself paged-incompatible (`CacheCoordinator.setPagedIncompatible`,
"so the disk tier is the only prefix-reuse mechanism — which is correct for
DSV4"), and `pagedKV.enabled=false` in the runtime config. With the disk tier
off, DSV4 had **no prefix cache at all**.

## Live evidence

Same build, same `VMLX_CACHE_FETCH_TRACE=1`, before the directory was repaired:

```
boundaries computed : 12
fetches             :  9   ->  HITs: 0   MISSes: 9
disk-store          :  0
KV entries on disk  :  0
```

The tool turn published `all=[193, 2442]` — history boundary gone, where every
non-tool turn published `all=[193, 2442, 2591]` — fell back to a stale 2592
store and cold-prefilled `remaining=3977`. The UI rendered that as
`Prefill 4128/6569`.

After `chown`ing the directory back, with no code change:

```
disk-store          : 10   (keys __dsv4_N_meta__ / __dsv4_N_nilmask__)
fetch               : HIT disk boundary=2591 remaining=186
                      HIT disk boundary=2441 remaining=151
```

`remaining` 3977 → 151, and the tool turn's history boundary reappeared
(`prompt=2592 all=[193, 2442, 2590]`).

## Why it went undiagnosed

`isDirectoryWritable` swallowed the probe error (`catch { return false }`) and
the caller logged a `.warning` naming only the path — no errno, no ownership,
and at a level nobody reads. A total prefix-cache outage looked exactly like
"this model is just slow".

## What this changes

`diskCacheProbeFailure` returns the reason instead of a bare `Bool`: the
underlying error plus the directory's owner and mode against the current user.
The caller logs it at `.error`, because losing the only prefix-reuse tier is
not a tuning detail.

`isDirectoryWritable` is kept as a thin wrapper, so **no behaviour changes** —
the same condition still disables the disk tier. This only makes the cause
sayable.

Example of what would have been logged:

```
buildCacheCoordinatorConfig: disk cache dir NOT WRITABLE — prefix caching is
disabled and every request will re-prefill from scratch.
path=/Users/eric/.osaurus/cache
reason=You don't have permission to save the file (owner=root mode=755, current user=eric)
```

## Validation

`xcodebuild -scheme osaurus-cli -configuration Release` → clean.

Not yet proven live that the new `.error` line renders for a fresh failure —
the condition was repaired to unblock the machine. Reproducing it means
deliberately breaking the cache directory again; happy to do that before merge
if you want the log line itself verified.

## Follow-up worth considering

Self-heal or surface in the UI. A silent, unbounded slowdown with no
user-visible signal is the worst shape this failure can take, and an `.error`
in os_log is still only findable by someone who already suspects the cache.

### DIFF
--- a/Packages/OsaurusCore/Resources/Localizable.xcstrings
+++ b/Packages/OsaurusCore/Resources/Localizable.xcstrings
@@ -142531,6 +142531,40 @@
         }
       }
     },
+    "Preparing tool call (%@)" : {
+      "localizations" : {
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Werkzeugaufruf wird vorbereitet (%@)"
+          }
+        },
+        "ko" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "도구 호출 준비 중 (%@)"
+          }
+        },
+        "ru" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Подготовка вызова инструмента (%@)"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "正在准备工具调用 (%@)"
+          }
+        },
+        "zh-Hant" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "正在準備工具調用 (%@)"
+          }
+        }
+      }
+    },
     "Preparing…" : {
       "localizations" : {
         "de" : {
--- a/Packages/OsaurusCore/Services/ModelRuntime.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime.swift
@@ -4056,10 +4056,20 @@ public actor ModelRuntime {
         if let diskCacheDir {
             OsaurusPaths.ensureExistsSilent(diskCacheDir)
         }
-        let diskDirUsable = diskCacheDir.map(isDirectoryWritable) ?? false
-        if let diskCacheDir, !diskDirUsable {
-            genLog.warning(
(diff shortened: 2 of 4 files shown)


jjang-ai:
## Live visual proof

Built and relaunched the dev app, loaded `deepseek-v4-flash-0731-jang` (99.8 / 128 GB resident), and asked it to write a Minesweeper HTML file.

The pending tool card now renders, verbatim from the running app:

```
Preparing tool call (3.2 KB)
```

Before this change the same phase showed a static `Preparing tool call` for the entire multi-minute envelope write, which is what read as a hang — the process was at 120–138% CPU decoding the whole time.

The disk-L2 half is proven on the same machine: after repairing the root-owned `~/.osaurus/cache` the trace went from 0 stores / 0 hits to real stores, e.g.

```
[vmlx][cache/disk-store] count=1426 keys=["__jang_cache_format_version__", "__layer_kind_0__", ...]
[vmlx][cache/fetch] HIT disk boundary=2591 remaining=186
```

with the DSV4 tool turn's `remaining` down from 3977 to 151.

Still not verified live: the new `.error` line for an unwritable cache dir, since the condition was repaired to unblock the machine. Reproducing it means deliberately breaking the directory again — happy to do that if you want the log line itself confirmed before merge.

PR #2276, opened 2026-08-03
Repin vMLX to the disk-store offset guard

jjang-ai:
Brings in vmlx-swift `8e2c3d6e` (main: #206 encoder-contract tests, #207 partial-restore equivalence proof, #208 the disk-store offset guard).

## Why

vmlx-swift#208 root-causes and fixes the DSV4 agent-loop reasoning degeneration: the disk tier persisted boundary snapshots whose rotating offset disagreed with the token-count key (live: `tokens=1831 rotatingOffsets=[1832]` — the paged tier refused that exact snapshot, the disk tier had no guard). Restoring such an entry re-feeds the boundary token, shifts every later position by one, and compounds each tool round → fluent, non-converging reasoning loops, only with disk-L2 on.

## Live proof (before this repin, on the affected machine)

Guarded build, shipped defaults (prefix ON / SSD disk ON / paged OFF), multi-turn tool chain to 9k+ tokens:

```
disk-store count=2325 / 2509 / 5791
HIT boundary=2325 remaining=185
HIT boundary=2509 remaining=3283
HIT boundary=5791 remaining=3366
```

Reasoning terminates normally ("Thought for 4m 45s" → action, 14.7 tok/s), no loops — user-confirmed on the same workload that previously degenerated.

## Scope

Pin strings only: 6 files, 6 lines (Package.swift, three Package.resolved, two pin-tripwire tests). **No settings, defaults, or toggles change.**

## Deployment note

Machines that ran disk-L2 before the guard should purge the KV cache dir once — pre-guard entries may be internally inconsistent and the guard only prevents new poison.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "087e43ef49ac2ddc6dbe2f7b4d7fb4d755896fa3"
+        "revision" : "8e2c3d6ebca6a43be6a516eb3dc55ef49c174a5d"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "087e43ef49ac2ddc6dbe2f7b4d7fb4d755896fa3"
+        "revision" : "8e2c3d6ebca6a43be6a516eb3dc55ef49c174a5d"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.swift
+++ b/Packages/OsaurusCore/Package.swift
@@ -138,7 +138,7 @@ let package = Package(
         // group normalization without changing the shared unload/cache APIs.
         .package(
             url: "https://github.com/osaurus-ai/vmlx-swift",
-            revision: "087e43ef49ac2ddc6dbe2f7b4d7fb4d755896fa3"
+            revision: "8e2c3d6ebca6a43be6a516eb3dc55ef49c174a5d"
         ),
         // FluidAudio 0.14.3 added a breaking `language:` parameter to TTS
         // calls that osaurus's `TTSService` doesn't pass. Pinning to the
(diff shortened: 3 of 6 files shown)


jjang-ai:
## Complete investigation record (2026-08-02)

This repin is the final step of a day-long root-cause chain. For future readers hitting "DSV4 reasoning loops" or "agent chat re-prefills every turn":

### Symptom timeline
1. DSV4 agent chains stalled on every tool round ("Preparing tool call" for minutes, full-transcript re-prefills, `Prefill 4128/6569`).
2. After fixing that, long chains degenerated into fluent, non-converging reasoning loops at ~9–12k tokens.

### Root causes found (three, independent)
1. **`~/.osaurus/cache` root-owned** → probe-write failed → osaurus silently forced memory-only → DSV4 (paged-incompatible) had **zero** prefix reuse → every tool round cold-prefilled. Fixed operationally (`chown`); #2273 makes the failure report its cause at `.error` instead of a silent downgrade.
2. **Agent stored `temperature: 0`** → per-request precedence forced argmax over the bundle's `do_sample: true, temp 1.0, top_p 0.95` → byte-verbatim repetition loops. Fixed by clearing the override; #2274 pins the trap in tests.
3. **The main event — disk-store offset poison** (vmlx-swift#208): the SSD tier persisted boundary snapshots whose rotating offset disagreed with the token-count key (`tokens=1831 rotatingOffsets=[1832]` live; the paged tier refused that same snapshot via `expectedOffset`, the disk tier had no guard). Restore re-feeds the boundary token → all later positions shift by one → fluent non-converging reasoning, compounding each round as new N-1 seeds are captured from corrupted state. Isolated by live A/B (RAM-prefix-only never loops; disk-L2 on loops; all-off never loops), reproduced model-free in a unit test, fixed by refusing the store.

### Hypotheses tested and rejected on the way (so nobody re-chases them)
- raw KV/partial-restore fidelity — `restore(B)+feed(N−B) == feed(N)` proven incl. ring-wrap boundaries and N-1 seeds (vmlx#207)
- DSV4 pool/buffer loss on restore — survives under realistic state
- DSML parser — tool chains parsed correctly throughout
- tool-call re-render ordering — real but latent (`rawArgumentsJSON` preserved); recorded as a known issue in vmlx#206
- the `stored+2` "anomaly" chased across vmlx#202–#205 — structural: 20/20 live boundary lines sit at exactly `prompt − 2`, the `<|Assistant|><think>` rail

### Live proof of this pin's payload
Guarded build, shipped defaults (prefix ON / SSD ON / paged OFF), multi-turn tool chain to 9k+ tokens: stores + HITs every round (`HIT boundary=5791 remaining=3366` vs 9k cold), reasoning terminating normally, no loops — user-confirmed on the previously-failing workload.

### Ops note
Purge the KV cache dir once when deploying this (pre-guard entries may be internally inconsistent; the guard blocks new poison, not old). Nothing in this PR or the pinned vmlx range touches settings, defaults, or toggles — 6 pin lines, and the guard itself only rejects stores that are invalid for any model.

PR #2319, opened 2026-08-08
Isolate test hosts from persistent data and Keychain

mimeding:
## Summary

- centralize test-host recognition and route recognized XCTest and SwiftPM processes to private per-process data roots instead of persistent Osaurus storage
- disable production Keychain access in isolated test hosts, while retaining a narrowly scoped UUID-namespaced Master Key proof interface
- propagate fail-closed isolation through model startup, MCP, plugins, Seatbelt, Knowledge Git, warmup, downloads, and other background work
- sanitize isolated child-process environments, including `HOME`, executable search paths, credential variables, and reserved isolation markers
- make legacy-root migration, external-model registry updates, storage-key caching, and asynchronous startup transactional across policy changes
- replace process-global test seams with task-scoped or lock-backed state where concurrent suites previously interfered with one another

## Why

The test matrix previously depended on launcher environment variables being forwarded consistently. App-hosted XCTest, SwiftPM helpers, and child processes could therefore resolve persistent `.osaurus` data or reach the login Keychain when a marker was absent, malformed, or discovered after early startup. Shared overrides and asynchronous model, plugin, MCP, migration, and Keychain work also left order-dependent transition windows.

This change establishes one runtime policy for data-root and Keychain isolation, then revalidates that policy immediately before persistent commits or child-process launch. Work admitted before a late transition either rolls back, retries in the isolated namespace, or fails with a typed error.

## Production behavior

For a normal Osaurus launch without test markers or isolation settings:

- the data root remains `~/.osaurus`
- production Keychain services remain available
- model discovery and startup remain enabled
- MCP, plugin, and Git children preserve the normal production environment
- legacy data migration remains available

The isolation path is fail-closed only for recognized or pending test hosts.

## Security and integrity properties

- ordinary isolated test hosts cannot read, write, enumerate, or delete production Keychain services through Osaurus wrappers
- storage-key and salt writes fail with a typed `errSecInteractionNotAllowed` error if isolation begins after work was admitted
- cached storage keys are namespaced by the active root and revoked when the process enters isolation
- factory reset uses the centralized gate-aware Keychain deletion path
- child environment overlays cannot replace reserved isolation markers or inherit ambient credential variables in isolated hosts
- model scans and external-model registry updates stage results and commit only while the admitting policy generation remains current
- legacy-root migration stages changes privately, publishes atomically, preserves a concurrently created live root, and cleans up deferred work
- MCP and plugin startup use generation checks and final launch snapshots; stale startup work is cancelled or retried
- path overrides and model-scan hooks used by concurrent tests are lock-backed or task-scoped

## Validation

Validated at `66a7a85c10b3ce4e514acc682e48ca043f9cc768` on `origin/main` `e9ab5551e08c53cedee492a92e9a52b543a3f0d3`:

- `make ci-test`: 7,976 total, 7,928 passed, 48 skipped, 0 failed
- CLI package: 130 XCTest cases and 18 Swift Testing cases passed
- Repository package: 116 tests passed under Thread Sanitizer with no runtime diagnostics
- focused Core concurrency/security matrix: 200 tests across eight suites passed under Thread Sanitizer with no runtime diagnostics
- PluginTestKit: 10 tests passed
- StatsPack: 20 tests passed
- Evals package: 315 tests passed; all 101 deterministic eval floors passed
- repeated red-team stress: 20 runs covering 1,840 sanitizer-instrumented executions passed
- migration single-flight regression: 20/20 repetitions passed
- MCP runner regression: 16/16 repetitions passed
- unsigned Release app build passed, including the embedded CLI, plugin host, and sandbox kernel
- i18n, full SwiftLint policy, ShellCheck, and `git diff --check` passed

## Limits and non-claims

- no live interaction with the user's login Keychain was authorized or run; this update does not claim live login-Keychain proof
- explicit roots are trusted launcher inputs; static unsafe state is rejected, but pathname validation is not a boundary against a hostile same-UID process replacing a path after validation
- host MCP execution remains intentionally unconfined; this PR prevents test credential, persistent-home, and executable-search leakage but does not add process containment
- a process launch cannot be made atomic with an external policy transition; immediate generation/snapshot checks minimize that interval and deterministic transition tests cover the supported design
- this PR does not change prompts, default tools, tool schemas, chat templates, generation defaults, or agent-loop behavior


jjang-ai:
Fresh current-main-adjacent evidence for this isolation lane (2026-08-30): Osaurus PR #2554's unrelated `test-core` failed twice while its changed MTP suite passed. Attempt 2 showed repeated CoreData/NSXPC connection failures and then `ChatSessionQueuedSendTests` observed a nil `sessionId`. The exact queued-send suite on the same PR head passed **24/24 tests in each of five consecutive local runs** (4.97–5.04 s), indicating CI host/isolation flakiness rather than a deterministic queued-send or MTP regression.

This is relevant evidence, not a merge recommendation for the current #2319 head: #2319 is presently `CONFLICTING` with main and its broad 57-file isolation diff needs a current-main rebase/audit. A minimal successor should reproduce the CoreData/XPC failure under CI concurrency and preserve the central fail-closed test-host root/Keychain policy, rather than adding retries to `ChatSessionQueuedSendTests` or weakening its persistence assertions.

Receipts: #2554 workflow `33296337386` attempts 1–2; local logs `/tmp/pr2554-queued-send-{1...5}.log` in the #2554 worktree.

PR #2324, opened 2026-08-08
fix(remote-providers): report Fireworks context windows in discovery

quartz-hertz:
Fireworks model discovery now reports per-model context windows the same way other OpenAI-compatible providers do (vLLM's `max_model_len`, etc.). The gate in `fetchModelsDiscovery` routed Fireworks down the plain `fetchModels` path, dropping `contextLengths`.

Add a Fireworks-specific discovery variant that:

- decodes `contextLength` from the gateway catalog pages
- merges catalog windows with the `/models` result, preferring `/models` when both advertise a window
- surfaces the merged map through `customProviderContextLengths` so the model picker and chat budget honor Fireworks' advertised windows

Refs #2304

### What

Fireworks model discovery wasn't reporting per-model context windows.
The gate in `fetchModelsDiscovery` routed Fireworks down the plain
`fetchModels` path, so `contextLengths` came back empty and the model
picker and chat budget fell back to a default instead of the advertised
window.

### How

Adds a Fireworks-specific discovery variant that decodes `contextLength`
from the gateway catalog pages, merges those with the `/models` result
(preferring `/models` where both advertise a window), and surfaces the
merged map via `customProviderContextLengths`.

## Changes

- [ x] Behavior change
- [ ] Refactor / chore
- [ x] Tests
- [ ] Docs

## Test Plan

Deepseek V4 Flash 0731 reported 108k tokens, now reports 891k under context budget.

## Screenshots

N/A

## Checklist

- [ x] I have read `CONTRIBUTING.md`
- [ x] I added/updated tests where reasonable
- [ x] I verified build on macOS with Xcode 16.4+

### DIFF
--- a/Packages/OsaurusCore/Models/API/FireworksAPI.swift
+++ b/Packages/OsaurusCore/Models/API/FireworksAPI.swift
@@ -25,7 +25,8 @@ struct FireworksModel: Decodable, Sendable {
     let kind: String?
     let supportsServerless: Bool?
     let conversationConfig: FireworksConversationConfig?
-
+    let contextLength: Int?
+    
     /// Serverless-hosted, chat-capable, live model — what the model picker
     /// should show. `conversationConfig` presence is Fireworks' signal that
     /// the Chat Completions API is enabled for the model, but embedding and
--- a/Packages/OsaurusCore/Services/Provider/RemoteProviderService.swift
+++ b/Packages/OsaurusCore/Services/Provider/RemoteProviderService.swift
@@ -5366,6 +5366,9 @@ extension RemoteProviderService {
             let discovery = try await fetchOpenAICompatibleModelsDiscovery(from: provider)
             return (discovery.models, discovery.contextLengths)
         }
+        if isFireworksProvider(provider), provider.authType != .xaiOAuth {
+            return try await fetchFireworksModelsDiscovery(from: provider)
+        }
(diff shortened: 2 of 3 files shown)


RaajeevChandran:
This looks good for the most part but could you fix these

1. Second test doesn't exercise the function.

`fireworksDiscovery_mergesModelsAndPreservesContextLengths` never calls `fetchFireworksModelsDiscovery`. It re-implements the merge loop inline and asserts on that copy, and it skips the final `mergedLengths` filtering step, so it won't catch regressions in the actual code. The comments in it ("you'd need to inject a transport... you can instead test the merge logic directly") also read like leftover notes rather than test docs. `fetchFireworksCatalogModelsDiscovery` already takes a transport parameter, so please rewrite this as a real test that feeds it a canned catalog response (a two-page response would also cover pagination and the context-length merge).

2. Duplication with the existing catalog helpers.

`fetchFireworksCatalogModelsDiscovery` and `decodeFireworksCatalogPageDiscovery` are near-verbatim copies of `fetchFireworksCatalogModels` and `decodeFireworksCatalogPage`, differing only in collecting `contextLength`. Can you make the old ones delegate to the new ones (e.g. `return try await fetchFireworksCatalogModelsDiscovery(...).models`)? Otherwise any future fix to pagination, filtering, or error handling has to land in two places and the two paths will drift.

Issue #216, opened 2025-12-16
Download from Command Line


asbrwl:
### Problem statement

The ability to download a model from the command line. Possibly providing the correct HF model name (exactly as in the UI)

I'm running Osaurus on a Mac Studio with 96GB of RAM. I mostly manage this server from the command line, so downloads is the one gap I'm seeing. 

### Proposed solution

Maybe a solution similar to what ollama uses

osaurus pull mlx-community/AceReason-Nemotron-7B-4bit

PR #2331, opened 2026-08-09
Fix first-turn double prefill (warmup cache-salt mismatch), warmup tool-scope kill, cache-size settings hazards; repin vmlx

jjang-ai:
Second integration PR of the DSV4-Flash campaign (follows #2328). Every fix here is live-proven on the dev app; full instrumented evidence below.

## Warmup cache-salt mismatch — the "first turn prefills 2-3x" bug (all families)

The post-response/pre-send warmup salted its stored KV from the RAW request context, while every real send salts from the PREPARED context — where vmlx merges the bundle-declared reasoning defaults (DSV4-0731: `enable_thinking=true` + `effort=low`). Identical token bytes, different content hash: the send could never restore what the warmup had just stored, so the same ~2.6k-token prefix prefilled twice (or three times with the pre-send handshake) on every first turn. This affects every model family whose bundle declares a reasoning default (DSV4, Raptor/Laguna QAT, Qwen/Ornith toggleables).

Diagnosis chain (vmlx instrumentation from PR vmlx#216): `stableDigests` proved token bytes identical at the stored boundary; the disk row + 37.7 MB payload were on disk while the probe still `noRow`'d; the new `cache/salt` trace named the differing component — warmup `scope=nil` vs send `scope=reasoning=on|effort=low`.

Fix: the send-invariant-prefix warmup path takes its scope salt from the probe render's prepared input (the fallback truncation path already did).

**Live proof (cold boot, DSV4-Flash, 2.6k-token tool-bearing chat):**
- Before: warmup stores `salt=f256f8c4… scope=nil`; send fetches `salt=c4dfcb3e… scope=reasoning=on|effort=low` → `MISS all tiers` → full re-prefill (census: 145 MISS / 4 HIT).
- After: warmup stores `salt=c4dfcb3e…` (identical to send); send logs `HIT disk boundary=2643 remaining=131` — 95% of the prompt restored, only the trailing 131 tokens prefilled.

## Warmup killed permanently by any historical tool call

`handleWarmupAfterRunCompleted` scanned ALL turns for tool activity and cancelled warmup on any hit, so a single tool call at turn N disabled post-response warmup for the rest of the session. Now scoped to the turn that just ran. Live-proven: warmup wall 49 s (cold) → 2 s (hot prefix) across a multiturn session.

## Cache-size settings hazards (settings-wiring audit)

- A non-positive Disk Cache Size was savable through both the UI (no field clamp, Save not gated on validation) and the admin API, and reached `DiskCache(maxSizeGB: 0)` — whose quota pass silently evicts EVERY entry at insert (the disk tier reports stores and hits nothing). The engine-effective path now normalizes non-positive caps to the 10 GB default with an explicit error log, covering both entry points.
- Settings Save is now disabled while any `.error`-severity validation issue is present (previously the banner showed the error and Save proceeded anyway).
- The Compiled Decode toggle now raises the "Restart required" chip: its env export is process-latched at model load, so a mid-session flip saved cleanly but silently kept the old policy until app restart.

## vmlx repin → 44072a46 (vmlx PR #216, merged)

Cache-boundary ladder (mid-transcript divergence keeps its longest prefix), DSV4 concurrency hardening (shared-memo materialization, MoE routing ids off the shared module, compiled regions built once under a lock), native-MTP policy parity (depth capped at 1, hybrid warmup verdict memoized per model), and the cache/salt/layer diagnostics used for the proofs above (all env-gated, off by default).

## Validation

- 23 instrumented cold-boot live trials across three builds on the dev app (graceful-quit + purged SSD cache per trial), driven through the real chat GUI.
- Release build green at the final pin; `swift build` clean on the vmlx side; sabotage-verified source guards unchanged.
- Known-open (tracked, not in this PR): DSV4 cold-boot token soup — now root-caused by the new weight checksums to nondeterministically mis-materialized layer-15/16 quantized tensors at load (~20% of boots); loader fix lands separately.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "3aaab2253172cc1cfae767edf9a1d5f952137b8c"
+        "revision" : "44072a46141c6235c83632c820c536e333c5445f"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "3aaab2253172cc1cfae767edf9a1d5f952137b8c"
+        "revision" : "44072a46141c6235c83632c820c536e333c5445f"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.swift
+++ b/Packages/OsaurusCore/Package.swift
@@ -170,7 +170,7 @@ let package = Package(
         // capped (VMLX_METAL_BUFFER_COUNT_GUARD=0 disables).
         .package(
             url: "https://github.com/osaurus-ai/vmlx-swift",
-            revision: "3aaab2253172cc1cfae767edf9a1d5f952137b8c"
+            revision: "44072a46141c6235c83632c820c536e333c5445f"
         ),
         // FluidAudio 0.14.3 added a breaking `language:` parameter to TTS
         // calls that osaurus's `TTSService` doesn't pass. Pinning to the
--- a/Packages/OsaurusCore/Services/Chat/ChatSessionWarmup.swift
+++ b/Packages/OsaurusCore/Services/Chat/ChatSessionWarmup.swift
@@ -285,7 +285,23 @@ extension ChatSession: ChatWarmupSessionContext {
     }
 
     func handleWarmupAfterRunCompleted(wasCancelled: Bool, hadError: Bool) {
-        let hadToolActivity = turns.contains { turn in
+        // Scope tool-activity detection to the CURRENT run (from the last
+        // user turn onward), not the whole session. Scanning every historical
+        // turn permanently disabled post-response warmup for any chat that
+        // ever used a tool — so a session that ran one tool call at turn 5
+        // paid a cold multi-turn prefill tax on every send from turn 6
+        // through turn 500. The re-render bytes stored by a warmup for a
+        // clean assistant turn remain a valid prefix of the next send's
+        // prompt regardless of what happened many turns ago, because the
+        // committed history is what the next send re-renders and warms
+        // against on both sides of the fingerprint gate.
+        let currentRunTurns: ArraySlice<ChatTurn> = {
+            guard
+                let lastUserIndex = turns.lastIndex(where: { $0.role == .user })
+            else { return turns[...] }
+            return turns[lastUserIndex...]
+        }()
+        let hadToolActivity = currentRunTurns.contains { turn in
             turn.role == .tool
                 || !(turn.toolCalls?.isEmpty ?? true)
                 || !turn.toolResults.isEmpty
--- a/Packages/OsaurusCore/Services/ModelRuntime.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime.swift
@@ -4187,6 +4187,20 @@ public actor ModelRuntime {
         freeFraction: Double = 0.25,
         minUsefulGB: Double = 1.0
     ) {
+        // A non-positive cap is savable through the settings UI and the admin
+        // API (Save is not gated on validation errors). Zero reaches
+        // `DiskCache(maxSizeGB: 0)`, whose quota pass then evicts EVERY entry
+        // at insert — the disk tier reports stores and hits nothing, with no
+        // eviction trace. Normalize here, on the engine-effective path, so
+        // both entry points are covered; the guard below must not run before
+        // this or a 0 cap on an unknown-free-space volume slips through.
+        let rawConfiguredCapGB = Double(config.diskCacheMaxGB)
+        if rawConfiguredCapGB <= 0 {
+            genLog.error(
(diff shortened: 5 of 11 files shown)


jjang-ai:
## Cross-family live proof (pre-merge gate)

Each row is an independent cold boot on the dev app: SSD KV cache purged, model switched, chat window opened through the GUI, prompt typed and sent so it overlaps the in-flight warm-up (the real first-turn case), then a screenshot of the actual chat window. Salt equality is the thing this PR fixes; it is read from the new `[vmlx][cache/salt]` trace on both the warm-up's store and the send's fetch.

| Family | Topology | warm-up vs send salt | Send outcome | Output |
|---|---|---|---|---|
| DeepSeek-V4-Flash (100 GB MoE) | MLA + hybrid pool | **was mismatched → now identical** (`c4dfcb3e`) | `HIT disk boundary=2643 remaining=131` | coherent |
| Ornith-1.0-9B JANG_4M | hybrid SSM (8 KV / 24 mamba) | identical (`0803a803`) | MISS — see note | coherent, 95.3 tok/s |
| Gemma-4-12B qat JANG_4M | rotating SWA | identical (`0803a803`) | no disk fetch — reused resident KV | coherent, 36.0 tok/s |
| Bonsai-27b-Ternary | ternary CRACK | identical (`820d70ea`) | MISS — see note | coherent |
| Raptor-1.0-16B-A3B qat JANG_4M | laguna, 10 KV / 30 rotating | identical (`820d70ea`) | stores clean | see defect note |

**No family regressed.** The salt divergence only ever bit families whose bundle declares a reasoning default (DSV4-0731 declares `enable_thinking=true` + `effort=low`); the other four already agreed on `scope`, and they still agree after the change.

### Boundary-ladder merge-safety A/B
Because this PR is what first ships the vmlx boundary ladder to users, Gemma 4 was run again with `VMLX_CACHE_BOUNDARY_LADDER=0`. Disk-store count was **9 with the ladder on and 9 with it off**, and stream durations sat in the same band — no evidence the ladder adds post-generation cost.

### Two findings this run surfaced (NOT caused by this PR, tracked separately)

1. **A second first-turn double-prefill mechanism.** On fast models the send's cache fetch happens *before* the warm-up's store lands (Ornith: send fetch at log line 24, warm-up store at line 87), so there is nothing on disk to reuse and both prefill the same prefix. This is a send/warm-up race, not a key mismatch, and needs the send to join an in-flight warm-up rather than race it.
2. **Post-output stall is engine-side, not UI-side.** Measured on this build: Gemma's last visible delta at ~+18.8 s with the stream completing at 22.97 s (~4.2 s dead), Bonsai ~+29.6 s vs 79.02 s (~49 s dead). The UI finished 0.12 s after the engine, and the engine's own post-generation tail (GPU drain + KV disk store) measures 5-20 **ms**, so neither the renderer nor the store explains it. Gemma's token count matches its visible text exactly, so no hidden tokens are being generated.

### CI
`test-core` initially failed on the two vmlx-revision tripwires — correctly: the repin lives in **six** files and `Packages/OsaurusCore/Package.resolved` was still on the old revision. Fixed in 9dc34b0c; all checks green.

jjang-ai:
## Post-merge verification — prefill fixes and speedups, measured live

Recording the numbers for this PR's changes now that the follow-up work is in, since the claims above were made before the full matrix ran.

### First-turn prefill (this PR's salt fix)

The warm-up stored its KV under a salt derived from the raw request context while the send salted from the prepared context, so identical token bytes produced different content hashes and the send could never reuse the warm-up's work.

| | before | after |
|---|---|---|
| send verdict | `MISS all tiers tokens=2774` | `HIT disk boundary=2643 remaining=131` |
| prompt re-prefilled | 2,774 tokens (100%) | 131 tokens (5%) |
| cold-boot cache misses per turn | 2 | 1 |

Reproduced on **8 consecutive cold boots** — every one `HIT boundary=2643 remaining=131`.

### Prefix reuse across *different* conversations

A different question entirely (hash tables / B-trees vs merge sort), sharing only the system+tools prefix, against a warm cache:

```
warm-up: HIT disk boundary=2643 remaining=1     ← 2,643 of 2,644 tokens restored
send:    HIT disk boundary=2643 remaining=107   ← 96% of the prompt restored
```

So reuse is not limited to extensions of one conversation — the shared prefix is found and reused by unrelated chats.

### Cross-family multiturn (turn 2 = the prefix cache doing its job)

| model | topology | turn-2 TTFT | prompt | notes |
|---|---|---|---|---|
| Ornith-1.0-9B JANG_4M | hybrid SSM (8 KV / 24 mamba) | **0.61 s** | 3,651 tok | 94.7 tok/s, ladder `[1256, 2330, 3644]` |
| Gemma-4-12B qat JANG_4M | rotating SWA | **0.66 s** | 3,233 tok | 29.3 tok/s, one disk miss in the whole 2-turn session, ladder `[267, 1086, 2400, 3233]` |

Sub-second TTFT on 3k+ token prompts is the multi-turn tax being gone; a cold prefill of that size is seconds.

### Post-answer stall (fixed in the follow-up repin, #2333)

The stall users saw as a spinner after the answer was fully visible was the synchronous cache store, not the renderer:

```
[vmlx][solo/info-gap] sinceLastText=12.3767s  textEvents=1027
[vmlx][gen/tail]      drain1=0.0012s  store=12.3754s  drain2=0.00003s
```

GPU drain 1.2 ms, UI trailing the engine by 0.12 s, store 12.4 s. One turn was writing five cache boundaries because the already-durable skip was gated to stable boundaries only. After widening it, on a two-turn DSV4-Flash chat: **turn 2 went 10.2 s → 0.374 s**. Turn 1 is unchanged (cold cache has nothing to skip).

### Correction to an earlier claim in this thread

An earlier comment ruled the disk store out as the stall's cause on the basis of a 5–20 ms measurement. That measurement was the **warm-up's** store (`max_tokens: 1`); the real per-turn store is three orders of magnitude larger. The instrumentation that made this visible only fires on the path hosts actually take — `maxBatchSize == 1` returns early through `startSoloFastPath`, so the batched loop's probe never ran at all.

Issue #2332, opened 2026-08-10
Inline math delimited with $…$ renders literally in chat markdown


jjang-ai:
## What happens

Inline math written as `$O(n)$` renders as the literal text `$O(n)$` in the chat transcript, while math the model emits in the other supported form renders correctly as proper math **in the same table, in the same message**.

## Reproduce

Gemma-4-12B (`OsaurusAI/OsaurusAI--gemma-4-12B-it-qat-JANG_4M`), asked to compare merge sort and quicksort. Reproduced on two independent runs.

Observed in one rendered table:

| cell | rendered as |
|---|---|
| Merge Sort / Worst-Case | *O(n log n)* — correct math |
| Merge Sort / Memory Use | `$O(n)$` — **literal, unrendered** |
| Quicksort / Worst-Case | *O(n²)* — correct math |

Second run, "Summary Comparison" table: `$O(n)$ (with Timsort) or consistent O(n log n).` — the first term literal, the second rendered, inside one cell.

## Why it matters

`$…$` is the most common inline-math convention in model output, so this is not an exotic edge case — any model that emits it produces transcripts with raw delimiters mixed into otherwise-correct math. It also looks like a model quality problem when it is a renderer gap.

## Expected

`$…$` inline math renders like the other supported delimiters. If supporting it is undesirable (bare `$` is ambiguous with currency), the delimiters should at least be stripped rather than shown raw, and the choice documented.

Issue #2327, opened 2026-08-09
Tool calling not reliable: OsaurusAI/Ornith-1.0-9B-JANG_4M


hk1954:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest `main` or release

### Describe the bug

mcp server tools didn't get called despite being loaded. Instead the custom agent made up its reply. After pointing it out, it called tools correctly. It also doesn't seem to use the knowledge enabled.

### Steps to reproduce

Select a custom agent that accesses mcp server.
prompt to it do something that needs it to call a mcp server tool.
It might hallucinate and make up its response.

### Provider type

Local model (MLX)

### Model name

OsaurusAI/Ornith-1.0-9B-JANG_4M 

### Osaurus version / commit

0.22.18

### macOS version

26.5.2

### Apple Silicon chip

M4 16GB

### Xcode version

26.5

### Logs

```shell
Full thread exported and its .md file attached.
```

### Screenshots

[Get my today's activity..md](https://github.com/user-attachments/files/30867239/Get.my.today.s.activity.md)

jjang-ai:
Hey, thanks for the report, we had some changes regarding tools and may have some parser changes I'm not aware of, I'll look into this and get resolved within 24 hrs.

hk1954:
Thanks! I'll wait for your feedback.

jjang-ai:
Same failure class reproduces on **Raptor-1.0-16B-A3B-qat-JANG_4M** (laguna), so this is not Ornith-specific.

Asked a plain computer-science question with no Osaurus context ("explain merge sort, compare with quicksort on worst-case complexity, memory use and cache behavior"). Instead of answering from its own knowledge, the model:

1. called **Osaurus help** — `Failed · 308ms`
2. thought for 408ms, called **Osaurus help** again — `Failed · 276ms`
3. concluded: *"I could not ground this in one of Osaurus's published topics; there is no `sorting-merge-comparison` or `algorithms` topic, so I will not answer. The nearest thing I can do is make a search agent `osaurus_agent` with `web_search_enabled` and let it find the detail."*
4. spawned an **Algorithm Research** agent and stopped — the user's question was never answered.

Final stats: TTFT 0.33 s, 141 tok/s, 65 tokens.

Two separable problems, and the second is the more damaging one:

- **The knowledge tool fails and the failure is not recovered.** Two consecutive `Failed` help calls should fall back to answering directly, not to spawning an agent.
[...]

PR #2333, opened 2026-08-10
Repin vmlx: refuse truncated model shards, and stop the post-answer stall on every warm turn

jjang-ai:
Repins vmlx to `fd7ce91c` (vmlx PRs #217 and #218). Both changes are things users feel directly, and both are live-proven on the dev app.

## 1. A truncated model no longer loads silently

Weights are memory-mapped, so a shard shorter than its own header does **not** fail to load — tensors whose data falls past end-of-file read as zeros on most launches and as leftover memory on the rest, fixed for the life of the mapping.

`dealign.ai/DeepSeek-V4-Flash-0731-JANG-CRACK` was missing **301.7 MB across 8 of its 102 shards**. `layers.15.ffn.shared_experts.*.scales` began ~66 MB past the end of shard 38. About one launch in five produced gibberish from the very first sampled token; the other four silently dropped that layer's shared-expert contribution and looked fine.

Because the outcome was decided when the file was mapped and then held for the whole process, it presented as a nondeterministic concurrency bug and survived five A/B legs that each disabled a suspected racing component and moved the failure rate not at all.

Load now checks every shard's length against its own header before mapping anything, and fails naming each short shard and its missing byte count.

**Verified:** after re-downloading the shards — bundle reports 102/102 complete, the two tensors read healthy values matching their layer-14 neighbours (0.000284 / 0.001262), and a cold-boot batch is **8/8 clean, zero token soup**. A library-wide scan found 67 of 68 bundles intact, so this was a single bad download rather than a systemic problem.

## 2. The post-answer stall

Every turn ended with the answer fully visible but the turn not finalized — spinner up, no stats. Two probes agree to the millisecond that the stall **is** the synchronous cache store:

```
[vmlx][solo/info-gap] sinceLastText=12.3767s  textEvents=1027
[vmlx][gen/tail]      drain1=0.0012s  store=12.3754s  drain2=0.00003s
```

Not the GPU drain (1.2 ms) and not the renderer — the UI finished 0.12 s after the engine. One turn was writing five cache boundaries, because the "already durable, don't rewrite" skip was gated to *stable* boundaries only, so the rest were rewritten every turn.

**Measured on a two-turn DSV4-Flash chat:** turn 2 went **10.2 s → 0.374 s**. Turn 1 is unchanged at 10.2 s, correctly — a cold cache has nothing to skip. Taking turn 1's cost off the request path requires moving the store behind `.info`, which the engine deliberately avoids today so a new request's GPU work cannot race the store's `MLX.eval`; that is a separate, riskier change and is not attempted here.

## Cache behavior re-verified after both changes

- Same conversation, extension: `HIT disk boundary=2643 remaining=131` — 95% of a 2,774-token prompt restored.
- **Different conversation sharing only the system+tools prefix, warm cache:** `HIT boundary=2643 remaining=1` on warm-up and `remaining=107` on the send. Prefix reuse works across conversations, not just extensions.
- Cross-family multiturn: Ornith-1.0-9B turn-2 TTFT **0.61 s** at 3,651 tokens; Gemma-4-12B turn-2 TTFT **0.66 s** at 3,233 tokens with a single disk miss across the whole two-turn session. Boundary ladders publishing 3–4 rungs on both.

All new diagnostics are gated behind `VMLX_CACHE_FETCH_TRACE=1` and off by default.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "44072a46141c6235c83632c820c536e333c5445f"
+        "revision" : "fd7ce91cde0be283c817142d96c9b3f87efcc5e5"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "44072a46141c6235c83632c820c536e333c5445f"
+        "revision" : "fd7ce91cde0be283c817142d96c9b3f87efcc5e5"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.swift
+++ b/Packages/OsaurusCore/Package.swift
@@ -170,7 +170,7 @@ let package = Package(
         // capped (VMLX_METAL_BUFFER_COUNT_GUARD=0 disables).
         .package(
             url: "https://github.com/osaurus-ai/vmlx-swift",
-            revision: "44072a46141c6235c83632c820c536e333c5445f"
+            revision: "fd7ce91cde0be283c817142d96c9b3f87efcc5e5"
         ),
         // FluidAudio 0.14.3 added a breaking `language:` parameter to TTS
         // calls that osaurus's `TTSService` doesn't pass. Pinning to the
--- a/Packages/OsaurusCore/Tests/Service/ImageGenerationBridgeContractTests.swift
+++ b/Packages/OsaurusCore/Tests/Service/ImageGenerationBridgeContractTests.swift
@@ -74,7 +74,7 @@ struct ImageGenerationBridgeContractTests {
             encoding: .utf8
         )
 
-        let expectedRevision = "44072a46141c6235c83632c820c536e333c5445f"
+        let expectedRevision = "fd7ce91cde0be283c817142d96c9b3f87efcc5e5"
         #expect(packageSwift.contains(#"revision: "\#(expectedRevision)""#))
         #expect(packageResolved.contains(#""revision" : "\#(expectedRevision)""#))
         #expect(workspaceResolved.contains(#""revision" : "\#(expectedRevision)""#))
--- a/Packages/OsaurusCore/Tests/Service/RuntimePolicySourceTests.swift
+++ b/Packages/OsaurusCore/Tests/Service/RuntimePolicySourceTests.swift
@@ -786,7 +786,7 @@ struct RuntimePolicySourceTests {
         // and both xcworkspace Package.resolved files. Miss one and a release
         // surface resolves a revision nobody proved. OsaurusEvals resolves
         // this manifest transitively and its local Package.resolved is ignored.
-        let expectedRuntimeHardenedRevision = "44072a46141c6235c83632c820c536e333c5445f"
+        let expectedRuntimeHardenedRevision = "fd7ce91cde0be283c817142d96c9b3f87efcc5e5"
         let manifestRevision = try Self.vmlxPinRevision(in: manifest)
         let coreResolvedRevision = try Self.vmlxPinRevision(in: coreResolved)
         let workspaceRevision = try Self.vmlxPinRevision(in: workspaceResolved)
--- a/osaurus.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/osaurus.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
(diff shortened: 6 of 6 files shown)


jjang-ai:
## Post-merge proof: prefill fixes, speedups, and the changing-prefix cases

All figures below are live on the dev app with this repin, DSV4-Flash unless noted, `VMLX_CACHE_FETCH_TRACE=1`.

### Post-answer stall — the thing users actually see

The spinner that stayed up after the answer was fully rendered was the synchronous cache store, not the renderer. Two independent probes agree to the millisecond:

```
[vmlx][solo/info-gap] sinceLastText=12.3767s  textEvents=1027
[vmlx][gen/tail]      drain1=0.0012s  store=12.3754s  drain2=0.00003s
```

GPU drain 1.2 ms; the host UI finished 0.12 s after the engine; the store 12.4 s. One turn was writing five boundaries because the already-durable skip was gated to *stable* boundaries only.

| two-turn DSV4-Flash chat | before | after |
|---|---|---|
| turn 1 (cold cache) | 10.2 s | 10.2 s — correct, every write genuinely new |
| **turn 2 (warm)** | ~10 s | **0.374 s** |
| warm-ups | 20 ms | 20–26 ms |

### First-turn prefill

| | before | after |
|---|---|---|
| send verdict | `MISS all tiers tokens=2774` | `HIT disk boundary=2643 remaining=131` |
| re-prefilled | 2,774 tokens (100%) | 131 tokens (5%) |
| misses per cold boot | 2 | 1 |

Held on **8 consecutive cold boots**.

### Prefix reuse across different conversations

Different question (hash tables / B-trees), sharing only the system+tools prefix, warm cache:

```
warm-up: HIT disk boundary=2643 remaining=1
send:    HIT disk boundary=2643 remaining=107
```

### Changing prefixes: reasoning-effort flip

Worth recording because it corrected an assumption. The disk key folds the reasoning scope in, so a flip looked like it might fragment the cache for a prompt whose bytes were unchanged. Measured, the bytes are **not** unchanged — the effort preface is injected inside the prefix:

```
effort=low :  prompt=2647  stable=[1114, 2643]
effort=high:  prompt=2675  stable=[1142, 2671]   (+28 tokens)
```

So the miss on the flip is a genuinely different prompt, not over-fragmentation, and the ladder re-establishes immediately at the new effort: `MISS` on the first warm-up, then `HIT disk boundary=2671 remaining=107` on the send. Correct behaviour, one cold turn per effort change.

### Cross-family multiturn

| model | topology | turn-2 TTFT | prompt | throughput |
|---|---|---|---|---|
| Ornith-1.0-9B JANG_4M | hybrid SSM | **0.61 s** | 3,651 tok | 94.7 tok/s |
| Gemma-4-12B qat JANG_4M | rotating SWA | **0.66 s** | 3,233 tok | 29.3 tok/s |

Ladders publishing `[1256, 2330, 3644]` and `[267, 1086, 2400, 3233]`; Gemma took a single disk miss across an entire two-turn session.

### Truncated-shard guard

Verified against the real artifact: the repaired bundle reports 102/102 shards complete and the two previously-unreadable tensors now read 0.000284 / 0.001262, matching their layer-14 neighbours. A deliberately truncated copy of shard 38 is refused with `declared=1030399712 onDisk=960399712 missing=70000000`. Cold-boot batch after repair: **8/8 clean, zero token soup** (was ~20% gibberish).

### Still open, tracked separately

- Turn 1's 10.2 s store is unchanged; removing it needs the store moved behind `.info`, which the engine deliberately avoids today so a new request's GPU work cannot race the store's `MLX.eval`.
- Tool-set changes mid-conversation are not yet measured (expected to behave like the effort flip — the `<tools>` block lives in the same prefix region).
- #2327 (Raptor tool-loop hijack) and #2332 (`$…$` math rendering) are unrelated to this repin.

jjang-ai:
## Tool availability, tool usage, and tool-loop cache reuse — DSV4-Flash, live

Prompt deliberately mixed a tool the model has with one it does not:

> "What is the current time, and then create a short note file named campaign-notes.txt containing the single line PREFIX CACHE OK. Use your tools to do it, do not just describe the steps."

**Result:**

- `Thought for 20s` (2.3k chars) — reasoning enforced on a fresh first turn, with the chip correctly reading **Low**
- **`Get current time · 1.3s`** — real tool call, succeeded
- Answered with the true value: *"Current time: Sunday, August 9, 2026 at 10:54 PM PDT (2026-08-09T22:54:59-07:00)"*
- Then correctly **declined** the impossible half: *"I can't create `campaign-notes.txt` with my available tools — my toolset here (Osaurus configuration tools, time lookup, web search) has no file-writing action, and `share_artifact` only works for files that already exist. If you'd like, I can create an agent whose job is to write that file for you."*

That last part is the behaviour we want and the exact opposite of #2327: the model enumerates what it actually has, uses the one that applies, and refuses to invent the one it doesn't — rather than looping on a failing tool or spawning an agent instead of answering.

No tool-call parse failures, no envelope leakage, no protocol errors in the log.

### Cache reuse across the tool loop

Every request in the loop — including the post-tool continuations — restored its prefix:

```
HIT disk boundary=2643 remaining=1      (warm-up)
HIT disk boundary=2643 remaining=125    (send)
HIT disk boundary=2767 remaining=676    (after the tool result)
HIT disk boundary=3442 remaining=174    (continuation)
```

Four consecutive hits, zero cold prefills. The tool result extends the transcript and the next request picks up the immediately-preceding boundary, so an agent turn costs one prefill of just the new tokens instead of the whole conversation.

### Decode speed

**34.7 tok/s**, TTFT 3.62 s — inside the 33–35 tok/s target band for DSV4-Flash short decode, measured on a real tool-using agent turn rather than a synthetic row.

PR #2335, opened 2026-08-10
Repin vmlx: capture cache boundaries during prefill (fixes a family that could never populate its cache)

jjang-ai:
Repins vmlx to `a8012d26` (vmlx PR #221).

## What was wrong

After generation the engine reconstructed the cache state at each boundary it stores. Topologies it can trim take a cheap path; everything else **replays the prefix through the model** — an extra prefill per boundary, on the request path, after the user's answer is already visible.

That reconstruction is also cancellable, and on a slow enough model it loses the race:

```
[vmlx][cache/store-boundary] boundary=2987 store=2986 allowRederive=true snapshot=nil
[vmlx][cache/rederive-failed] tokens=2986 error=CancellationError()
```

**Bonsai-27B therefore never wrote its stable seed at all** — every new conversation cold-prefilled the entire shared system+tools prefix, permanently. Ornith-9B re-derived fast enough to finish, which is why the identical topology worked on one model and not the other.

## The fix

Prefill already passes through those boundaries, so the state is captured as it goes by instead of being rebuilt afterwards. Segments run through the model's real prepare/forward path in order, so sampling and template behaviour are unchanged.

## Live result — Bonsai-27B, two conversations sharing only the system+tools prefix

| | before | after |
|---|---|---|
| warm-up | `MISS all tiers tokens=2987` | **`HIT disk boundary=2986 remaining=1 ssm=96`** |
| send | `MISS all tiers tokens=3113` | **`HIT disk boundary=2986 remaining=126 ssm=96`** |
| re-derive failures | `CancellationError()` | none |

`ssm=96` is the recurrent companion state for all 48 mamba layers restored alongside the KV.

## Regression checks on the other families (unchanged, all still hit)

Reasoning **on and off**, two conversations each, judged within one reasoning state:

| model | reasoning on | reasoning off |
|---|---|---|
| DeepSeek-V4-Flash | HIT 2671 / 2752 / 2855 | HIT 2592 ×2, 2694 |
| Ornith-1.0-9B | HIT 2350 (ssm=48) | HIT 2350 (ssm=48) |
| Gemma-4-12B | HIT 2321 | HIT 2319 |

Reasoning shifts the prefix itself (`stable=[1142,2671]` on vs `[1063,2592]` off) and reuse re-establishes correctly inside each state.

## Not claimed

Bonsai's **cold**-turn store is still ~25 s. The cold path writes genuinely new boundaries; reducing that is separate work and is not addressed here.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,400 +1,400 @@
 {
-  "originHash" : "a70de644fa9a976403022014284317c01c48d4d19f71b277a396f9c2786a3827",
-  "pins" : [
+  "originHash": "a70de644fa9a976403022014284317c01c48d4d19f71b277a396f9c2786a3827",
+  "pins": [
     {
-      "identity" : "aachartkit-swift",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/AAChartModel/AAChartKit-Swift.git",
-      "state" : {
-        "revision" : "b51e53c3b04372c1ed1906b0ec3ba45733953ef4",
-        "version" : "9.5.0"
+      "identity": "aachartkit-swift",
+      "kind": "remoteSourceControl",
+      "location": "https://github.com/AAChartModel/AAChartKit-Swift.git",
+      "state": {
+        "revision": "b51e53c3b04372c1ed1906b0ec3ba45733953ef4",
+        "version": "9.5.0"
       }
     },
     {
-      "identity" : "aptabase-swift",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/aptabase/aptabase-swift.git",
-      "state" : {
-        "revision" : "cfd67fac2a228d448d9d2ac92ffc71589cc3ef00",
-        "version" : "0.3.11"
+      "identity": "aptabase-swift",
+      "kind": "remoteSourceControl",
+      "location": "https://github.com/aptabase/aptabase-swift.git",
+      "state": {
+        "revision": "cfd67fac2a228d448d9d2ac92ffc71589cc3ef00",
+        "version": "0.3.11"
       }
     },
     {
-      "identity" : "async-http-client",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/swift-server/async-http-client.git",
-      "state" : {
-        "revision" : "7744c2a035c68ec14726c709f031835e3e30bde1",
-        "version" : "1.34.0"
+      "identity": "async-http-client",
+      "kind": "remoteSourceControl",
+      "location": "https://github.com/swift-server/async-http-client.git",
+      "state": {
+        "revision": "7744c2a035c68ec14726c709f031835e3e30bde1",
+        "version": "1.34.0"
       }
     },
     {
-      "identity" : "containerization",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/containerization.git",
-      "state" : {
-        "revision" : "44bec8b9933bc491d0cbf44abac90a1f6aaebf6b",
-        "version" : "0.35.0"
+      "identity": "containerization",
+      "kind": "remoteSourceControl",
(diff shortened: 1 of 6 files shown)


jjang-ai:
## Stop-mid-turn behaviour on this build (DSV4-Flash, live)

Sent a long essay prompt, pressed **Stop** well into generation, then sent a second unrelated prompt in the same chat.

**No corruption, no cache loss.** Every request in the session restored its prefix, including the one immediately after the Stop:

```
MISS all tiers tokens=2593          (cold first warm-up)
HIT disk boundary=2592 remaining=111
HIT disk boundary=2702 remaining=1521
HIT disk boundary=4222 remaining=1440
HIT disk boundary=5661 remaining=1553
HIT disk boundary=7213 remaining=2265
HIT disk boundary=9502 remaining=1     <- turn after Stop, ~13k context
```

- `REFUSED offset/key mismatch`: **0** — the interrupted turn stored nothing invalid. `includeGeneratedBoundary` is already gated on `stopReason == .stop`, so a user Stop never publishes the generated boundary, and with this PR the stable snapshots are taken during prefill (before any generation), so they are clean regardless of when the Stop lands.
- Continuation stayed coherent: *"A hash collision is when two distinct inputs produce the same hash value, causing them to map to the same bucket."* — 43.5 tok/s, TTFT 15.68 s at ~13k context.

## One inefficiency this exposed (not fixed here)

```
[vmlx][cache/rederive-failed] tokens=1062 error=CancellationError()   x3
[vmlx][cache/rederive-failed] tokens=2591 error=CancellationError()   x3
```

`1062` and `2591` are the stable N−1 seeds (`1063-1`, `2592-1`) — exactly the boundaries the prefill-time capture in this PR is meant to make cheap. They still went through the cancellable re-derive because `prepare()` returns early on the pre-existing `prefillBoundaryCapture` branch, so when the disk-seed capture applies the stable-boundary pass never runs.

Here it cost wasted work rather than correctness (those boundaries were stored anyway and every fetch hit), but it is the same mechanism that left Bonsai unable to populate its cache at all, so the two capture paths should be merged into one ordered pass. Tracked; not in this PR.

jjang-ai:
## Follow-up: the Stop-mid-turn cancellations are gone

The earlier comment recorded that a Stop mid-turn left six cancelled boundary rebuilds:

```
[vmlx][cache/rederive-failed] tokens=1062 error=CancellationError()   x3
[vmlx][cache/rederive-failed] tokens=2591 error=CancellationError()   x3
```

Tracing the store loop's own decisions showed why the prefill-time capture in this PR did not cover them:

```
[vmlx][cache/store-boundary] boundary=1063 store=1062 stable=true allowRederive=true snapshotTokens=2702 snapshot=nil
[vmlx][cache/store-boundary] boundary=2592 store=2591 stable=true allowRederive=true snapshotTokens=2702 snapshot=nil
```

On a turn that **restores from cache**, prefill begins at the restored boundary (2592 here) and never crosses 1062 or 2591 — so no snapshot can exist for them. They were already on disk from earlier turns, but the skip was gated on `hasValidatedDiskEntry`, which trusts only what the current process wrote or read. A fresh process, or any restored turn, therefore rebuilt bytes it already had — and that rebuild is the cancellable one.

Two commits: continue capturing stable boundaries past the first split (`e4971694`), and skip any boundary already durable on disk regardless of who wrote it (`98bc9f9e`, via a new content-addressed `hasDurableEntry` that checks the index row against the file rather than this process's bookkeeping).

**Live, same Stop-mid-turn scenario:**

| | before | after |
|---|---|---|
| `rederive-failed` (CancellationError) | **6** | **0** |
| `REFUSED offset/key mismatch` | 0 | 0 |
| fetch verdicts | all HIT | all HIT — `2592, 2592, 2702, 4156, 5776, 7214` |

## Note on the CI failure earlier in this PR

`test-core` went red on the two revision tripwires, and the cause was mine: repinning by rewriting the `.resolved` files through a JSON serializer reformatted every pin from `"revision" : "..."` to `"revision": "..."`. The tripwires match that substring exactly, so all three files stopped matching — the guard working precisely as intended against a change that looks like a no-op in a diff. Repinning is now a targeted text substitution inside the `vmlx-swift` pin object, and both tripwire suites (110 tests) pass locally.

PR #2336, opened 2026-08-10
Bound capability group-load schemas + repin vmlx (Muse Glimmer runtime)

jjang-ai:
**Draft until live proof lands.** Nothing here has been run in the dev app yet.

## 1. Raptor "returned empty output after tool execution"

A plugin group load appends every loaded tool's full JSON schema to its own tool
result. That is deliberate — delivering schemas in the conversation *suffix*
keeps the paged-KV `<tools>` prefix byte-stable instead of rewriting it mid-run —
but the payload was unbounded. `enabledManifestToolCap` limits how *many* tools
may load, not how *large* the result gets.

The Mail plugin is nine tools whose skill document already spends ~2.2k tokens.
Loading it left the continuation almost nothing to generate into, and the model
answered with empty turn after empty turn until the agent loop gave up. Smaller
results like web search were unaffected, which matches the report exactly.

Three things ruled out first, so the fix isn't chasing the wrong cause:
- The loop logic is sound (`maxEmptyTurnRetries = 2`, nudge notice, counter
  resets on a productive turn) — the fallback firing means **three** consecutive
  empty turns even after being nudged.
- `lengthExhausted` did not fire.
- A generic `capabilities` repro produced **0** empty-output signals, so the
  large payload is necessary to trigger it.

Schemas now degrade a tier at a time rather than truncating mid-JSON — a
fragment is worse than no schema because the model tries to call it. Full →
compact skeletons → names with one-line descriptions. Every tier keeps the tools
callable (registry dispatch is by name) and says how to pull a single tool's
full schema back. Tiers are all-or-nothing so the result does not depend on
registry iteration order.

5 new tests pass.

## 2. Repin vmlx-swift to `ed9abd9e` — Muse Glimmer 30B runtime

vmlx-swift #222: dense 30B image+video VLM. Text tower (five Gemma divergences,
each numerically test-guarded), windowed 50-layer vision tower, image/video
processor, and the ATEM tool-call parser. 25 tests across 3 suites, full build
clean, existing 189 tool-call tests unaffected.

All six pin sites verified at exactly one occurrence, and `"revision" : `
spacing preserved (reformatting that is what turned `test-core` red on #2335).

## Live proof — NOT YET RUN

Gate for un-drafting, per `MuseGlimmer_PROOF_MATRIX.md`:
- Raptor + Mail plugin: the original failure no longer reproduces
- Muse Glimmer: >= 25 tok/s sustained (no spec decode), EOS, stop-mid-turn +
  continue, SSD block match/write/reuse, TTFT, reasoning on/off, effort flips
  and tool toggles mid-conversation, image and video turns

Note for the cache legs: Muse renders reasoning strength **and** the tool block
into the *system prefix*, so an effort flip and a tool toggle each invalidate the
entire prefix — no partial suffix match is possible for this family. A cache HIT
on the turn that changed the prefix is a collision bug, not a win.


jjang-ai:
Final proof summary before merge — all on the live dev-app GUI with the real JANG bundles:

**Muse Glimmer runtime (vmlx #222, now merged)**
- Coherent generation after the normed-embedding fix; the earlier multilingual token soup is gone. Tokenizer/template/specials proven clean via bundle probes first (specials atomic, single `Reasoning strength:` line, exact round-trip), isolating the fault to the forward pass.
- Foreground → new window → generate no longer crashes (previously 6/6 reproducible rotating-KV overrun; fixed at three layers and re-run clean).

**Reasoning enforced via the system prompt, proven from live salts**
User-driven runs show `strength=high` ×6, `strength=low` ×3, `strength=medium` ×3 arriving in the cache scope salt — card → `reasoning_strength` template variable → rendered sentence → per-strength cache identity (distinct salts, no key collision). The four-level card (low/medium/high/**xhigh**) ships in this PR; lower strengths legitimately emit less/no `to=self` thinking.

**Channel envelope routed to the thinking rail**
Live transcript exposed `<|start|>assistant to=self<|message|>…<|eom|>` / `to=user<|message|>…`; the parser routes `to=self` to reasoning and strips every marker. Tests pin the exact transcript.

**Tool path**
ATEM stamps resolve end to end (`jang_config.chat.tool_calling.parser="atem"` → capability gate → parser); 11 parser tests incl. malformed inputs; zero parse/`invalid_args` errors in any live log this session. The Raptor capability-budget fix in this PR carries its own 5 tests.

Known and documented (MuseGlimmer_PROOF_MATRIX.md): effort flips mid-conversation cold-prefill by design (strength is a system-prefix sentence, so the whole prefix changes); optimization legs (sustained tok/s, cold-boot reuse timing) tracked for the follow-up pass.

PR #2344, opened 2026-08-11
Fix #2327: a reasoning-only turn with tools offered was classified as a finished answer

### LINKED ISSUE #2327: Tool calling not reliable: OsaurusAI/Ornith-1.0-9B-JANG_4M

jjang-ai:
Closes the mechanism behind #2327.

## What the model does

Ornith-1.0-9B, two tools offered, asked for the current date and time:

```
<think> The user is asking for the current date and time. I have a tool called
"get_current_time" that can return this information. Let me call it. </think>
<|im_end|>
```

49 tokens, natural `stop`, **no tool call, no content**. Deterministic, and
identical at JANG_4M and JANG_6M — so not quantization. Unchanged by adding a
system prompt. The same model answers a knowledge question correctly with no
spurious call, so it is not general tool-framing either.

## Why the app called that a success

`classifyTerminal` puts both of its reasoning guards behind
`requiresVisibleFinalResponse`, which is
`hasStructuredToolWork || isRemoteAgentTarget` — **both false on the first
turn**, which is precisely when this occurs. The blank-content branch below them
only fires when thinking is *also* blank. A turn with reasoning and nothing else
therefore fell through to `.finalResponse`.

The run ended successfully with a thought bubble and no answer. That is what the
reporter experienced as the model "making up its reply", and why pointing it out
made the next turn work — the nudge re-enters with the intent in context.

## The change

A turn that produced no content and no tool call is not a finished answer when
the model had tools available. The classifier's own doc already says the
reasoning rail is "visible diagnostic work, while the user-facing answer belongs
in content"; this makes the code agree with it.

Deliberately narrow — reasoning-only bundles with **no** tools offered still
complete exactly as before, and that path has its own test so the behaviour
cannot quietly widen.

## Tests

Five cases pinning each branch: reasoning-only with tools (now recoverable),
reasoning-only without tools (still completes), after structured tool work
(unchanged), fully empty (still `.emptyResponse`), normal answer (unaffected).
77 agent-loop tests pass.

Also includes the probe that reproduces the model behaviour, gated behind
`ORNITH_TOOLS` so CI does not load 9GB of weights. Two measurement traps are
documented in it, both hit while writing it: `Generation.chunk` is `nil` for
`.toolCall`, so collecting only chunks scores a correct call as an empty turn;
and the model's reasoning names the tool it is considering, so string-searching
the prose scores a call that never happened. Scoring reads the tool-call channel
only.

## Not covered

The over-calling half of #2327 (Raptor calling `osaurus_help` for a merge-sort
question) did not reproduce on Ornith and is likely a separate defect. The issue
should stay open for it.


jjang-ai:
**Proof level, stated plainly: this is unit-proven, not GUI-proven.**

What is verified:

- five classifier tests pinning every branch — reasoning-only with tools (now
  recoverable), reasoning-only without tools (still completes), after structured
  tool work (unchanged), fully empty (still `.emptyResponse`), normal answer
  (unaffected);
- 77 agent-loop tests pass with the change;
- the model behaviour it responds to is reproduced deterministically by the
  probe in this branch, at both quants.

What is **not** verified: the fix has not been exercised through the running app
UI. Doing that needs the reasoning-only stop to occur live, which is
model-specific — it reproduces on Ornith-1.0-9B, and I could not switch the app
to that model reliably through the GUI (the model picker did not open from a
synthetic click, and `defaultModel` governs new chats rather than a restored
session).

So the classifier logic is proven and the model behaviour is proven; the
join between them, inside the app, is not. Worth a manual pass on Ornith before
this is treated as closing the user-visible half of #2327.

jjang-ai:
## Live proof on the current build

Rebuilt this branch merged with `main` (vmlx pin `751a5779`), signed, launched, and drove the real chat UI with **Ornith-1.0-9B-JANG_4M** — the exact model from #2327.

Asked "What is the current date and time?" (the reporter's prompt). The model called the tool and answered from the result:

> The current date and time is **Tuesday, August 11, 2026 at 7:28 PM** in the America/Los_Angeles timezone (UTC-7), which corresponds to **19:28:17 on 2026-08-11**.

66 generated tokens, 162 chars of content, natural `stop`, no fabricated reply.

The gated probe also ran against the real weights this time — **32.0s / 32.1s** on the 4-bit and 6-bit bundles, both passing.

## One fix pushed since the last review

Running the probe the way this PR's body describes silently **skipped** it and still printed `** TEST SUCCEEDED **`:

```
Test case 'OrnithToolSelectionTests/quantAB()' skipped (0.000 seconds)
** TEST SUCCEEDED **
```

`xcodebuild` runs tests in a separate `xctest` process that does not inherit the invoking shell's environment — only `TEST_RUNNER_`-prefixed variables are forwarded, with the prefix stripped. So it is:

| runner | spelling |
|---|---|
| `swift test` | `ORNITH_TOOLS=1` |
| `xcodebuild test` | `TEST_RUNNER_ORNITH_TOOLS=1` |

The gate now accepts both and its doc comment carries both invocations, so the next person gets a real run instead of a green skip. Given this PR exists because a turn that measured nothing was reported as a success, a probe that reports success while measuring nothing was worth closing too.

PR #2360, opened 2026-08-11
Nemotron 3.5 Lightning: native MTP head + fix the fallback template eating its newlines

jjang-ai:
Repins vmlx-swift to `b3b0f078`.

## The reasoning bug

Nemotron 3.5 Lightning ships its chat template **only** as a standalone `chat_template.jinja` — there is no `chat_template` in `tokenizer_config.json` — and vmlx's tools branch engages the `NemotronMinimal` fallback on every tool-enabled turn. So the fallback, not the bundle file, rendered every Lightning prompt.

Its generation prompt was written as

```jinja
<|im_start|>assistant
{%- if enable_thinking %}
<think>
```

and each `{%-` strips the newline in front of it, so the tail collapsed to `<|im_start|>assistant<think>` where the bundle emits `<|im_start|>assistant\n<think>\n`. The newline after `<|im_end|>` went the same way, gluing every turn boundary.

The model had never seen that spelling, so it closed the reasoning block on its first token: reasoning enabled, block open, zero thinking.

## Proof

Captured with `VMLINUX_REASONING_PROMPT_DUMP_DIR` on the running app, same prompt and same persisted `disableThinking: false` before and after:

| | prompt tail | generated tokens | reasoning stored |
|---|---|---|---|
| before | `...step by step.<\|im_end\|><\|im_start\|>assistant<think>` | 637 | **0 chars** |
| after | `...step by step.<\|im_end\|>\n<\|im_start\|>assistant\n<think>\n` | 1585 | **2492 chars** |

Reasoning read straight from `~/.osaurus/chat-history/history.sqlite`.

## Also in this pin

Nemotron-H native MTP head (`mtp.layers.0/1`), default OFF, with the bounded-KV-window guard so speculation is declined when `maxKVSize` is set.

## Scope check

Only the four Lightning bundles in the local library are affected by the standalone-template + fallback combination; the survey of all 45 toggleable-thinking bundles is in the vmlx PR.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "56bcef0f7d5ddbfaa39216bd64188fe8d1864e97"
+        "revision" : "b3b0f0787633fbace1d7c21053d219235195880d"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "b3b0f0787633fbace1d7c21053d219235195880d"
+        "revision" : "2c79c27e2416283b542c663c0f707bc554aa6af7"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "6daa2ca04df88dc5f591f65bcf678acf54879d62"
+        "revision" : "56bcef0f7d5ddbfaa39216bd64188fe8d1864e97"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "56bcef0f7d5ddbfaa39216bd64188fe8d1864e97"
+        "revision" : "b3b0f0787633fbace1d7c21053d219235195880d"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "b3b0f0787633fbace1d7c21053d219235195880d"
+        "revision" : "2c79c27e2416283b542c663c0f707bc554aa6af7"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "6daa2ca04df88dc5f591f65bcf678acf54879d62"
+        "revision" : "56bcef0f7d5ddbfaa39216bd64188fe8d1864e97"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.swift
+++ b/Packages/OsaurusCore/Package.swift
@@ -194,9 +194,20 @@ let package = Package(
         // end of the first think block, and stored in `thinking` where history
         // replayed it back to the model as prose. `to=user` still closes
         // reasoning and `to=self` still opens it.
+        // vmlx-swift#234 implements the Nemotron-H native MTP head (270
+        // `mtp.*` tensors that `sanitize` previously dropped) and declines
+        // speculation whenever the KV window is bounded, since a
+        // `RotatingKVCache` cannot un-write a rejected draft.
+        //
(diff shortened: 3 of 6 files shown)


jjang-ai:
## Live proof — Nemotron 3.5 Lightning 30B A3B JANG_4M

All of the following was driven through the running dev-built app's own GUI (chat window via `screencapture -l` + OCR + `CGEventPostToPid`; Server settings via the accessibility tree), and read back from `~/.osaurus/chat-history/history.sqlite` and `~/.osaurus/config/server-runtime.json`.

### 1. Reasoning — the bug and the fix

Same prompt, same model, same persisted `disableThinking: false`:

| | prompt tail (from `VMLINUX_REASONING_PROMPT_DUMP_DIR`) | gen tokens | reasoning |
|---|---|---|---|
| before | `…step by step.<\|im_end\|><\|im_start\|>assistant<think>` | 637 | **0 chars** |
| after | `…step by step.<\|im_end\|>\n<\|im_start\|>assistant\n<think>\n` | 1585 | **2492 chars** |

### 2. Reasoning holds across the cases that kept regressing

| case | gen tokens | reasoning | TTFT |
|---|---|---|---|
| **first turn, fresh conversation** | 790 | **2035 chars** | 1.83 s |
| turn 1 (train problem) | 1585 | 2492 chars | 1.85 s |
| turn 2, same conversation | 2367 | 3821 chars | 2.11 s |
| tank problem, MTP Force-On | 519 | 547 chars | 1.82 s |
| tank problem, MTP Off | 358 | 147 chars | 1.97 s |

TTFT stays flat (1.82–2.11 s) even as history grows, so prefix reuse is holding.

### 3. Tool calling

Recorded on a Lightning turn — well-formed JSON, correct name, no malformed XML:

```json
{"type":"function","id":"call_9010768198224DEAB4886B16",
 "function":{"name":"get_current_time","arguments":"{\"timezone\":\"America/Los_Angeles\"}"}}
```

27 tool schemas are present in the rendered prompt.

### 4. MTP toggle — driven live in Settings → Server → Speculative Decoding

Each change made by clicking the radio and pressing **Save Changes**, then reading the config file:

| action | `mtp.mode` |
|---|---|
| start | `auto` |
| click Force On + Save | `force_on` |
| click Off + Save | `off` |
| restore Auto + Save | `auto` |

Before this PR the control was inert for Nemotron — the model-type allowlist was Qwen-only and the head was behind a process-level env `let`.

Under Force-On the model still loads and answers normally (519 tokens, 547 chars reasoning). That is the fail-closed path: the bundle ships no `vmlx_mtp_tuning.json`, so Force-On resolves to `.blocked`, `nativeMTP` stays false, nothing throws — and the user now gets the true reason instead of a family rejection.

### 5. Runtime diagnostics after the multiturn session

| metric | value |
|---|---|
| Native MTP | **Not active** (D1, the default) |
| Prefix hits / misses | 0 / 0 |
| Disk L2 hits / misses / stores | **8 / 81 / 8** |
| SSM hits / misses / re-derives | **8 / 0 / 0** |

Every hybrid-SSM state lookup hit with zero re-derives — the expensive part for a Mamba/attention hybrid — and SSD-L2 is both storing and hitting.

### Scope

Of the 45 toggleable-thinking bundles in the local library, only the four Lightning bundles hit the standalone-template + fallback combination that caused this. Muse Glimmer was never affected (it has no `enable_thinking` kwarg at all).

jjang-ai:
## Blast radius of the template fix

Scanned every local bundle for the condition that engages `NemotronMinimal` (`bos == "<s>"` + `<|im_end|>` in the tokenizer + tools present). **11 bundles, all Nemotron family:**

- `JANGQ-AI/Nemotron-3.5-Lightning-30B-A3B-{JANG_2L,JANG_4M,JANG_6M,MXFP8}` — sidecar `.jinja`
- `OsaurusAI/Nemotron-Labs-Audex-{2B-4bit,2B-6bit,2B-8bit,30B-A3B-4bit,30B-A3B-6bit}` — sidecar `.jinja`
- `dealign.ai/Nemotron-Omni-Nano-{JANGTQ4,MXFP4}-CRACK` — template in `tokenizer_config.json`

No other family can reach this template, so nothing outside Nemotron changes.

And the corrected separators are not Lightning-specific — Audex's own template emits **byte-identical** strings:

```
'<|im_end|>\n'
'<|im_start|>assistant\n<think>\n'
'<|im_start|>assistant\n<think></think>'
```

So the five Audex bundles were being handed the same `assistant<think>` prompt and are fixed by the same change.

## Quant note

`JANG_2L` is **not** llama.cpp imatrix — it is `quantization.method = "jang-importance"` (weight-magnitude scoring, MSE, asymmetric) with a per-tensor bits/group_size manifest under `config.json > quantization`, which the existing mixed-bit loader already consumes. No special handling needed.

## Not in this PR

`JANGQ-AI/Nemotron-3.5-Lightning-*` is not published on Hugging Face (only NVIDIA's originals and third-party GGUF/NVFP4 exist), so there is no curated-catalog entry to add yet — users can run it only from a local `~/models` copy.

jjang-ai:
## Reasoning ON *and* OFF, both proven live through the UI

Toggled via the chat model picker (`/model` → MODEL OPTIONS → Thinking), same prompt each time, prompt captured with `VMLINUX_REASONING_PROMPT_DUMP_DIR`:

| Thinking | rendered prompt tail | gen tokens | reasoning stored |
|---|---|---|---|
| **ON** | `…<\|im_end\|>\n<\|im_start\|>assistant\n<think>\n` | 1173 | **2161 chars** |
| **OFF** | `…<\|im_end\|>\n<\|im_start\|>assistant\n<think></think>` | 290 | **0 chars** |

Both tails now carry the separators the bundle specifies — `<|im_end|>\n` before the header and `assistant\n` inside it — which is exactly what this PR fixes. Before it, both rails rendered `assistant<think>` and the model closed the block immediately, so "off" and "on" were indistinguishable and "on" produced nothing.

Enforcement is correct in both directions, on the first turn of a fresh conversation and across multiturn.

jjang-ai:
## Correction to the blast-radius comment above

I said the five `Nemotron-Labs-Audex-*` bundles were broken the same way and are fixed by this change. **That is wrong.** I tested it by switching to Audex 2B 8-bit in the live picker and the app returned:

```
Error: Unsupported model type: nemotron_dense_audex
```

Their config `model_type` values are **not** `nemotron_h`:

| bundle | model_type | dispatched by vmlx? |
|---|---|---|
| Nemotron-Labs-Audex-2B-* | `nemotron_dense_audex` | **no** |
| Nemotron-Labs-Audex-30B-A3B-* | `nemotron_h_audex` | **no** |
| Nemotron-3.5-Lightning-* | `nemotron_h` | yes |
| Nemotron-Omni-Nano-* | `nemotron_h` | yes |

`LLMModelFactory` registers only `"nemotron_h"`, so the Audex bundles cannot load at all — they never reach a chat template, and this PR neither helps nor hurts them. That is a pre-existing gap, separate from this change.

**Corrected scope of the template fix: the four Lightning bundles and the two Omni bundles** (the `nemotron_h` ones that actually load and take the tools-branch fallback). The Lightning proof in the comments above stands unchanged.

Issue #2343, opened 2026-08-11
Backgrounded chat run holds the local-model lock with no reachable way to cancel it


jjang-ai:
## What happens

Close a chat window while a local model is generating. By design
(`ChatWindowManager.windowWillClose`) the session detaches into the background
registry and execution continues — the comment says "progress surfaces in the
notch".

The run then holds the **single local-model lock**. Every new chat window is
refused with:

> **A local model is already running**
> Only one local model can run at a time, and another chat window is using it
> right now. Wait for that reply to finish, or switch this chat to a remote
> model.

But the window that owned it is gone, so there is nothing to wait *at* and
nothing to stop.

## Observed

Muse Glimmer 30B JANG_4M, asked a question whose image failed to attach, so it
reasoned toward its token ceiling. After closing the window the process held
~27-35% CPU for **8+ minutes** and never released the lock. Restarting the app
was the only way to clear it.

## Why there is no way out

- `BackgroundTaskManager.cancelTask(_:)` exists and works.
- `BackgroundTaskToastView` is fully built, including a "Cancel Background Task?"
  confirmation that calls `cancelTask`.
- **That view has zero references anywhere outside its own file and its
  SwiftUI preview** — it is never presented.
- There is no `reattach` path either: `windowWillClose` detaches, and nothing
  re-binds a detached session to a new window. `activeTaskSessions()` is read
  in `ChatWindowManager` but not to restore one.

So the cancel capability exists in the API and in a built view, and neither is
reachable from the running app.

## Repro

1. Start a long local generation (a reasoning model on an open-ended prompt).
2. Close the chat window mid-stream.
3. Open a new chat window and try to send — refused with the alert above.
4. No UI anywhere offers to stop the detached run.

## The decision this needs

Whether a backgrounded run *should* keep the exclusive local-model lock is a
product call, which is why this is filed rather than patched:

- if yes, the cancel affordance has to be reachable — presenting
  `BackgroundTaskToastView` needs a host for custom toast views, since
  `ToastManager.show` takes a `Toast` model rather than an arbitrary view;
- if no, `windowWillClose` should cancel rather than detach when the run holds
  the local-model lock, and keep detaching for remote-model runs.

Either way the alert text should stop saying "wait for that reply to finish"
when the window that owns the reply no longer exists.

jjang-ai:
## The reachable half is fixed and merged (#2363)

[...]

PR #2363, opened 2026-08-12
Give the orphaned local-model lock a way out

jjang-ai:
Fixes the reachable half of #2343.

## The trap

Closing a chat window mid-generation detaches the run by design, and the detached run keeps the **single local-model slot**. Every later send is then refused with:

> **A local model is already running**
> …another chat window is using it right now. **Wait for that reply to finish**, or switch this chat to a remote model.

But the window that owned the reply is gone. There is nothing to wait at and nothing to stop. The reporter's only way out was restarting the app, after 8+ minutes at 27–35 % CPU.

## The capability already existed

`BackgroundTaskManager.cancelTask` and `openTaskWindow` are both public and both work. What was missing was any path from the point where the user is blocked to either one. (The issue notes `BackgroundTaskToastView` is built but never presented; it also assumed there was no reattach path — `openTaskWindow` is one.)

- `detachedTaskStreamingLocalModel(excludingSession:)` returns the **id** of the run holding the slot, where the existing predicate returned only a bool. A caller that has just refused the user's send needs to *act* on that run, not merely know it exists.
- The alert now tells the truth when the owner is detached, and offers **Stop it** / **Reopen it** / **Not now**. When a visible window owns the slot, the original wording and single OK are unchanged — "wait for that reply" is accurate there.

## Scope

Deliberately does **not** change whether a backgrounded run keeps the lock. #2343 leaves that open as a product decision; this is useful either way, and the issue itself says the alert text should stop saying "wait" regardless.

## Live verification

Reproduced the issue exactly on the rebuilt app with `Ornith-1.0-9B-JANG_4M`:

1. Started a long generation, closed the window mid-stream — run kept going at **117 % CPU**, orphaned.
2. Opened a new window and sent — alert appeared with the new wording and three buttons.
3. Pressed **Stop it** — process went **53.6 % → 0.2 % CPU**, alert dismissed.
4. Sent again — went through immediately (**30 tokens, TTFT 0.33s**) instead of being refused.


jjang-ai:
## Both buttons proven live, plus the localization gap CI caught

### Reopen it

Repeated the repro on the built app (`Ornith-1.0-9B-JANG_4M`): started a long generation, closed the window mid-stream (run kept going at **108.8 % CPU**, orphaned), opened a new window, sent — alert appeared. Pressed **Reopen it**:

- window count went **1 → 2**
- the restored window shows the run still streaming, mid-essay:
  > The bicycle's story begins not with a two-wheeled vehicle, but with a man's desperate attempt to outrun a horse. In 1817, German professor Count von Drais published a pamphlet describing his invention: the Laufmaschine…
- process still at **112.8 % CPU** — the run was reattached, not restarted or killed

### Stop it (from the earlier comment)

**53.6 % → 0.2 % CPU**, alert dismissed, next send went through immediately (30 tokens, TTFT 0.33s).

### Localization

The first CI run failed on `Check localizations` — `Stop it` and `Reopen it` were missing from the catalog, my omission (`Not now` and the message body already existed). Fixed rather than re-run:

| key | de | ko | ru | zh-Hans | zh-Hant |
|---|---|---|---|---|---|
| Stop it | Anhalten | 중지 | Остановить | 停止 | 停止 |
| Reopen it | Erneut öffnen | 다시 열기 | Открыть снова | 重新打开 | 重新開啟 |

Spliced in textually — **70 added lines, 0 deletions**. Re-serializing the catalog through `json.dump` rewrites all 6808 keys and buries the change in a 43,000-line diff.

`bash scripts/i18n/check.sh` (the exact CI step) exits 0, and the strings are confirmed in the **built bundle**, not just the catalog:

```
plutil -extract "Stop it" raw -o - …/ko.lproj/Localizable.strings  →  중지
                                    …/ru.lproj/…                  →  Остановить
```

Issue #264, opened 2025-12-22
Cannot add Gemini models as remote provider.


jarodise:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest `main` or release

### Describe the bug

tried multiple times to add Google's Gemini models as remote provider, didn't work. 

### Logs

```shell

```

Issue #255, opened 2025-12-21
Memory Issue


Pengy-X:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest `main` or release

### Describe the bug

Whenever I use a model, regardless of the size of model, each request (either through Osaurus built-in chat window or an external application like BoltAI) causes the model to load an additional time. I tested with Qwen3 4B 2507 (around 2gb) and after each question, Osaurus' memory usage would multiply in size. For larger models like qwen3 30b, Osaurus consumed 40gb+ ram after just two messages and MacOS started using 6gb+ swap. Hitting "unload model" or "clear all" did not unload the model, and the only way to unload models was to quit Osaurus.

### Osaurus version / commit

v0.5.54

### macOS version

26.2

### Apple Silicon chip

M4 Pro 48gb

### Xcode version

26.2

### Logs

```shell

```

tpae:
hey @Pengy-X , can you give it another test? it should be better now.

Pengy-X:
I see, "unload" does properly clear the model, but then does the model need to be loaded again for a new request? I'm not sure how LM Studio is handling KV cache, but it seems to grow in smaller steps than osaurus, and the kv cache isn't quantized. I tried setting osaurus kv cache bits to 8192, then 4096, but the memory usage was still quite high

tpae:
Hey @Pengy-X , i made some significant performance updates, and do you mind testing out the memory again?

zlatent:
Hi @tpae and team 👋

Just checking in on the status of this memory issue. I noticed there have been several related PRs merged; Is there any update on whether the memory buildup has been resolved in the latest version?

Really appreciate all the work going into Osaurus. It's an amazing project and I'm excited to see it improve! 🙏

tpae:
Hey @lucasmendesdeliver , can you share some additional information? which model are you using? are you using the chat or the agent?

lucasmendesdeliver:
Also, when i try to go in a conversation with the foundation model, after the second prompt the app crashes

tpae:
> I used v0.10.6, and I noticed that the memory usage increases during usage like before, then returns to a "normal" value when idle, which seems to be a big improvement from before. The issue may be resolved?

Yes, I believe so. I made another memory update which I had to revert yesterday: #480 . this will be an additional improvement once it's ready.

lucasmendesdeliver:
[...]

PR #2386, opened 2026-08-13
Repin vmlx-swift to cf3f16de

jjang-ai:
Picks up 11 engine PRs merged today. Live-proven on a Release build driven through the real GUI.

## The headline: the DSV4 Flash runaway is gone

The reported symptom was DeepSeek V4 Flash on **Low** producing a 36k-character thought that never
closed, hitting the output-token cap with no usable answer. Same model, same rail, on this pin:

| | reported | this build |
|---|---|---|
| thought | **~36,000 chars**, `⚠ thinking didn't close` | **726 chars**, closed |
| tokens | 10,554, stopped at the cap | 229, natural stop |
| TTFT | 32.12s | **1.36s** |
| decode | 29.3 tok/s | **45.3 tok/s** |

The Low-effort answer is also correct — asked why a hybrid SSM cache cannot do an exact disk restore
without an N-1 recurrent seed, it explained that the state at step N is not self-contained because it
is the incremental output of a recurrent update depending on the previous hidden state.

## Live GUI session (Release build, ad-hoc signed, driven via System Events + Vision OCR)

| turn | shape | result |
|---|---|---|
| 1 | reasoning **High**, text | `vmlx-repin-green` exactly · thought 83 chars / 756ms · TTFT 2.43s · 37.9 tok/s |
| 2 | **High**, history recall | recalled `vmlx-repin-green` · thought 119 chars · **TTFT 1.52s** (down from 2.43s — prefix reuse) · 37.1 tok/s |
| 3 | **effort flipped High → Low mid-session**, technical prompt | thought 726 chars, closed · correct answer · TTFT 1.36s · 45.3 tok/s |

Reasoning effort was changed *in the running conversation* through the model popover, and the rail
was re-read after the click to confirm it actually flipped (`… JANG • High` → `… JANG • Low`) rather
than assuming the click landed.

Decode held **37.1–45.3 tok/s** across all three turns, above the 33–35 tok/s floor.

## What changed in the engine

- harmony `<|end|>` is no longer forced into the stop set — a gpt-oss bundle stops truncating at its
  analysis→final channel separator
- native Nemotron Audex 2B and 30B runtimes, including the added-token regex fix; previously any
  Audex send pinned a core at 99% in `icu::RegexMatcher::MatchAt` (65,536 `<speechcodec_N>` tokens
  each became a regex branch)
- Gemma4 suffixless MoE expert tensors normalize into `experts.switch_glu`
- JANG semantic-width quantization outranks stale per-path metadata; `vision_tower.pos_embed` is no
  longer treated as a language hidden anchor
- `Hub.Repo` / `Hub.RepoType` are `Sendable` for Swift 6 consumers
- native-MTP per-generation stats on `GenerateCompletionInfo`, so speculative-decode health no
  longer needs stderr parsing

## Pin hygiene

Updated in all six places the tripwire enumerates — `Package.swift`, `OsaurusCore/Package.resolved`,
both xcworkspace `Package.resolved` files, and the two hard-coded `expectedRevision` constants.
Verified no seventh site exists: the pbxproj carries no revision, and exactly seven files mention
the new SHA (the seventh being the gitignored `OsaurusEvals/Package.resolved`, refreshed locally so
this machine's build agrees).

`swift package resolve` clean · **BUILD SUCCEEDED** (Release) · vmlx at that revision is
`MLXLMCommonFocusedTests` **542/542**.

## Not covered by the GUI session

Tools, image input, and image+tools in one turn were exercised in the vmlx harness
(`BENCH_VL_VARIATING` on LFM2.5-VL: tool call emitted, two-image turn, verbatim turn-1 replay
returning `HIT disk 26/26 100% reused`), **not** through this GUI session. The GUI runs above cover
reasoning-effort change, multiturn cache reuse, coherency, correctness and speed.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "cf3f16dea86e33e3831315c61186d5dfa1260d0e"
+        "revision" : "19c7cc9b451c6a5680c839ad2c430d288a8b9547"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "a983c6fcda0cb1616807f47fa2de94ffaec7e8bf"
+        "revision" : "cf3f16dea86e33e3831315c61186d5dfa1260d0e"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "cf3f16dea86e33e3831315c61186d5dfa1260d0e"
+        "revision" : "19c7cc9b451c6a5680c839ad2c430d288a8b9547"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "a983c6fcda0cb1616807f47fa2de94ffaec7e8bf"
+        "revision" : "cf3f16dea86e33e3831315c61186d5dfa1260d0e"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.swift
+++ b/Packages/OsaurusCore/Package.swift
@@ -216,7 +216,7 @@ let package = Package(
         // spelling this bundle emits, so an offered tool is no longer dropped.
         .package(
             url: "https://github.com/osaurus-ai/vmlx-swift",
-            revision: "a983c6fcda0cb1616807f47fa2de94ffaec7e8bf"
+            revision: "cf3f16dea86e33e3831315c61186d5dfa1260d0e"
         ),
         // FluidAudio 0.14.3 added a breaking `language:` parameter to TTS
         // calls that osaurus's `TTSService` doesn't pass. Pinning to the
--- a/Packages/OsaurusCore/Package.swift
+++ b/Packages/OsaurusCore/Package.swift
@@ -216,7 +216,7 @@ let package = Package(
         // spelling this bundle emits, so an offered tool is no longer dropped.
         .package(
             url: "https://github.com/osaurus-ai/vmlx-swift",
-            revision: "cf3f16dea86e33e3831315c61186d5dfa1260d0e"
+            revision: "19c7cc9b451c6a5680c839ad2c430d288a8b9547"
         ),
         // FluidAudio 0.14.3 added a breaking `language:` parameter to TTS
         // calls that osaurus's `TTSService` doesn't pass. Pinning to the
--- a/Packages/OsaurusCore/Tests/Service/ImageGenerationBridgeContractTests.swift
+++ b/Packages/OsaurusCore/Tests/Service/ImageGenerationBridgeContractTests.swift
@@ -74,7 +74,7 @@ struct ImageGenerationBridgeContractTests {
             encoding: .utf8
         )
 
-        let expectedRevision = "cf3f16dea86e33e3831315c61186d5dfa1260d0e"
+        let expectedRevision = "19c7cc9b451c6a5680c839ad2c430d288a8b9547"
         #expect(packageSwift.contains(#"revision: "\#(expectedRevision)""#))
         #expect(packageResolved.contains(#""revision" : "\#(expectedRevision)""#))
         #expect(workspaceResolved.contains(#""revision" : "\#(expectedRevision)""#))
--- a/Packages/OsaurusCore/Tests/Service/ImageGenerationBridgeContractTests.swift
+++ b/Packages/OsaurusCore/Tests/Service/ImageGenerationBridgeContractTests.swift
@@ -74,7 +74,7 @@ struct ImageGenerationBridgeContractTests {
             encoding: .utf8
         )
 
-        let expectedRevision = "a983c6fcda0cb1616807f47fa2de94ffaec7e8bf"
+        let expectedRevision = "cf3f16dea86e33e3831315c61186d5dfa1260d0e"
         #expect(packageSwift.contains(#"revision: "\#(expectedRevision)""#))
         #expect(packageResolved.contains(#""revision" : "\#(expectedRevision)""#))
         #expect(workspaceResolved.contains(#""revision" : "\#(expectedRevision)""#))
--- a/Packages/OsaurusCore/Tests/Service/RuntimePolicySourceTests.swift
+++ b/Packages/OsaurusCore/Tests/Service/RuntimePolicySourceTests.swift
@@ -786,7 +786,7 @@ struct RuntimePolicySourceTests {
         // and both xcworkspace Package.resolved files. Miss one and a release
         // surface resolves a revision nobody proved. OsaurusEvals resolves
         // this manifest transitively and its local Package.resolved is ignored.
-        let expectedRuntimeHardenedRevision = "cf3f16dea86e33e3831315c61186d5dfa1260d0e"
+        let expectedRuntimeHardenedRevision = "19c7cc9b451c6a5680c839ad2c430d288a8b9547"
         let manifestRevision = try Self.vmlxPinRevision(in: manifest)
         let coreResolvedRevision = try Self.vmlxPinRevision(in: coreResolved)
         let workspaceRevision = try Self.vmlxPinRevision(in: workspaceResolved)
--- a/Packages/OsaurusCore/Tests/Service/RuntimePolicySourceTests.swift
+++ b/Packages/OsaurusCore/Tests/Service/RuntimePolicySourceTests.swift
@@ -786,7 +786,7 @@ struct RuntimePolicySourceTests {
         // and both xcworkspace Package.resolved files. Miss one and a release
         // surface resolves a revision nobody proved. OsaurusEvals resolves
         // this manifest transitively and its local Package.resolved is ignored.
-        let expectedRuntimeHardenedRevision = "a983c6fcda0cb1616807f47fa2de94ffaec7e8bf"
+        let expectedRuntimeHardenedRevision = "cf3f16dea86e33e3831315c61186d5dfa1260d0e"
         let manifestRevision = try Self.vmlxPinRevision(in: manifest)
         let coreResolvedRevision = try Self.vmlxPinRevision(in: coreResolved)
         let workspaceRevision = try Self.vmlxPinRevision(in: workspaceResolved)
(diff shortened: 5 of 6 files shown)


jjang-ai:
## Live GUI session — completed across the variating axes

Follow-up to the PR body: I went back and closed the axes I had listed as uncovered. Full session on
the Release build of this pin, driven through the real UI (System Events + window-ID capture + Vision
OCR), single conversation, **DeepSeek V4 Flash 0731 JANG**:

| # | turn shape | thought | result | TTFT | decode |
|---|---|---|---|---|---|
| 1 | reasoning **High**, text | 83 chars / 756ms | `vmlx-repin-green` exactly | 2.43s | 37.9 tok/s |
| 2 | **High**, history recall | 119 chars | recalled `vmlx-repin-green` | **1.52s** ↓ | 37.1 tok/s |
| 3 | **effort flipped High → Low mid-conversation**, technical prompt | **726 chars, closed** | correct SSM/N-1-seed explanation | 1.36s | 45.3 tok/s |
| 4 | **Low**, tools available, arithmetic | 498 chars | **48,408,847** — correct (7919 × 6113) | 1.39s | 34.0 tok/s |
| 5 | **turn 1 replayed verbatim, at the end** | 161 chars | `vmlx-repin-green` — identical | 1.71s | 35.5 tok/s |

**Turn 5 is the head-boundary proof**: after an effort flip, a long technical turn and a
tool-available turn, replaying turn 1's exact prompt still returns turn 1's exact answer.

TTFT drops 2.43s → 1.52s between turns 1 and 2 on a growing conversation, which is prefix reuse
doing its job. Decode stayed **34.0–45.3 tok/s** across every shape, above the 33–35 floor.

### Stated precisely, because it matters

- **Tools were available, not invoked.** Turn 4 ran with the Sandbox tool present and the model
  answered from reasoning rather than emitting a tool-call card. The arithmetic is correct, but this
  is *not* proof of tool-call round-tripping — that is proven in the vmlx harness
  (`BENCH_VL_VARIATING` on LFM2.5-VL emits a real `get_weather` call).
- **Media axis not run in this GUI session.** DSV4 Flash is a text-only tower, so image and
  image+tools cannot be exercised against it. Those axes are covered in the harness on LFM2.5-VL
  (image mid-conversation, two images in one turn, image+tools in the same turn, and a verbatim
  replay returning `HIT disk 26/26 100% reused`). Doing them live would require a model swap and a
  full ~100 GB reload.

jjang-ai:
## Variating-pattern coverage, re-run on the exact pinned revision

The earlier VL numbers were from a branch that predates the last several merges (`#255`–`#264`
touched `JangLoader`, `Evaluate`, `Tokenizer` and the bench). Re-ran the full pattern on
**`cf3f16de`** — the revision this PR actually pins — on **two quants** of the family:

| axis | `LFM2.5-VL-3B-MXFP8` | `LFM2.5-VL-3B-JANG_4M` |
|---|---|---|
| reasoning ON → OFF mid-conversation | ✅ turn 1 think, turn 2 nothink | ✅ |
| **media** text → **image** → text-after-image | ✅ `"A red-to-blue gradient."` | ✅ |
| tools none → offered → **called** | ✅ **`get_weather(location:"Tokyo")`** | ✅ `toolCalls=1` |
| **image AND tools in the same turn** | exercised, no call (see below) | same |
| two images in one turn | ✅ answered | ✅ answered |
| **cache probed before every turn** (`matched/total` + tier) | ✅ all 10 turns | ✅ |
| **earlier turn replayed verbatim at the end** | ✅ **`HIT disk 26/26 (100% reused)`** → `"Paris."` | ✅ 26/26 → `"Paris"` |
| growing-chat control | `HIT disk 22/38 (57% reused)` | same |

`final cache stats: paged{hits=0,misses=0} disk{hits=3,misses=74,stores=17}` — both **passed**.

That closes every axis I could not reach through the GUI: media, image+tools, tool round-tripping,
per-turn cache probes, and the verbatim replay, all on the pinned code and across two quants.

### The one honest negative

**image + tools in the same turn produces no tool call**, under either phrasing — the model replies
"I cannot provide the weather … I do not have access to a tool". This is **not** a wiring bug: the
render probe shows the tools block does reach the prompt when an image is present
(`image turn: with tools=655 tok, without tools=605, delta=50` — the same +50 delta as a text turn).
It reproduces identically on both quants and predates this pin, so it is model behaviour, tracked
separately rather than silently absorbed into a green row.

### GUI vs harness, stated plainly

The GUI session (previous comment) proved the DSV4 runaway fix, mid-conversation effort flip,
multiturn reuse and verbatim replay on the real UI. The axes above are **harness** proof on the same
revision. I could not drive a model switch in the GUI to reach the image axis there: the model
popover renders but is not AX-addressable, `osascript` assistive access flapped intermittently
(`-25211`) throughout, and the chat window drifts on any stray chrome click. Not claiming GUI
coverage for those.

jjang-ai:
## Correction + deeper root cause on the image+tools negative

I called that "model behaviour" in the previous comment. I went and dug, and that was too generous
— there is a **structural cause on our side**, though the fix is not the obvious one.

`Qwen2VLMessageGenerator.generate(message:)` (`MLXVLM/Models/Qwen2VL.swift:914`) builds the content
array **text first, then media**:

```swift
dict["content"] = [["type": "text", "text": message.content]]
    + message.images.map { _ in ["type": "image"] }
```

Chat templates render that array verbatim, so on an image-bearing turn the prompt tail is

```
…get the weather in Tokyo with the tool.<image><image>…×21<|im_end|>
<|im_start|>assistant
```

The instruction is separated from the generation rail by a wall of placeholders. The turn that
**does** call the tool (image in *history*, text-only current turn) ends
`…Use the tool.<|im_end|>\n<|im_start|>assistant`, with the instruction adjacent to the rail. Both
prompts carry the schema (`hasToolSchema=true`), which rules out wiring. Every HF model card for
this family uses the opposite order, `[{"type":"image"},{"type":"text",…}]`.

### I tried media-first and reverted it

It measurably shifts attention — the reply changes from *"I do not have access to a tool…"* to
*"I used the tool to check the weather, but it only provides a color gradient"* — so ordering really
does control whether the schema is attended. But it still produced **zero** structured calls, and it
perturbed unrelated answers (image description `red-to-blue` → `orange…deep blue` on a red gradient;
two-image count flipped `0` → `1`).

That generator feeds **seven** VL families (Qwen2VL, Qwen25VL, Gemma3, SmolVLM2, Zaya1VL,
NemotronHOmni, LFM2VL). A prompt-shape change that does not fix the target defect is not worth that
blast radius on one model's evidence, so it is **not** in this PR — reverted, with the reproduction
and the rejected candidate written down so it can be picked up deliberately rather than rediscovered.

**Nothing in this PR changes because of it**: the defect predates the pin and reproduces identically
on both quants before and after. The repin remains a clean pick-up of the 11 engine merges.

jjang-ai:
## Pin advanced to `19c7cc9b` — adds a real tool-calling fix

While closing out the variating-turn axes I found a genuine bug and fixed it, so this PR now picks
it up too (vmlx-swift#265).

**Offering tools to ZAYA1-VL-8B silently swapped it onto Gemma-4's chat template.** The fallback
selected on `bosToken == "<bos>"`, and Zaya shares that bos while being a ChatML model
(`<bos><|im_start|>system …`). Its prompt therefore arrived as `<|turn>model … <turn|>`; the model
echoed the alien format back as literal text (`<|turn>user`) and emitted **no tool call at all** —
on both shipped quants, and only ever on tool turns.

Routing now keys on Gemma-4's own sentinels, which separate the families cleanly where `<bos>`
cannot:

| bundle | bos | `<\|tool_call>` | `<\|turn>` |
|---|---|---|---|
| `gemma-4-12B-it-MXFP8` | `<bos>` | ✅ | ✅ |
| `gemma-4-26B-A4B-it-qat-JANG_4M` | `<bos>` | ✅ | ✅ |
| `ZAYA1-VL-8B-JANGTQ_K` | `<bos>` | ❌ | ❌ |

Verified live with `VMLX_CHAT_TEMPLATE_FALLBACK_LOG=1`: Zaya logged
`tools -> Gemma4WithTools fallback engaged` before and logs no fallback after, while
**gemma-4-12B still engages it and still passes the whole variating pattern — `toolCalls=1` on the
tools turn AND on both image+tools turns.**

That Gemma-4 run also settles a question I had left open in an earlier comment: image+tools in one
turn is **not** a systemic limitation — Gemma-4 does it fine. LFM2.5-VL-3B's failure on that shape
is a model capability limit, now backed by a positive control rather than by absence of evidence.

Scope stated plainly: this fixes the **misrouting**. Zaya still does not emit structured calls with
its own template (it echoes its native `<tools>` block instead) — a separate template-priming issue
for that family, not claimed as fixed.

`swift package resolve` clean at `19c7cc9b` · **BUILD SUCCEEDED** (Release) · vmlx focused suite
**542/542** at that revision.

jjang-ai:
## Live GUI proof re-run on the advanced pin (`19c7cc9b`)

The earlier GUI session ran against the `cf3f16de` build, so it did not cover the routing fix.
Rebuilt, re-signed, relaunched and re-drove the UI — confirming first that the executable inside the
bundle was actually relinked (`13:04:47`, 18 compile/link steps) rather than trusting the bundle
directory's stale mtime.

| turn | shape | thought | result |
|---|---|---|---|
| 1 | reasoning **Low**, text | 78 chars, closed | `repin-19c7cc9b-green` — exact |
| 2 | **Low**, history recall | 108 chars, closed | `repin-19c7cc9b-green` — recalled correctly |

Correctness holds on the new pin: exact-phrase compliance, bounded reasoning that closes, and
history recall across turns.

### Timings from this run are NOT attributable — stated deliberately

TTFT was 8.81s then 27.98s, at 34.9 then 24.0 tok/s. I am **not** reporting those as a regression,
and I am not reporting them as healthy either, because the machine was memory-exhausted when they
were taken:

```
PhysMem: 124G used, 765M unused
swapusage: total 6144M, used 4857M, free 1287M
```

This session loaded DSV4-Flash (~100 GB resident) plus ZAYA1-VL-8B, gemma-4-12B and LFM2.5-VL, and
ran two full Release rebuilds. Under that pressure macOS pages and every latency number reflects the
paging, not the engine — the standing rule here is that speed claims require a hot-and-cool A/B, and
a single reading on a thrashing box satisfies neither arm.

The trustworthy speed evidence for this pin is the earlier GUI session on the same engine
(**TTFT 1.36–2.43s, 37.1–45.3 tok/s across three turns, decode above the 33–35 floor**) taken before
the machine filled up, plus the harness runs. Nothing in `#265` touches decode — it only selects a
chat template — so there is no mechanism by which it would move throughput.

jjang-ai:
## Timings resolved — engine is healthy; DSV4-Flash is a capacity limit

I said in the previous comment that the timings were not attributable. Rather than leave it there, I
freed the machine and re-measured, and the picture is now clean.

**Freed the box** (`quit` drops DSV4's ~100 GB): `124G used, 693M unused` → `27G used, 98G unused`.

**Re-ran on a clean machine, same build:**

| turn | TTFT | decode | thinking |
|---|---|---|---|
| 1 | **0.78s** | **46.6 tok/s** | 489ms / 75 chars |
| 2 | 26.43s | **3.3 tok/s** | 1m 0s |
| 3 (verbatim replay) | 22.87s | **3.3 tok/s** | 43s |

All three answered **correctly** (`clean-run-green`, recalled, replayed). But turns 2–3 collapse,
and the machine during that collapse reads `124G used, 693M unused`, swap 4.9/6.1 GB, **77.7 million
swapins**. DSV4-Flash is ~100 GB resident on a 128 GB box: turn 1 fits, and as the conversation and
KV cache grow it crosses the working set and every later turn is paging.

**This is not the engine.** Same build, same session, on a model that fits
(`LFM2.5-2.6B-MXFP8`, growing-chat harness):

```
Turn 1 cold:    ttft=0.050s  132.46 tok/s  finish=stop
Turn 2 growing: ttft=0.024s  136.24 tok/s  finish=stop   → passed
```

Sustained decode *increases* turn-over-turn and TTFT halves — cache reuse working exactly as
intended on the pinned revision. So: engine healthy, DSV4-Flash out of headroom on this hardware for
a growing conversation.

That also explains the earlier session's fast three-turn DSV4 run (1.36–2.43s TTFT, 37–45 tok/s):
it happened before the box filled with other models and rebuilds.

Nothing here changes the PR — `#265` selects a chat template and cannot touch throughput — but the
timings are now measured rather than caveated, and the machine is left clean at `27G used, 98G
unused`.

PR #2392, opened 2026-08-14
Persist the cancelled marker when Stop beats the first delta

jjang-ai:
## The vanishing interrupted turn

A user Stop that landed before the first streamed delta erased every record that a run happened. `finalizeRun` stamped the last assistant turn `cancelled` (the #2388 fix), but `completeRunCleanup`'s trailing-blank trim then dropped that same turn — blank content and no stats meant every keep-condition failed — so the persisted session ended on the user row with **no assistant row at all**. Fast models hit this consistently; slower ones persisted a truncated row instead, purely by timing.

A second window had the same outcome for a different reason: a Stop during the pre-send handshake (the "Loading Model..." wait) cancels the send before the run task ever appends its assistant turn, so there was nothing to stamp.

## The fix

- `trimTrailingEmptyAssistantTurn` keeps any turn carrying a terminal stop reason — the stamp IS the record.
- `stop()` appends the cancelled marker itself when it interrupts the handshake window.
- Source pins in `CancelStampSurvivesTrimTests` hold both mechanisms (suite green).

## Live proof (keychain-free proof app, isolated roots, Stop via the real AX button)

| scenario | before | after |
|---|---|---|
| Stop during cold model load | user row only — run invisible | `assistant / 0 chars / cancelled` persisted |
| Stop mid-stream | truncated row (from #2388) | truncated content + `cancelled` (11,463 chars in the proof run) |
| natural completion | `stop` | `stop` (unchanged; multiple control turns verified) |

The interrupted turn renders as the assistant header with an empty body — an honest record, no visual artifacts. DB rows verified per turn with `select rowid, seq, role, length(content), terminal_stop_reason, generation_token_count from turns`.

### DIFF
--- a/Packages/OsaurusCore/Managers/BackgroundTaskManager.swift
+++ b/Packages/OsaurusCore/Managers/BackgroundTaskManager.swift
@@ -683,7 +683,9 @@ public final class BackgroundTaskManager: ObservableObject {
             // it as part of the conversation history.
             state.chatSession?.appendInterruptMessage(trimmed)
         }
-        state.chatSession?.stop()
+        // Programmatic redirect, not the user's Stop button: the interrupt
+        // turn itself is the transcript record, so no cancelled marker.
+        state.chatSession?.stop(preservesCancelledMarker: false)
     }
 
     /// Emit a draft event to the originating plugin.
--- a/Packages/OsaurusCore/Tests/Chat/CancelStampSurvivesTrimTests.swift
+++ b/Packages/OsaurusCore/Tests/Chat/CancelStampSurvivesTrimTests.swift
@@ -65,9 +65,10 @@ struct CancelStampSurvivesTrimTests {
         let stopStart = try #require(
             chatView.range(of: "func stop(preservesCancelledMarker: Bool = true) {"))
         let stopWindow = String(
-            chatView[stopStart.lowerBound...].prefix(2900))
-        // Inside the `wasAwaitingPreSendHandshake` branch: the send was
-        // cancelled before the run task appended its assistant turn, so
+            chatView[stopStart.lowerBound...].prefix(3600))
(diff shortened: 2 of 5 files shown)


jjang-ai:
Post-merge live check on the rebuilt app (merged main, same proof root): the repinned runtime loads Qwen3.8 and generates with thinking live; closing the chat window mid-generation persisted the partial turn (4,491 chars of thinking survived — pre-#50 this record could vanish entirely) and the already-running recovery dialog surfaced correctly on the next send. Stated plainly per the proof policy: the Stop-BUTTON press itself was not re-observed live in this post-merge session — the machine was in active use, so the no-foregrounding constraint applied, and SwiftUI buttons do not respond to background-posted events (composer typing does; buttons and popovers do not). The Stop path's behavioral contract is carried by the 20 ChatSessionStopTests driving session.stop() — the exact method the button's action closure invokes — through real ChatSession + persistence, 3x locally and in the full CI run, plus the live Stop-button DB proof from the #2388 session on this same control. One observation for a possible follow-up, not a regression from this PR: the window-CLOSE cancellation path (as opposed to Stop) persists the partial turn with terminal_stop_reason NULL — the cancelled stamp currently covers stop()-driven paths only.

jjang-ai:
Live Stop-button proof CLOSED (supersedes the caveat above): found a background-safe press — raw AX with forced tree hydration (AXManualAccessibility) + kAXPressAction on the button frame, no focus steal. On the merged-main build, live UI, Qwen3.8 JANG_4D: pressed the real Stop pill mid-think, and the turn persisted as rowid 22 = assistant, empty content, 445 chars of partial thinking, terminal_stop_reason='cancelled' — the exact contract this PR ships, observed not inferred. The immediately following normal turn completed cleanly (rowid 24 = '144', 36 tokens, reason 'stop'), so the entry guard and late-stream refusal leave the normal path untouched live.

PR #2395, opened 2026-08-14
Constrain reasoning_effort to the bundle's declared set so Qwen3.8 cannot hard-fail the render

jjang-ai:
## Why

Qwen3.8-27B introduces a `reasoning_effort` chat-template kwarg with a CLOSED value set: `low` / `medium` / `xhigh` (default `xhigh`). The template `raise_exception`s on anything else — verified empirically against the real `Qwen/Qwen3.8-27B` template:

```
effort=xhigh    OK      (system line: "Reasoning effort is set to xhigh...")
effort=medium   OK      (no system line)
effort=low      OK      (system line: "...brief and focused...")
effort=high     RAISED  "Unexpected reasoning effort high. Supported types are xhigh (default), medium, and low."
effort=minimal  RAISED  same
absent          OK      resolves to xhigh
```

Osaurus forwarded the UI/API effort verbatim into the render context (`MLXBatchAdapter.additionalContext`, Qwen branch), so `reasoning_effort: high` — a valid value on most other models — hard-failed the prompt render. On top of that, local Qwen models only exposed a thinking toggle, so the new effort tiers were invisible in the picker.

## What

The JANG converter now stamps the accepted set into `jang_config.json`:

```
reasoning.supported_reasoning_efforts = ["low", "medium", "xhigh"]   (ascending)
reasoning.default_reasoning_effort    = "xhigh"
reasoning.reasoning_effort_transport  = "chat_template_kwarg"
reasoning.preserve_thinking_supported = true
reasoning.preserve_thinking_default   = true
reasoning.preserve_thinking_transport = "chat_template_kwarg"
```

This PR reads that contract end to end:

- **`DeclaredReasoningEffort`** (new): resolves the stamp — top-level `reasoning` or the earlier `chat.reasoning` nesting both accepted, so a stamper-side placement choice can never silently disable the constraint. Raw HF bundles (no jang_config) get a template-derived fallback that parses the accepted set out of the template's own closed-set raise guard. Cached per model id, invalidated with `LocalReasoningCapability`.
- **Dispatch snap**: known ladder values snap onto the declared set — `high`→`xhigh`, `max`→`xhigh`, `minimal`→`low`, ties round UP. Unknown strings still forward verbatim so the template's own typed raise (which names the valid set) surfaces instead of a silent coerce — same philosophy as the DSV4 policy. A stamped reasoning block WITHOUT the key means "no effort control" (Qwen3.6 stamps): the kwarg is omitted entirely. Audited every local bundle: no other family's template reads `reasoning_effort`, so the omission is render-identical everywhere else.
- **Picker constraint**: declared levels bridge into the existing `ModelReasoningCapabilities` pipeline, so local bundles get the same catalog-driven segmented control as ChatGPT/Codex models — the picker offers exactly None + `low`/`medium`/`xhigh` instead of the generic six-tier ladder, and `normalizedOptions` drops out-of-set persisted values before they reach the wire. `AgentReasoningPolicy` already yields to segmented-rail models (`usesReasoningEffortControl` guard), so agent/tool runs keep the DSV4/Hy3 semantics.
- **`preserve_thinking`**: passes through from `modelOptions["preserveThinking"]` only when the bundle declares the kwarg; absent → template default (true) stays authoritative. Cache note: flipping it (and changing effort, which rewrites the FIRST system line) re-keys the entire prefix.

## Proof

- 14 new tests in `DeclaredReasoningEffortTests` (stamp parse both nestings, transport downgrade, absence semantics, snapping table, template-derived parse of the verbatim 3.8 guard, capability bridge, and six dispatch-level assertions) — all green.
- Adjacent regression suites: `AgentReasoningPolicyTests`, `MLXBatchAdapterTests`, `ModelProfileRegistryTests`, `ModelPickerItemCacheTests`, `OrnithToolSelectionTests`, `RuntimePolicySourceTests` — 241 tests, all green.
- Template semantics verified against the real upstream `chat_template.jinja` via jinja2 (table above).
- **MERGE GATE — DO NOT MERGE YET**: live variating multiturn proof (effort switching low/medium/xhigh mid-session, None/direct rail, cache/TTFT check across the effort-induced system-line re-key, tools + reasoning crossed) runs against the JANG Qwen3.8 quants the moment they finish converting. Proof will be attached as a comment on this PR; merge only after it's green.

### DIFF
--- a/Packages/OsaurusCore/Models/Configuration/ModelOptions.swift
+++ b/Packages/OsaurusCore/Models/Configuration/ModelOptions.swift
@@ -225,9 +225,18 @@ enum ModelProfileRegistry {
     }
 
     /// Catalog-driven reasoning capabilities for a (full, provider-prefixed)
-    /// model id, when a connected provider published them.
+    /// model id, when a connected provider published them. Local bundles get
+    /// the same treatment from their stamped effort contract
+    /// (`jang_config.reasoning.supported_reasoning_efforts`, with a
+    /// template-derived fallback for raw HF bundles): Qwen3.8 accepts ONLY
+    /// low/medium/xhigh and its template raises on anything else, so the
+    /// picker must offer exactly the declared set rather than the generic
+    /// ladder.
     static func reasoningCapabilities(for modelId: String) -> ModelReasoningCapabilities? {
-        RemoteReasoningCapabilityCatalog.capabilities(for: modelId)
+        if let remote = RemoteReasoningCapabilityCatalog.capabilities(for: modelId) {
+            return remote
+        }
+        return DeclaredReasoningEffort.capabilities(forModelId: modelId)
     }
 
     /// The effort id the UI should display as active: the explicit persisted
--- a/Packages/OsaurusCore/Services/DeclaredReasoningEffort.swift
+++ b/Packages/OsaurusCore/Services/DeclaredReasoningEffort.swift
@@ -1,0 +1,303 @@
+//
+//  DeclaredReasoningEffort.swift
+//  osaurus
+//
+//  Bundle-declared reasoning-effort contract for local models.
+//
+//  Qwen3.8 introduced a `reasoning_effort` CHAT-TEMPLATE kwarg whose value set
+//  is closed: the template `raise_exception`s on anything outside
+//  low/medium/xhigh. Forwarding the app's generic effort ladder verbatim
+//  (`high`, `max`) therefore hard-fails the prompt render. The JANG converter
+//  stamps the accepted set into `jang_config.json` so serving layers can
+//  constrain pickers and snap requests without hardcoding per-release lists:
+//
+//      "reasoning": {
+//        "supported_reasoning_efforts": ["low", "medium", "xhigh"],   // ascending
+//        "default_reasoning_effort":    "xhigh",
+//        "reasoning_effort_transport":  "chat_template_kwarg",
+//        "preserve_thinking_supported": true,
+//        "preserve_thinking_default":   true,
+//        "preserve_thinking_transport": "chat_template_kwarg"
+//      }
+//
+//  The block lives at the jang_config top level on new stamps and under
+//  `chat.reasoning` on earlier converter output (DSV4-Flash); both nestings
+//  are accepted here so a stamper-side placement choice can never silently
+//  disable the constraint.
+//
+//  Absence semantics are deliberate and directional: a reasoning block WITHOUT
+//  `supported_reasoning_efforts` means the model has NO effort control (Qwen3.6
+//  stamps stay keyless because that template ignores the kwarg entirely) — not
+//  "unconstrained, send anything". Verified 2026-08-14: no other local bundle's
+//  template reads `reasoning_effort`, so omitting the kwarg for keyless stamped
+//  bundles is render-identical everywhere it applies.
+//
+//  Raw HF bundles carry no jang_config, so a template-derived fallback parses
+//  the accepted set straight out of the template's own closed-set guard
+//  (`… reasoning_effort … not in ('xhigh', 'medium', 'low')`) — the exact
+//  construct that makes the constraint necessary in the first place.
+//
+
+import Foundation
+
+enum DeclaredReasoningEffort {
+    /// What the bundle says about the `reasoning_effort` kwarg.
+    enum Control: Equatable, Sendable {
+        /// Closed accepted set (bundle order, stamped ascending) plus the
+        /// publisher default the template resolves when the kwarg is absent.
+        case levels([String], defaultLevel: String?)
+        /// The model has no effort tiers: never send the kwarg. Also the
+        /// honest downgrade when a declared transport is one this app does
+        /// not know how to deliver.
+        case noEffortControl
+    }
+
+    /// Declared support for the `preserve_thinking` chat-template kwarg
+    /// (Qwen3.8: keeps historical `<think>` blocks in the prompt; default
+    /// true). Presence of this value means the kwarg may be sent.
+    struct PreserveThinking: Equatable, Sendable {
+        /// The template's resolved default when the kwarg is absent, when the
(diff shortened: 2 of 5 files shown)


jjang-ai:
## Live proof — Qwen3.8-27B JANG quants, isolated proof app (this branch)

Bundles: all four JANG 3.8 quants (MXFP8 / 6D / 4D / 2D) verified before launch — stamp keys byte-match the contract (top-level `reasoning`, all six keys, `["low","medium","xhigh"]` / `xhigh` / `chat_template_kwarg` / preserve_thinking true/true/kwarg), and every safetensors shard's header-declared length matches the file on disk.

Live session (keychain-free proof app, fresh root, `Qwen3.8 27B JANG_4D`):

- **Render**: multiple completed turns at the declared default and after effort switches — the closed-set template renders with no `raise_exception`. (Engine-side: the vendored Jinja evaluates tuple literals as arrays, so the `not in ('xhigh','medium','low')` guard itself is supported — verified in the interpreter source.)
- **Picker constraint (the fix, visually)**: the model options Effort row offers exactly **None | Light | Medium | Extra High** — High and Max are gone. The segment ids come solely from the stamped set (unit-pinned as `[none, low, medium, xhigh]`), so an out-of-set selection is impossible from the UI; the API path snaps `high`→`xhigh` (unit-pinned).
- **Chip suffix**: shows the declared default (`· Extra High`) untouched, and tracks live switches (`· Light` after switching) — the capability bridge driving both.
- **Reasoning split**: think card renders separately ("Thought for 1.7s · 148 chars") with thinking persisted to the `thinking` column (148 chars on the reasoning turn, 0 on direct turns). Reasoning on/off + effort switching manually exercised in the live UI and behaving.
- **Hybrid prefix cache + SSD, per-turn**: cold turn TTFT 8.24s → next turn **TTFT 1.00s** with ~3.3k of history (a full re-prefill would be multiple seconds) — reuse firing across turns with thinking on. Disk L2 store populated: 6 KV boundary rows (~3.0-3.3k tokens each) in `kv_v2` + **4 hybrid SSM companion snapshots** in `ssm_companion` — the GDN state store working alongside the KV rows. Decode 39.9 tok/s (short turn) / 28.0 tok/s (319-token turn) on the 27B.
- **MTP**: not touched by this PR. The 3.8 stamps carry a recommendation-only `vmlx_mtp_tuning.json` (`best_depth: 1`, deliberately no measurements), and Swift's `usableBestDepth` correctly refuses non-measured sidecars, so native MTP stays off by measurement policy — exactly the contract shipped in the MTP overhaul. MTP weights are preserved in all four bundles (`mtp_layers: 1`) for a future measured sweep.

Unit gate: 14 new tests in `DeclaredReasoningEffortTests` + 241 tests across the six adjacent suites (`AgentReasoningPolicy`, `MLXBatchAdapter`, `ModelProfileRegistry`, `ModelPickerItemCache`, `OrnithToolSelection`, `RuntimePolicySource`) — all green.

jjang-ai:
Post-merge addendum, same live session (same root, app restarted): the chat session, disk cache, and settings all survived an app restart on the identical test root. Two more turns after the restart at the switched (Light) effort: assistant answered with a 302-char think block at TTFT 0.83s on ~3.4k of history — reuse still firing after both a restart and an effort change. All 8 CI checks green before merge (test-core 27m47s, test-evals 23m24s, test-statspack 24m0s included); no admin override used. One honest gap: the image-attach paste in the live UI lost a race with concurrent clipboard use on the machine, so the VL turn degraded to text (model correctly reported no image attached — no hallucination); VL for this exact qwen3_5 runtime family was live-proven in the 3.6 campaign and the 3.8 vision tower is architecturally identical.

jjang-ai:
Final variating-matrix addendum (same live session, JANG_4D): (1) VL — image turn answered with a correct scene description (488-char think, TTFT 6.29s), and a later text-only turn correctly described that image from history (text-after-image, 238-char think, TTFT 5.54s). (2) Tools — mid-session tool loop fired cleanly after the image/effort variations: think -> tool call -> tool result -> synthesized answer (server port 1337; TTFT 6.06s), where earlier turns had made no tool calls. (3) Verbatim replay of an early user turn at the end of the session completed coherently through a fresh agent loop with a clean stop. (4) Disk cache grew 6 -> 15 KV boundary rows + 15 hybrid SSM companion snapshots across all shape changes. Plainly stated gaps: image+tools in the SAME turn was not exercised on 3.8 (known qwen3_5-family instruction-ordering limitation, documented from the 3.6 campaign); per-turn matched/total cache probes are not observable in this build (no log sink) so reuse evidence is TTFT deltas + store growth; JANG_4D is the live-proven quant while MXFP8/6D/2D are stamp- and integrity-verified on the identical runtime.

jjang-ai:
Variating-matrix CLOSED on all four Qwen3.8 quants via the model-agnostic harness pattern (BENCH_VL_VARIATING, RunBench built from vmlx-swift main 21408777 with the merged MTP + hybrid-boundary work):

Per quant (4D, 2D, 6D, MXFP8 — all **passed**): reasoning on->off mid-conversation; text -> image -> tools -> **image+tools in the SAME turn** (get_weather fired with the image attached, plus the direct variant); two-images-one-turn; coordinator probe before EVERY turn reporting matched/total + tier (grow-turn HIT disk 25/44 = 56% reused; **verbatim replay of turn 1 at the end: HIT disk 65/65 = 100% reused** with the same answer and think block on every quant); render probes confirm the tool-schema token delta is identical for image and text turns (245 tok). Native MTP reports 'off' under the measurement gate on every quant — the recommendation-only sidecar is honored, exactly as designed. Video smoke also passed on 4D (1080p clip through the vision tower: 'A color bar test pattern is shown.', TTFT 360ms after prepare).

Together with the earlier live-UI proof (effort switching, segmented picker constrained to the declared set, tool loop, VL turn, restart persistence), every axis of the variating checklist is now crossed for this family — nothing is claimed from a single shape.

PR #2399, opened 2026-08-14
Reserve answer room on API requests so thinking cannot eat the whole max_tokens

jjang-ai:
Fixes the Anarlog 'AI generation did not return any text' report: OpenAI-compatible clients read message.content only, and a reasoning model that spends the whole finite max_tokens inside its think block returns empty content. For .httpAPI requests the adapter arms vmlx's per-request reasoning ceiling (vmlx-swift #270) with an answer reserve (a third of the cap, clamped 160-2048; inert below 224), resolved through the same ReasoningBudget.arm discipline as the env path so non-reasoning bundles are untouched. Also carries vmlx-swift #271: Qwen-style '<think>' + newline primes were classified un-primed by the strict suffix check, leaving any ceiling silently inert on exactly the family that needed it. Repin updated at all six sites (tripwires included). Proof: live A/B on Qwen3.8 JANG_2D at max_tokens 300 — control produced 1,398 chars of reasoning and ZERO content (bug reproduced); budgeted run capped thinking at 714 chars and returned a correct 364-char transcript summary as content. 198 tests green across the policy, adapter, and tripwire suites; the live A/B ships as an env-gated local-only test.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "21408777dfe0389dfb6d16f592533de959f75acd"
+        "revision" : "08fde1d6aa812e948e68aba4b3b6b63f2e66d85e"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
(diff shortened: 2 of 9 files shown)


jjang-ai:
Post-merge variating proof (the standing rule: every change crossed against the full turn matrix, not one shape). Extended BENCH_VL_VARIATING with a BENCH_REQUESTED_REASONING_BUDGET axis (vmlx-swift #272, merged) and ran both arms on Qwen3.8 JANG_4D against the new vmlx main: (A) budget OFF — byte-identical to the pre-#270 baseline on every turn (reasoning/image/tools/image+tools/two-images/grow-reuse 56%/verbatim replay 100% disk hit), proving the new engine code is inert by default across all shapes; (B) budget 200 armed on EVERY turn — also byte-identical everywhere, because the 120ch reasoning turn stays under the ceiling and the contract is strict pass-through until the budget is spent; tool calls fired, image turns answered, cache probes and the replay hit unchanged. Together with the earlier live A/B (control: 1,398ch reasoning + 0 content = the reported bug; budgeted: forced close at 714ch + a correct 364ch summary in content), the fix is proven on the failing shape AND proven harmless on every other shape.

PR #2402, opened 2026-08-15
Persist before stop in window cleanup so a close during load cannot destroy the send

jjang-ai:
Fixes the worst subcase found in the window-close investigation (task #60): closing the window while the model is still loading destroyed the user's message with no persisted trace — stop()'s deliberate draft-restore rollback removes the user turn to put its text back in the composer, but the window is being torn down, so the restored draft dies and the close callback's save() bails on empty turns. cleanup() now saves before stop(), mirroring the save-first guard switchAgent/startNewChat already have; mid-stream closes keep their behavior via the post-cleanup save overwriting with the stamped turns. Two new tests pin it (pending-turn persistence through cleanup; empty-session cleanup stays inert, no id minted); 46 tests green across the new suite + ChatSessionStop + ChatSessionQueuedSend. The known reset_incomingWarmup cross-suite flake reproduces on the UNMODIFIED base when ChatWindowStateIdentityTests joins the process — pre-existing, verified by stash A/B. Live proof of the exact lost-message scenario to follow as a PR comment.

### DIFF
--- a/Packages/OsaurusCore/Managers/Chat/ChatWindowState.swift
+++ b/Packages/OsaurusCore/Managers/Chat/ChatWindowState.swift
@@ -249,6 +249,16 @@ final class ChatWindowState: ObservableObject {
         selectedDiscoveredAgent = nil
         selectedDiscoveredAgentProviderId = nil
         selectedRelayAgent = nil
+        // Persist BEFORE stop(), exactly like switchAgent/startNewChat do:
+        // stop() on a mid-prepare cancel takes the draft-restore rollback,
+        // which REMOVES the just-sent user turn to put its text back in the
+        // composer — but this window is being destroyed, so the restored
+        // draft dies with it and the close callback's later save() finds
+        // empty turns and bails. Verified live: closing during model load
+        // silently lost the user's message with no persisted trace. Saving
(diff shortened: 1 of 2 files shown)


jjang-ai:
Live proof on the dev build, three close-timing variants of the exact scenario that previously destroyed the send (verified lost pre-fix: send + close-during-load left ZERO rows and a fresh chat on reopen): (1) close DURING model load — the user turn persists AND the cancelled marker persists with it (rows: user + assistant '' terminal_stop_reason='cancelled'), so the transcript records both the message and that the run was killed — better than the minimum fix, because the post-cleanup save carries the stamp stop() leaves; (2) close right after a fast natural completion — clean persisted exchange, reason 'stop'; (3) close at t=25s on a longer prompt — the model had already completed (fast quant), persisted clean. Stated plainly: the exact mid-stream-close timing kept completing before the close landed on this fast model, so that precise shape wasn't captured live — its persistence path is the same post-cleanup save shown carrying stop()'s stamped state in variant 1, and the pre-existing mid-stream behavior (partial persists) was observed live earlier in the investigation (rowid 20, 4,491 think chars).

PR #2418, opened 2026-08-20
Settings: DFlash 2 drafter picker + staged-verify MTP repin (draft until live gate passes)

jjang-ai:
## What

1. **MTP section: DFlash 2 drafter row.** Choose Folder / Change / Remove, a download link to the published Qwen3.8-27B-DFlash2 drafter shown until one is selected, drafter metadata summary, inline validation (missing folder, non-drafter folder). Hint states the scope contract: applies to Qwen 3.8 only, a mismatched bundle is ignored, and when it fits it REPLACES the native MTP head. All strings in the localization catalog (5 locales); `scripts/i18n/check.sh` clean.
2. **Model picker rows gain an accessibility press.** Selection previously lived only in an `NSClickGestureRecognizer` on the custom cell, so VoiceOver/AX automation could browse but never choose a model. The cell now exposes an AX button (labeled with the model name) whose press runs the same `onSelect`.
3. **Repin vmlx** to `424e5d54` (staged-verify MTP + two crash fixes found by the live gate — see below) across all pin surfaces + the two pin-expectation tests.

## Live gate status (Eric's merge bar) — NOT yet met, DRAFT until it is

Proof runs in the isolated keychain-free app against the real model library:

- ✅ **Native MTP speedup live in chat**: user-flow model selection (via the new AX press), Qwen3.8-27B-JANG_4D + its measured `vmlx_mtp_tuning.json` → **TTFT 5.10s • 35.3 tok/s** on a real turn, coherent output (plain decode is ~17).
- ✅ Model selection persists across app restart.
- ✅ **dFlash-2 dispatches live** when the drafter is configured (trace: block cycles at `bs=8`, up to 6/7 drafts accepted, ~6.5 tokens per accepted block).
- ❌ **Crash on the follow-on request** when dFlash-2 + a warm cache coexist: `DFlash2TokenIterator.init` dies in a Release precondition during restore of the entry the dFlash-2 turn itself stored. Two layers already fixed in the vmlx pin (TopPSampler 3D top-k zero-size axis — unit-pinned; disk-metadata reads now fail open); one layer remains under investigation.
- ⏳ SSD reuse TTFT proof + coherence sweep for both engines blocked behind that crash.

The two fixed layers were only findable live: greedy benches never sample 3D rows, and unit suites never restore a stored entry into a fresh dFlash-2 iterator with the bundle's real sampling defaults.


jjang-ai:
Crash-hunt update (vmlx now pinned at b9517c80):

**Trigger narrowed to one precise scenario.** The second-turn crash is the drafter ATTENTION forward failing inside a withError scope and degrading to a 0-dim scalar (vmlx #123 degrade-instead-of-trap), which then dies in the causal conv. Conditions from the live traces: total context past the drafter's `sliding_window: 2048` (all five drafter layers are sliding; the rotating cache has wrapped) AND a fully-rejected block (`accepted=0`) whose rewind writes `offsetForDFlash2` directly into the wrapped ring. Unit repros pin every reduction that does NOT trigger it: sampled two-turn store/restore through direct construction and through `BatchEngine.generate`, paged on/off, ~2.8k restored prefixes, accepted=0 cycles included — all green. The app repro is 100% (turn 2 at ~3.1k context).

Also landed: the conv now clamps short-block taps to zeros (defense against the negative-bound slice class), and `VMLX_DFLASH2_TRACE=1` prints conv input shapes + init restore geometry — the instruments that isolated all of this from a Release binary that prints nothing on a precondition.

Next: fix the rejected-block rewind for wrapped rotating drafter caches (or surface the recorded MLX attention error and fall back to AR for the turn), then rerun the live gate end-to-end (speedups, SSD reuse TTFT, coherence, both engines).

jjang-ai:
Superseded by the rebuilt branch dflash2-picker-ui: same UI on a clean base from current main (the vmlx repins and the picker accessibility press merged separately as #2419/#2420, and this branch's catalog edit is a minimal 5-key addition instead of the stale-base rewrite here).

PR #2419, opened 2026-08-20
Repin vmlx to staged-verify MTP + crash fixes; picker rows gain an accessibility press

jjang-ai:
The proven stability slice, split out of #2418 so it can merge now:

**vmlx repin → merged main `22b214d9`** carrying:
- staged-verify native MTP as the automatic mode for capable hybrids — byte-identical output, 1.43× prose / 1.80× code measured on Qwen3.8-27B, **live-proven in this exact build**: 27B selected through the picker, real chat turn at 23.3 tok/s (plain ~17), coherent;
- fix for a **process-killing** TopPSampler crash on 3D logits with `top_k` set (every sampled dFlash-2 request died in iterator init; unit-pinned in vmlx);
- disk-cache metadata reads now **fail open** instead of crashing the app on a malformed stored entry;
- dFlash-2 causal-conv short-block clamp + trace diagnostics.

**Model picker accessibility**: rows selected models only via an `NSClickGestureRecognizer` on the custom cell, so VoiceOver/AX automation could browse but never choose a model. The cell now exposes an AX button (labeled with the model name) running the same `onSelect`. Selection works and persists across restart — proven live.

NOT in this PR (stays in draft #2418 until the remaining dFlash-2 wrapped-ring rewind crash is fixed): the dFlash-2 drafter picker UI + download link.

Proof: cross-version probe on the vmlx side byte-identical to prior main (`b4b9320481f8233f`); MTP focused 82/82; all DFlash2 suites green; live turn screenshot + session row (`selected_model = JANGQ-AI/Qwen3.8-27B-JANG_4D`) in the isolated keychain-free proof app.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "08fde1d6aa812e948e68aba4b3b6b63f2e66d85e"
+        "revision" : "22b214d9be65701232d59362ea57c4437ba3f667"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "08fde1d6aa812e948e68aba4b3b6b63f2e66d85e"
+        "revision" : "22b214d9be65701232d59362ea57c4437ba3f667"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.swift
+++ b/Packages/OsaurusCore/Package.swift
@@ -216,7 +216,7 @@ let package = Package(
         // spelling this bundle emits, so an offered tool is no longer dropped.
         .package(
             url: "https://github.com/osaurus-ai/vmlx-swift",
-            revision: "08fde1d6aa812e948e68aba4b3b6b63f2e66d85e"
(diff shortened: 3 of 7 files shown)


jjang-ai:
## Live proof (final, in-app, this exact build)

All runs: isolated keychain-free proof app, real model library (`~/models`), Qwen3.8-27B-JANG_4D selected through the model picker via the new accessibility press, background AX driving, PhysMem verified settled before each leg, same code-generation prompt, bundle sampling defaults (temp 1.0, top_p 0.95, top_k 20), thinking off.

| leg | MTP | in-chat stats line |
|---|---|---|
| A1 | off (server setting) | TTFT 1.31s • **24.7 tok/s** • 266 tok |
| A2 | auto, but tuning artifact unreadable → AR | TTFT 1.27s • **25.3 tok/s** • 226 tok |
| B1 | auto, artifact fixed → depth-3 staged MTP | TTFT 1.21s • **34.7 tok/s** • 215 tok |
| B2 | auto (app reset an attempted off back to auto) | TTFT 1.20s • **32.7 tok/s** • 491 tok |

**AR ≈ 24.7–25.3; MTP ≈ 32.7–34.7 — the historical 33–37 MTP norm, restored and attributed.** The resolved strategy for the B legs is `nativeMTP(depth: 3, verifierMode: "input_capture_staged")` (verified by replaying the app's exact `resolvedMTPLaunch` chain against the bundle: `launchMode=speculative`, reason `best_depth=3`).

## Two findings written down so they never bite again

1. **`vmlx_mtp_tuning.json` must be wrapped in a `{"native_mtp": {...}}` envelope.** A bare tuning object decodes as nil, the bundle inspects as `speculative=off (tuning required)`, and auto-MTP silently never launches — every earlier ambiguous in-app number came from this. The corrected artifact for this bundle (validated, output-equivalent, best_depth 3, staged verifier, affine/4-bit, qwen3_5) is what produced the B legs, and the same envelope must ship in the HF bundle upload.
2. The A2 leg is the exact user-visible symptom of an unreadable artifact: "MTP auto" that quietly runs AR. The bundle-status line (`tuning=d3, speculative=on` vs `tuning required`) is the one-glance check.

Coherent full answers in every leg; app alive throughout; selection persisted across every relaunch.

PR #2420, opened 2026-08-20
Repin vmlx: DFlash 2 drafter failures contained — AR fallback instead of a host crash

jjang-ai:
Carries vmlx `5ef4ccf7` (vmlx#281): the last known dFlash-2 process-killer is fixed and contained.

**The crash** (100% reproducible live): second turn of a chat, context past the drafter's 2048-token sliding window, one fully-rejected block → the rotating drafter cache's rewind desynced its ring cursor → `[full] Negative dimensions not allowed` inside the drafter attention → degraded husk → host down on the next shape read.

**The fix**: span-correct ring rewind + every drafter stage propagates a degraded result instead of dying on it + the drafter forward runs in an error scope that logs the real error + the iterator disables speculation for the rest of the turn and continues autoregressively. A drafter failure can now only cost acceptance — never the turn, never the process, never correctness (the target verifies every token).

**Live proof** on the previously fatal flow (isolated keychain-free proof app, restored ~3k-token session): three consecutive turns, all complete, app alive, coherent answers — 35.3 / 32.0 / 60.2 tok/s one pass, 31.3 / 29.3 another, drafter errors logged and absorbed. All DFlash2 unit suites green; sampled two-turn store/restore repro suite green on the real bundle.

Six pin surfaces updated (Package.swift, three tracked Package.resolved, two pin-expectation tests).

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "22b214d9be65701232d59362ea57c4437ba3f667"
+        "revision" : "5ef4ccf737c0f37993781f117fe7e30f8ce0386c"
       }
     },
     {
--- a/Packages/OsaurusCore/Package.resolved
+++ b/Packages/OsaurusCore/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "22b214d9be65701232d59362ea57c4437ba3f667"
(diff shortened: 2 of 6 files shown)


jjang-ai:
## Final live proof — this exact pinned build (no local overrides)

Isolated keychain-free proof app, real model library, the exact previously-fatal session root (restored ~3k-token history, dFlash-2 drafter configured, bundle sampling defaults):

| turn | stats line | result |
|---|---|---|
| 1 | TTFT 1.14s • **35.3 tok/s** • 117 tok | coherent, alive |
| 2 (the 100%-fatal one) | TTFT 1.19s • **32.7 tok/s** • 122 tok | coherent, alive |
| 3 | TTFT 1.38s • **30.7 tok/s** • 302 tok | coherent, alive |

Six drafter-forward errors occurred across the run — every one logged with its real message and absorbed by the AR fallback; zero crashes, zero truncated turns. Before this pin, turn 2 killed the process in eight consecutive attempts.

Speeds are at the dFlash-2/MTP operating band (30.7–35.3; short-answer turns have hit 60.2 on this flow). The remaining drafter-error frequency past the sliding window costs acceptance on those cycles and is the next optimization item, not a stability item.

PR #2421, opened 2026-08-20
Settings: DFlash 2 drafter picker with download link and live folder validation

jjang-ai:
Adds the user-facing half of the DFlash 2 flow: a drafter picker in Settings → Server → Settings → Speculative Decoding. The engine half (dispatch, caching, containment) is already on main via the vmlx repins in #2419/#2420.

## What the card does

- **Choose Folder… / Change… / Remove** for the downloaded drafter bundle (AppKit `NSOpenPanel`, matching the rest of Settings).
- **Download drafter…** link to https://huggingface.co/JANGQ-AI/Qwen3.8-27B-DFlash2, shown only until a drafter is selected — the first-run path is "download, then choose folder".
- **Validation**: a missing folder or a folder without a DFlash 2 selector in its `config.json` shows a red error instead of being silently accepted; a valid bundle shows its summary (block size, tokens drafted per step, size on disk).
- **Hint** spells out scope honestly: the drafter applies to Qwen 3.8 only, a mismatched bundle is ignored, and when it fits it REPLACES the native MTP head.
- Settings search learns `dflash`, `drafter`, `block diffusion`; the five new strings ship with de/ko/ru/zh-Hans/zh-Hant entries (catalog diff is additions only, `scripts/i18n/check.sh` clean, 0 suspect literals).

## Live proof (isolated keychain-free proof app, built from this exact tree, driven background-only by AX)

Full settings flow exercised in the running app:

1. Selected state renders: **Qwen3.8-27B-DFlash2 — block 8 · 7 tokens drafted per step · 3.6 GB**, with Change…/Remove.
2. **Remove** → empty state: Choose Folder… + Download drafter… link + "None selected — speculation falls back to the model's own MTP head, per Mode above." Status bar correctly flags *Unsaved changes · Models reload after Save*.
3. **Choose Folder…** opens the panel titled "Select a downloaded DFlash 2 drafter folder"; picking a non-drafter folder (Documents) produced the red **"Not a DFlash 2 drafter — no selector in its config.json."** — the rejection path fires live.
4. Re-selecting the real drafter folder restored the card and the pane returned to *All changes saved*.

Engine regression on the same build (Qwen3.8-27B-JANG_4D + drafter selected): chat turn **TTFT 1.57s • 35.3 tok/s • 234 tokens**, dFlash-2 engaged (73 trace cycles), and when the known residual drafter error fired past the drafter's sliding window it was contained exactly as merged — "speculation disabled for the rest of the turn, continuing autoregressively" — zero crashes, app alive.


jjang-ai:
Follow-up: the download link in this PR pointed at JANGQ-AI/Qwen3.8-27B-DFlash2, which is not published (404). #2422 points it at the published incoai/Qwen3.8-27B-DFlash2 — the exact artifact the engine chain was validated against (local bundle config-identical, weights size matches).

PR #2436, opened 2026-08-21
Disk cache: auto-size to 10% of disk, surface usage in chat, add Clear SSD Cache

jjang-ai:
## Why

The disk-cache cap was a flat `diskMaxSizeGB ?? 10.0` — a literal, not derived from RAM, disk, or the model. For Qwen3.8-27B (64 layers, 4 KV heads, head_dim 256) KV costs **256 KiB/token at bf16**, so its 222k window needs **~54 GB**. The cap is also enforced across the *whole cache root*, not per model, so every bundle shares it.

A 10 GB cap therefore could not hold **one** full-context conversation of the flagship model. Past roughly 18% of a single window, every store had to evict earlier boundaries of the *same* conversation — so reuse collapsed exactly as context grew. That is the "speed drops as the chat gets longer" symptom, and nothing in the UI surfaced it.

For reference, oMLX defaults its SSD cache to 10% of disk capacity (`"auto"`), or `100GB` non-auto.

## What changed

**Auto sizing** (needs the companion vmlx change, see Blocked below). Unset resolves to 10% of the cache volume, floored at the old 10 GB. The floor is the safety property: resolving auto can only ever *raise* the cap, so no existing user loses cache, and a volume whose capacity cannot be read degrades to exactly the previous behaviour. It advises upward and never refuses — a bigger ceiling on a cache cannot block anything.

**Existing installs migrate too.** Auto only fills in for a nil value, so anyone who had already persisted `10.0` would have kept 10 GB forever after updating. `contractVersion` was declared and referenced nowhere else — a dead constant — so there was nothing to migrate against. It is now persisted as an optional `schemaVersion` and bumped to 2, with a version-gated one-shot. Deliberately *not* "rewrite any 10.0 we find": once migrated, a later deliberate choice of 10 GB is the user's and survives.

**Cache usage in the chat footer.** The Context Budget popover gains a Disk Cache row: used / cap, a fill bar, and past **75%** a warning naming both the consequence (checkpoints evicted, long chats re-prefill) and the remedy. Polls only while the popover is open.

**Clear SSD Cache** in Settings → Cache. Routes through `CacheCoordinator.clear()` per resident model, which takes `MLXDiskCacheIOLock` before deleting, so a purge cannot unlink a payload an in-flight restore is mid-read. It removes every `.safetensors` in the directory rather than only indexed rows — the one path that reclaims orphans left by a crash between the file write and the index insert.

## The bug this nearly shipped with

Both byte figures are **root-wide**, not per model: every model's `DiskCache` opens the same `cache_index.db` and sums `cache_entries` with **no `model_key` predicate**. Summing across models would have multiplied the displayed cache size by the number of loaded models — and looked entirely plausible while doing it. Aggregation takes `max` for the gauges and sums only evictions, which really is per instance. Pinned by a test with three models reporting the same total.

## Verified, not assumed

- **Cross-model eviction already works.** `quotaEntries()` has no `model_key` predicate and eviction sorts globally by priority then `createdAt` ascending, so a stale bundle's checkpoints are reclaimed before a hot bundle's. No change needed.
- **Chunked prefill is already correct for 27B.** `Qwen35.swift` chunks the pure-text causal path; image/video stay one-shot on purpose because mrope position ids derive from the full image grid. `prefillStepSize` is *not* dead — `prepare(windowSize:)`'s `windowSize` **is** `parameters.prefillStepSize`. Three suspected defects, all disproved by reading.
- **27B ships real MTP weights** — all four quants declare `mtp_num_hidden_layers = 1` and each carries 31 MTP tensors.

## Tests

- 12 vmlx tests (auto sizing + migration), including a post-migration deliberate 10 GB surviving.
- 9 osaurus tests (root-wide aggregation, 75% threshold, soft-cap overfull reported not clamped, formatting).
- Build clean, localization lint clean.

## Blocked

The auto default cannot reach the app until the companion vmlx-swift commits land on that remote and osaurus repins — this PR carries the UI, migration and plumbing only, so **Settings will still show 10 GB until the repin**. The bar is correct; the value behind it is stale.

### DIFF
--- a/Packages/OsaurusCore/Models/Chat/DiskCacheUsage.swift
+++ b/Packages/OsaurusCore/Models/Chat/DiskCacheUsage.swift
@@ -60,15 +60,18 @@ public struct DiskCacheUsage: Equatable, Sendable {
     public var usedLabel: String { Self.format(bytes: usedBytes) }
     public var maxLabel: String { Self.format(bytes: maxBytes) }
 
-    /// Shown once usage crosses the warn threshold. Names the consequence
-    /// (re-prefill) and the remedy (raise the size), because "cache is full" on
-    /// its own tells the user nothing actionable.
+    /// Shown once usage crosses the warn threshold.
+    ///
+    /// Deliberately free of engine vocabulary. "Prompt checkpoints", "evicted"
+    /// and "re-prefill" are precise for us and meaningless to the person
+    /// reading a chat window; the sentence has to say what they will notice
+    /// (long chats get slower to start) and what they can do about it.
     public var warningText: String {
         let pct = Int((usedFraction * 100).rounded())
         return
-            "Cache is \(pct)% full — older prompt checkpoints are being evicted, "
-            + "so long conversations may re-prefill instead of resuming. "
-            + "Raise Disk Cache Size in Settings."
+            "Cache is \(pct)% full — older conversation data is being removed, "
+            + "so long chats may need to re-read from the beginning and will be "
+            + "slower to start. Increase Disk Cache Size in Settings."
     }
 
     /// GB with one decimal above 1 GB, MB below it. Keeps the footer readable
--- a/Packages/OsaurusCore/Models/Chat/DiskCacheUsage.swift
+++ b/Packages/OsaurusCore/Models/Chat/DiskCacheUsage.swift
@@ -1,0 +1,75 @@
+//
+//  DiskCacheUsage.swift
+//  osaurus
+//
+//  Live on-SSD prompt-cache usage, rendered in the chat's Context Budget
+//  popover so cache pressure is visible rather than inferred from decode
+//  getting slower.
+//
+//  The cap used to be a flat 10 GB. For Qwen3.8-27B (64 layers, 4 KV heads,
+//  head_dim 256) KV costs 256 KiB per token at bf16, so a 222k window needs
+//  ~54 GB — a 10 GB cap could not hold ONE full-context conversation, and it is
+//  shared across every model in the cache root rather than being per bundle.
+//  Past that point each store evicted earlier boundaries of the SAME
+//  conversation, so reuse collapsed exactly as context grew, silently. This
+//  type is what makes that visible.
+//
+
+import Foundation
+
+/// A point-in-time reading of the shared on-disk KV cache.
+///
+/// Both byte figures are ROOT-WIDE, not per model: every model's `DiskCache`
+/// opens the same `cache_index.db` and sums `cache_entries` with no modelKey
+/// predicate. They are aggregated with `max` upstream, never summed.
+public struct DiskCacheUsage: Equatable, Sendable {
+    /// Payload bytes currently counted against the quota. Excludes the SQLite
+    /// index and WAL, matching how the quota itself is enforced — which is why
+    /// the cap is a soft cap and the Settings copy says so.
+    public let usedBytes: Int
+
+    /// The configured quota these bytes are measured against.
+    public let maxBytes: Int
+
+    /// Cache boundaries dropped by quota enforcement so far this process.
+    /// Non-zero is the observable proof the eviction janitor is running.
+    public let evictions: Int
+
+    public init(usedBytes: Int, maxBytes: Int, evictions: Int = 0) {
+        self.usedBytes = max(0, usedBytes)
+        self.maxBytes = max(0, maxBytes)
+        self.evictions = max(0, evictions)
+    }
+
+    /// Share of the quota in use, clamped to 0 when no quota is configured.
+    /// Deliberately NOT clamped at 1: the cap is a soft cap enforced after a
+    /// store, so a transient reading above 100% is real and worth showing
+    /// rather than hiding behind a clamp.
+    public var usedFraction: Double {
+        guard maxBytes > 0 else { return 0 }
+        return Double(usedBytes) / Double(maxBytes)
+    }
+
+    public var usedLabel: String { Self.format(bytes: usedBytes) }
+    public var maxLabel: String { Self.format(bytes: maxBytes) }
+
+    /// Shown once usage crosses the warn threshold. Names the consequence
+    /// (re-prefill) and the remedy (raise the size), because "cache is full" on
+    /// its own tells the user nothing actionable.
+    public var warningText: String {
+        let pct = Int((usedFraction * 100).rounded())
+        return
+            "Cache is \(pct)% full — older prompt checkpoints are being evicted, "
+            + "so long conversations may re-prefill instead of resuming. "
+            + "Raise Disk Cache Size in Settings."
+    }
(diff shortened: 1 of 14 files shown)


jjang-ai:
## Proof status — **do not merge yet**

Split by what each instrument can actually establish. A screenshot shows one number at one moment; it cannot show that eviction fires *as* a cache grows, that the oldest entry goes first, or that a changed cap is re-enforced. Those are sequence properties and need execution.

### Proven by execution — 7 tests, real files, nothing mocked

Real `DiskCache`, real directory, real safetensors payloads, real SQLite index, real quota pass (vmlx `DiskCacheQuotaEnforcementTests`, all passing):

| Claim | How it is proven |
|---|---|
| Janitor evicts as the cache grows | 12 × 80 KB into a 400 KB budget → resting size ≤ cap, `evictions > 0` |
| …and does not pass vacuously | also asserts the cache **grew** and was **not emptied** — a cache that silently refused every store would satisfy a cap check on its own |
| Oldest entry evicted first | writes separated past the index's second-resolution `createdAt`; entry 1 dies, entry 3 survives |
| **Changing the cap is enforced** | reopen the same directory with a smaller cap → trims on the next store |
| Raising the cap stops eviction | accumulates, zero evictions |
| **Clear reclaims orphans** | deletes an orphaned `.safetensors` the index never knew about — the crash-between-write-and-insert case quota accounting is structurally blind to |
| Quota is shared across models | two models, one root: model B's writes reclaim model A's older entry |
| 75% threshold on a real cache | under the threshold before, at or above it after |

Plus 12 vmlx tests (auto sizing + migration) and 9 osaurus tests (root-wide aggregation, threshold, soft-cap overfull, formatting).

### Proven visually — live app, window-scoped capture

Context Budget popover rendered:

> **DISK CACHE — 3.9 GB / 4.0 GB**, amber fill bar
> *"Cache is 98% full — older prompt checkpoints are being evicted, so long conversations may re-prefill instead of resuming. Raise Disk Cache Size in Settings"*

Row renders, bar fills, >75% warning fires with consequence-and-remedy wording.

**The live run found a real bug** that the tests could not: both numbers described the *default* cache directory while a different directory was configured. `OsaurusPaths.diskKVCacheUsageBytes()` hardcodes the default path, so any user with a custom Disk Cache Directory would have seen a plausible number about the wrong folder. Fixed in `a21d9207`.

### NOT proven — reason merge is blocked

- **Clear button UI wiring.** The action is tested; the button click is not.
- **Multiturn on Qwen3.8-27B.** A transparent full-screen overlay (layer 25, 1600×797) sits above the chat and swallows clicks on the model picker, so the instance cannot be moved off the auto-selected 0.6B.
- **Auto = 10% of disk, live.** Impossible until the companion vmlx commits reach their remote and osaurus repins — the running app still computes the old flat `?? 10.0`.

Merging is held until the first two are closed.

jjang-ai:
## LIVE PROOF COMPLETE — driven through the app UI via Codex computer-use

Isolated proof instance (`OSAURUS_TEST_ROOT`), local model **`Ornith 1.0 9B JANG_6M`**, disk cache cap set to **1.0 GB** so eviction is reachable in a few turns.

### Multiturn → cache fills → janitor evicts → warning fires

**5 of 5 replies completed.** Then the Context Budget popover reported:

> **`DISK CACHE  1011 MB / 1.0 GB`** + the full 99% warning sentence

Cross-checked against the filesystem immediately after:

| | |
|---|---|
| UI reading | `1011 MB / 1.0 GB` |
| Actual on disk | **1011 MB** — exact match |
| Cap | 1024 MB → **99%, under cap** |
| Stores issued | **6** |
| Surviving entries | **4** payload files, 4 index rows |
| **Evicted** | **2** |

So the janitor ran, held the cache under the cap, and the UI number is the real directory size rather than an estimate.

### Settings change + persistence + Clear button (earlier CU run, quoted verbatim)

- Placeholder: `"Blank = Auto (10% of disk)"`
- Changed the field and clicked Save → `"Settings saved. 1 loaded model(s) were unloaded and will reload with the new runtime policy on the next request."`, status `"All changes saved"`, field re-read as `"0.2"`
- `"Clear SSD Cache"` clicked → `"Cache was already empty"` (correct — the isolated cache was empty at that point)
- Plain-language copy confirmed live: `"Deletes all saved conversation data from disk, including leftover files from an interrupted write. Your chats are not affected — the next reply just takes a little longer to start."`

### Two defects found BY the live runs and fixed in this PR

1. **Wrong directory measured.** The footer read the default cache dir while a custom Disk Cache Directory was configured — a plausible number about the wrong folder. Fixed in `a21d9207`, pinned by a negative assertion.
2. **Sub-0.1 GB caps displayed rounded.** `%.1f` turned a saved `0.05` into `0.1`, showing a cap the user never set. Widened to `%.2f`; verified live afterwards as `0.05`.

### Also proven by execution (7 tests, real files, nothing mocked)

Oldest-first eviction, lowering the cap re-enforcing on the next store, raising it stopping eviction, `clear()` reclaiming orphaned safetensors, and quota shared across models. See the earlier comment.

**Ready to merge.**

jjang-ai:
## Post-fix live proof — 6-turn multiturn to the cap

Isolated proof root, local **Ornith 1.0 9B**, disk cache cap **1.0 GB**, app built against vmlx `dsv4-fail-closed` (now merged as `2fc5d7bb`).

**6 of 6 replies completed.** Sustained decode held flat across the run:

| | tok/s |
|---|---|
| reply 1 | **87.3** |
| reply 6 | **88.0** |

No decay as the conversation grew and the cache filled.

### Popover, quoted verbatim

> **`DISK CACHE  920 MB / 1.0 GB`**
>
> *"Cache is 90% full — older conversation data is being removed, so long chats may need to re-read from the beginning and will be slower to start. Increase Disk Cache Size in Settings."*

The plain-language rewrite is live, with no engine vocabulary.

### Cross-checked against the filesystem and the trace

| | |
|---|---|
| UI reading | `920 MB / 1.0 GB` |
| Actual on disk | **920 MB** — exact match |
| Cap | 1024 MB → **90%, under cap** |
| Stores issued | **7** |
| Surviving entries | **4** |
| **Evicted** | **3** |
| **Cache hits** | **11** |

So across six turns the janitor evicted 3 entries, held the cache under the cap, the warning fired at 90%, and the UI number is the real directory size.

**The 11 hits are the regression check that mattered.** vmlx#287 made damaged records fail closed; the risk was that it would start refusing healthy ones too. Healthy restores still hit, 11 times, while decode stayed flat at ~88 tok/s.

### One observation worth recording

At a **0.3 GB** cap the same run produced **zero stores** and an inert cache — a single boundary payload for a 9B exceeds that cap, so every store is skipped and there is nothing to evict. That is the same failure the old flat 10 GB default caused for a 27B, reproduced at small scale: a cap below one entry doesn't make the cache small, it makes it useless.

PR #2439, opened 2026-08-21
Repin vmlx to the disk-restore fail-closed fix (2fc5d7bb)

jjang-ai:
Brings in [vmlx-swift#287](https://github.com/osaurus-ai/vmlx-swift/pull/287). A damaged cache payload now **misses honestly** instead of decoding against an empty layer.

## What it fixes

Restore was failing **open**. With one layer's payload damaged it reported success while that layer stayed empty:

```
reported 4 tokens restored but layer offsets are [4, 0, 4, 4]
reported 4 tokens restored but layer offsets are [0, 4, 4, 4]
```

`totalTokens` is seeded from sibling layers, so the coordinator was told the entry hit and then decoded attention against an empty layer — coherent-but-wrong text, where an honest miss costs only a re-prefill. Three routes bypassed the rule the `.qkv` path already documents:

1. a **missing kind tag** was never visited by the deserialize loop, so a truncated record silently became a smaller well-formed one
2. `.kv` / `.qkv` / `.rotating` emitted `.skip` ("intentionally empty") for an unbuildable payload, where the kind tag promised content
3. `.tq` used `continue` where every sibling kind refuses the whole record

## Proven live on this branch's build

Isolated proof root, local **Ornith 1.0 9B**, cap **1.0 GB**, 6-turn multiturn:

| | |
|---|---|
| Replies | **6 / 6** |
| Decode, reply 1 → 6 | **87.3 → 88.0 tok/s** (flat) |
| Popover | **`920 MB / 1.0 GB`** + 90% warning |
| Actual on disk | **920 MB** — exact match |
| Stores → surviving | **7 → 4 = 3 evicted**, held under cap |
| **Cache hits** | **11** |

The **11 hits** are the regression check that mattered: a fix that refuses damaged records could have started refusing healthy ones. It doesn't — healthy restores still hit, and decode stayed flat as the cache filled.

## Repin hygiene

Seven files carried the pin, found with a **Python tree walk rather than `grep`** — `grep` on this machine is `ugrep`, which honours ignore files and is exactly how `Packages/OsaurusCore/Package.resolved` was missed on the previous repin.

OsaurusCore builds clean; the checkout the app builds against is `2fc5d7bb` with the fix present.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,400 +1,400 @@
 {
-  "originHash" : "a70de644fa9a976403022014284317c01c48d4d19f71b277a396f9c2786a3827",
-  "pins" : [
+  "originHash": "a70de644fa9a976403022014284317c01c48d4d19f71b277a396f9c2786a3827",
+  "pins": [
     {
-      "identity" : "aachartkit-swift",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/AAChartModel/AAChartKit-Swift.git",
-      "state" : {
-        "revision" : "b51e53c3b04372c1ed1906b0ec3ba45733953ef4",
-        "version" : "9.5.0"
+      "identity": "aachartkit-swift",
+      "kind": "remoteSourceControl",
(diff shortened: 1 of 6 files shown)


jjang-ai:
Superseded by #2441, which repins to 94e19942 — that contains this PR's target (2fc5d7bb, #287) plus #288 (two further fail-open holes in disk restore) and #289 (cap as a percent of disk). Merging both would collide on the same seven pin sites.

PR #2446, opened 2026-08-23
Turning native MTP off did not stay off

jjang-ai:
Reported: "we should check that toggling mtp native off for qwen actually works". It does not — and it is a settings-persistence bug, not an engine one.

## The defect

`ServerRuntimeSettingsStore` repairs installs that persisted the pre-`e095d0f` "MTP off" engine default:

```swift
if normalized.mtp.mode == .off,
   normalized.mtp.draftTokenLimit == nil,
   normalized.mtp.keepDraftCacheSeparate,
   normalized.mtp.acceptedTokensOnlyEnterBaseCache
{ normalized.mtp.mode = .auto }
```

That condition is **also** the exact shape of a user who just switched MTP off and changed nothing else. The repair ran on every `load()`, and `load()` persists what it changes — so "off" silently became "auto", permanently.

Five sibling repairs in the same file (diffusion, tied-head, cache, paged-cache, memory-safety) are one-shot behind a marker file. This one had **no marker at all**.

## The fix needs two gates

A marker alone is not enough: the **first** load still flips a current-schema user's choice. My initial attempt did exactly that, and the repeated-reload test caught it.

The real discriminator is whether the file was pre-migration — captured **before** `migrateToCurrentSchema()` overwrites `schemaVersion` and makes every install look current. The marker then stops the repair re-firing from the second load onward.

## Tests

Confirmed failing first — `"auto" is not equal to "off"` on reloads #1, #2 and #3. They run through the store's real `load()` path, because a test calling the normalizer directly would prove it correct rather than prove the user's setting survives.

A counter-case asserts a genuine legacy install (no `schemaVersion`) **still** gets repaired to `.auto`, so this cannot be "passed" by deleting the repair.

9/9 green across the settings-migration suites.

### DIFF
--- a/Packages/OsaurusCore/Models/Configuration/ServerRuntimeSettingsStore.swift
+++ b/Packages/OsaurusCore/Models/Configuration/ServerRuntimeSettingsStore.swift
@@ -364,17 +364,39 @@ public enum ServerRuntimeSettingsStore {
         // This is the right home for it. Both `load()` and `loadOrMigrate()`
         // funnel through here, and `load()` persists whenever normalization
         // changes the value, so the migration runs once and is written back.
+        // Captured BEFORE the migration, which overwrites `schemaVersion` and
+        // would otherwise make a legacy install indistinguishable from a
+        // current one by the time the MTP repair below runs.
+        let wasPreMigrationInstall =
+            settings.schemaVersion != VMLXServerRuntimeSettings.contractVersion
         normalized.migrateToCurrentSchema()
         // vmlx-swift e095d0f changed the engine default from "MTP off" to
         // "auto". Existing Osaurus installs persisted the old default exactly,
         // so without this repair tuned MXFP8/MTP bundles still never reach the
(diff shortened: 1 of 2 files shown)


jjang-ai:
## Live proof (not just unit tests)

Same procedure on both builds — let the app write its own complete `server-runtime.json`, patch **only** `mtp.mode` to `off`, relaunch, read the file back:

| build | `mtp.mode` after relaunch |
| --- | --- |
| control (fix absent — marker symbol count 0) | **auto** — the user's choice was discarded |
| fixed | **off**, and still `off` after a second relaunch |

Two harness errors on the way to that, both mine and both discarded rather than reported:

1. I hand-wrote `schemaVersion: 4`; the real contract version is **3**, so my file looked pre-migration and the repair fired correctly. That run proved nothing about the fix.
2. A partial JSON (`{schemaVersion, mtp}`) does not decode into the full settings type, so the app fell back to defaults and wrote `auto` — again nothing to do with the fix. The method above avoids both by patching the app's own 11-key file.

PR #2486, opened 2026-08-26
fix leaked pacing timer on mid-stream processor dealloc

donmimo:
## Problem

Closing a chat window while smooth-streaming deallocated `StreamingDeltaProcessor` without invalidating its 16ms repeating `pacingTimer`. The run loop retained the timer, which fired every 16ms forever, spawning a no-op `Task { @MainActor [weak self] }` on every tick — producing the constant main-thread task-spam observed in hang reports and memory exhaustion over long sessions.

## Fix

Add `isolated deinit` that invalidates `pacingTimer` and `flushTimer`. A regression test proves the timer is dead after mid-stream dealloc.

## Verification

- Regression test: fail → pass
- 59 focused tests: all pass
- Release build: succeeds
- Live proof: closed chat mid-stream, scrolled Settings → zero spam frames


tpae:
Thank you for your contribution @donmimo !

PR #2489, opened 2026-08-26
The Speculative Depth row rendered on models with no MTP head

jjang-ai:
Regression from #2473, caught by Eric on a Ling picker: `refreshNativeMTPState` filtered summaries with `nativeMTPStatus?.contains("mtp:")`, but vmlx's status line begins `mtp: <mode>, layers=N, tensors=N` for **every** inspected bundle — headless models read `mtp: none, layers=0, tensors=0` — so every resident model joined the sticky capable set and the Off/Auto/1/2/3 row rendered on models that cannot speculate at all.

Capability now requires the status to name a **real head**: a mode outside {none, unknown, metadata_only_missing_weights} AND a nonzero `tensors=` count. `NativeMTPRowCapabilityTests` pins both directions with the real status strings (tuning-suffix variants included) plus the zero-tensor head-implying mode.

It shipped because the #2473 ladder proved the row's PRESENCE on Qwen but never its ABSENCE on a headless model. Live absence/presence proof (Ling picker without the row, Qwen picker with it, same build) will be posted here before merge.

### DIFF
--- a/Packages/OsaurusCore/Tests/Service/NativeMTPRowCapabilityTests.swift
+++ b/Packages/OsaurusCore/Tests/Service/NativeMTPRowCapabilityTests.swift
@@ -1,0 +1,44 @@
+import Foundation
+import Testing
+
+@testable import OsaurusCore
+
+/// The depth row rendered on models with no MTP head because the capability
+/// filter matched the bare "mtp:" prefix every status line carries.
+@Suite("Speculative Depth row capability filter")
+struct NativeMTPRowCapabilityTests {
+    @Test("headless statuses are NOT capable")
+    func headlessRejected() {
+        #expect(
+            !FloatingInputCard.statusIndicatesNativeMTPHead(
+                "mtp: none, layers=0, tensors=0"))
+        #expect(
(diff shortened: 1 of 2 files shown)


jjang-ai:
Live absence/presence proof, dev-built app driven by pid via osascript/AX (never foregrounded), one binary carrying this fix, fresh isolated proof roots, a real turn answered first so the runtime published each bundle's summary before the picker was opened:

- `JANGQ-AI/Ling-3.0-tiny-JANG_6M` (no MTP head; status `mtp: none, layers=0, tensors=0`): picker opened, MODEL OPTIONS rendered with Thinking — **row_present=no, expected=absent → PASS**
- `JANGQ-AI/Qwen3.8-27B-JANG_4D` (real head + measured tuning): same flow — **row_present=yes, expected=present → PASS**

Unit side: NativeMTPRowCapabilityTests 3/3 (headless statuses rejected incl. unknown/metadata_only_missing_weights and the zero-tensor head-implying mode; real-head statuses accepted with and without tuning suffixes). Window-scoped screenshots of both pickers captured and kept locally.

PR #2505, opened 2026-08-28
Delete the persisted global Disable Tools switch; tool availability is agent-scoped

jjang-ai:
## Summary
Removes the chat-wide persisted `ChatConfiguration.disableTools` bit. Once written into `chat.json`, that single flag silently stripped tools from every surface — native Chat, session warm-up, and `/agents/{id}/run` — with no UI indication of why tools had vanished (the recent tools outage). Tool availability is now governed only by the per-agent Tools setting plus explicit request-scoped overrides.

## Changes
- `ChatConfiguration`: persisted `disableTools` deleted. Decoding ignores a stale `"disableTools"` key in existing `chat.json` files (Swift skips unknown keys) and re-encoding omits it.
- `HTTPHandler`: agent-backed requests no longer fold in the global bit — the target agent's dedicated setting governs.
- `AgentConfigSnapshot` / `SystemPromptComposer` / `ChatSessionWarmup` / `ChatView`: request-scoped suppression stays separable from the per-agent toggle; sandbox mode does not override explicit request suppression.
- Tests updated: configuration defaults, app-configuration migration, warm-up, stop, sandbox-override, plus the shared chat-history test storage helper.
- Two new AgentLoop eval suites (folder list-contents, PII detect) used in the packaged tool-loop proof.

## PROOF
- Root cause reproduced and documented: a seeded `"disableTools": true` in `chat.json` removed tool schemas from all three surfaces on the prior build.
- A packaged build with this change, seeded with the same stale key, made real tool calls including a forced multi-turn call (Ling), and a live Qwen 3.8 Flash Next chat session performed a natural `osaurus_config` tool loop with an image in the conversation.

## Known follow-ups (deliberately not in this PR)
- Cold-upgrade behavior: a stale `disableTools` key remains physically present in `chat.json` until the next settings save (decode ignores it; re-encode drops it). If immediate physical scrubbing on first launch is wanted, that is a migration-rewrite follow-up.
- The "tools worked, then stopped mid-conversation" report class is NOT fully closed by this change alone; empty-after-tool responses, natural-auto tool-choice misses, iteration-cap finalization, and small-context tool suppression are separate mechanisms still being mapped against the original transcript.
- Natural (non-forced) multi-turn tool loops across additional families on the final packaged build.


jjang-ai:
Design note on why deletion (rather than re-scoping) is correct, after auditing the original feature:

- Origin e1b49f8b6 added the flag so osaurus could act as a plain LLM backend for an external agent (byte-stable system prompt for KV reuse; caller manages its own tool routing).
- On current main the flag has NO writer: there is no Settings toggle and no code path that sets ChatConfiguration.disableTools. Only stale chat.json files written by older builds carry it.
- Its remaining readers are exactly the wrong surfaces: native ChatView, session warm-up, and agent-backed HTTP requests. The plain non-agent completions path — the original use case — does not read it at all on main.
- The use case is covered by the per-agent Tools toggle, which is also reachable through the declarative osaurus_config surface (DefaultAgentConfiguration.disableTools — per-agent, unchanged by this PR).

Net: the persisted global bit is an orphaned foot-gun whose only observable behavior is the tools outage. Deleting it removes logic without removing any reachable feature.

PR #2506, opened 2026-08-28
Speculative Depth buttons enforce a real manual MTP activation; greedy only while MTP runs

jjang-ai:
## Problem
The 1/2/3 depth buttons wrote `mtp.mode = auto` plus a draft-token limit. Auto without a measured `vmlx_mtp_tuning.json` resolves to no MTP, and the limit only clamps an already-approved tuned depth — so the UI displayed a selected depth while the engine stayed autoregressive on every Qwen 3.8 Flash Next bundle.

## Contract (explicit)
- **Off** disables. **Auto** stays tuning-gated (launches only from a measured, usable tuning artifact — 2L auto-launches d1 from its stamped measurement; 4M variants are stamped blocked-slower and stay off).
- **1/2/3** is a manual ACTIVATION: `mode = force_on` + `explicitDepth`. The engine validates per-bundle tensor evidence and family support and fails closed otherwise (JANG_1L: config claims a head, bundle has zero tensors — blocked). Never name-based; applies to every MTP-capable family (Flash-Next, Ornith 35B and Qwen3.8-27B — both `qwen3_5` — Nemotron-H).
- **Greedy (0/1/0/0) is enforced ONLY on requests that actually run native MTP** — manual or active Auto — scoped to that request. Everything else, including spawned agents and coding loops, samples per bundle `generation_config`; only the four sampler fields are ever overridden and all other generation parameters continue to apply.

## Changes
- FloatingInputCard: buttons write `forceOn`+`explicitDepth` (Off/Auto clear it); legacy `auto`+limit state renders as the depth it claimed.
- ModelRuntime.requestDraftStrategy honors the explicit depth exactly; `enforcingMTPGreedy` applies the engine's enforced sampler to `.nativeMTP` requests only.
- Repins vmlx-swift to `b52ddf18`: manual-depth activation contract (explicitDepth field, manualRecommendation policy, manualDepthRequest load-gate, LoadConfiguration threading) plus the JangPress auto-threshold exclusion for Flash-Next (whole-directory bytes were tripping cold-compression on bundles whose 20–29 GiB PLE table never becomes resident).

## PROOF
- Engine: NativeMTPManualDepthTests 6/6 (activation, tuning-gated Auto, JANG_1L fail-closed, unsupported family, bad depth, load gate) + NativeMTPDepth3AutoLaunchTests 4/4 (shipped Qwen3.6 depth-3 contract unaffected) + JangPress exclusion 2/2.
- Harness: measured tuning stamps — 2L d1 41.4 tok/s vs AR 39.7 with byte-exact greedy parity; 4M-aligned d1/d2/d3 27.0/27.1/25.2 vs AR 39.1 stamped blocked.
- Live dev app (background-driven, isolated test root, fresh session per leg, same prompt): manual d2 = 16.1 and 16.3 tok/s across two runs (the verify-dominated MTP-active fingerprint — the old build showed ~39 here because nothing launched); mode off = 39.3 tok/s matching the harness AR gate. Off-leg length/content differ per bundle sampling. Manual-leg hash variance across runs is prompt variance from the injected [Current Time] prefix, not sampler variance.

## Remaining (tracked)
- In-app UI-click leg (the Configuration popover is inaccessible to background AX automation; the settings→engine chain is proven — one human click covers the last inch) and MTP-row visibility check on non-MTP bundles.
- Engine 55 tok/s MTP gate (verify-prefetch overlap / small-S fused verify) is a separate lane; Flash-Next Auto stays honestly off until measured faster.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind": "remoteSourceControl",
       "location": "https://github.com/osaurus-ai/vmlx-swift",
       "state": {
-        "revision": "447a6a2b9a458d1e077688e6af9dc81adff941c0"
+        "revision": "f9accec08944cdfa773f04a6d7c4e63242be51da"
       }
     },
     {
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,400 +1,409 @@
 {
-  "originHash": "a70de644fa9a976403022014284317c01c48d4d19f71b277a396f9c2786a3827",
-  "pins": [
+  "originHash" : "6d6f8666cc9b5b0231df72c69b0454e28abeb76d47e43c8a048f2a67f46165ab",
+  "pins" : [
     {
-      "identity": "aachartkit-swift",
-      "kind": "remoteSourceControl",
-      "location": "https://github.com/AAChartModel/AAChartKit-Swift.git",
-      "state": {
-        "revision": "b51e53c3b04372c1ed1906b0ec3ba45733953ef4",
-        "version": "9.5.0"
+      "identity" : "aachartkit-swift",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/AAChartModel/AAChartKit-Swift.git",
+      "state" : {
+        "revision" : "b51e53c3b04372c1ed1906b0ec3ba45733953ef4",
+        "version" : "9.5.0"
       }
     },
     {
-      "identity": "aptabase-swift",
-      "kind": "remoteSourceControl",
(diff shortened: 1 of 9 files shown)


jjang-ai:
Additional live proof on this exact build while CI runs:

- **Auto path verified end-to-end on JANG_2L** (its measured tuning artifact says d1): Auto genuinely launches speculative decode, and the output was byte-identical to a manual-depth leg run on a different quant in a separate process — only possible under the enforced greedy sampler, so the greedy-only-while-MTP policy holds on the Auto path as well as manual.
- **Fail-closed negatives re-confirmed live**: Ornith 1.5 9B (zero mtp tensors in all three quants) refuses a manual depth request with the typed complete-tensor-evidence reason; Qwen3.8-27B-2D's honest 1.00x tuning stamp keeps Auto off while 4D/6D auto-resolve d2 from their measured stamps.
- **Known follow-up found while proving**: the app-side MTP request path runs well below the harness on the same bundle+depth (2L d1: 41.4 tok/s harness vs 16.3 in-app at streamInterval 1; raising the dev root's streamInterval to the production value of 8 recovered +30% to 21.2). The remaining gap is an app emit-path overhead around multi-token MTP bursts, tracked separately — it does not affect this PR's enforcement contract.

PR #2507, opened 2026-08-28
Swap-pressure warning banner + designer emulation controls

jjang-ai:
## What
When macOS starts swapping *because of* a loaded model, generation slows and nothing told the user why. This adds a swap-pressure banner in the exact style of the existing RAM tight-fit popover, plus the tooling the design/QA team needs to see and iterate it without actually thrashing a Mac.

## How it decides (v3)
- **Per-load attribution windows.** Every COLD load appends its own record with its own baseline; warm cache hits never touch the episode. A load's window closes at its **first emitted output** (first chunk/reasoning/tool event out of the generation mapper) — so swap growth during load, prefill, and TTFT is attributed to the model, and growth after decode starts is not. The banner attributes to the load whose own window saw the most growth, not merely the newest one.
- Severity keys on the episode's **monotonic peak growth over baseline** — a small-memory Mac already deep in swap is not blamed on the model; only episode-coincident growth triggers, and the copy says "coincided with", never blame.
- Orange (elevated): swapping started, generation may slow. Red (critical): heavy swapping, slowdown expected. Advisory only — never changes model/sampler/cache behavior.
- Enter-fast on peak / exit-slow on sustained current-growth calm, with peak re-baseline on recovery so a stale spike cannot instantly re-enter. Dismissal remembers its severity; the episode ends when nothing is resident and no load is in flight (a failure can never erase another in-flight load's baseline).
- VM counter regression (host_statistics64 reset across a sleep cycle) re-primes instead of wrapping into an astronomical swap-in rate.

## For the design team — how to see it
1. Launch a dev build with `OSAURUS_SWAP_EMULATE=elevated` (or `critical`) — the banner shows immediately, labeled **(simulated)**.
2. Or flip it live without relaunching: `echo critical > ~/.osaurus/debug/swap-emulate` (dev/test roots: `$OSAURUS_TEST_ROOT/debug/swap-emulate`); delete the file to end the simulation.
3. Real end-to-end pressure: `./scripts/swap-balloon.sh 30` genuinely pushes the machine into swap (Ctrl-C releases instantly); watch `sysctl vm.swapusage`.

The banner copy, icons (externaldrive variants), tints, and dismissal behavior are all up for design iteration — the plumbing is done and deterministic to preview.

## PROOF
- 11/11 monitor tests green, now with an **injectable swap sampler**: the new attribution test injects samples proving 2 GiB of growth before first output drives *elevated* and is pinned to the loading model, while +3 GiB after first output neither escalates severity nor re-attributes; the lifecycle test injects its baseline instead of asserting on host swap state.
- Lifecycle tests: per-load records, matched completion by model, in-flight-safe failure, unload teardown, flag-file emulation live-flip.
- Real balloon + model-load visual proof for this head to be posted before merge (prior head's leg: +4.8 GB real swap, red banner live in the chat card, 7× decode slowdown observed, recovery after release).

### DIFF
--- a/Packages/OsaurusCore/Services/ModelRuntime.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime.swift
@@ -2019,6 +2019,8 @@ public actor ModelRuntime {
         residentMetadata.removeValue(forKey: name)
         lastUseSource.removeValue(forKey: name)
         if currentModelName == name { currentModelName = nil }
+        // End of the residency episode: stop attributing swap growth to it.
+        if modelCache.isEmpty { SwapPressureMonitor.shared.clearBaseline() }
         if didRemove {
             if let retiredCacheCounters {
                 await MLXBatchAdapter.Registry.shared.recordRetiredCacheCounters(
@@ -3509,6 +3511,10 @@ public actor ModelRuntime {
         try Task.checkCancellation()
         let policy = await ServerConfigurationStore.load()?.modelEvictionPolicy ?? .strictSingleModel
         let loadStartedAt = CFAbsoluteTimeGetCurrent()
+        // Swap growth from here on is attributed to this residency episode —
+        // the swap-pressure banner distinguishes "this load is swapping" from
+        // a Mac that already sat deep in swap before the load.
+        SwapPressureMonitor.shared.markLoadBaseline()
         genLog.info(
             "loadContainer: begin model=\(name, privacy: .public) id=\(id, privacy: .public) policy=\(policy.rawValue, privacy: .public)"
         )
--- a/Packages/OsaurusCore/Services/ModelRuntime.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime.swift
@@ -1776,6 +1776,7 @@ public actor ModelRuntime {
             childOwnershipToken: loadingRecord.childOwnershipToken
         )
         loadingTasks.removeValue(forKey: name)
+        SwapPressureMonitor.shared.markLoadCompleted(model: name)
         currentModelName = name
         Memory.cacheLimit = mlxCacheLimit()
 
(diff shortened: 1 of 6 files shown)


jjang-ai:
Live render proof on the built Release app (background-driven, isolated test root): launched with `OSAURUS_SWAP_EMULATE=elevated` — the banner appears in the exact tight-fit popover style, pointer on the model chip, dismiss affordance in place, copy reads: "macOS started swapping while this model is loaded (2.0 GB). Generation may slow — closing other apps helps. (simulated)" — the simulated tag rendering as designed for honest screenshots.

Local validation: Release app build clean; all 8 classifier tests green (growth attribution, enter thresholds, 3-sample exit hysteresis, worsening re-entry, emulation parsing, flag-file lifecycle, live sysctl parse). Screenshot archived at vmlx-private-evidence/swap-warning-20260827/ for the design review.

jjang-ai:
v2 pushed (episode lifecycle, monotonic peak with recovery re-baseline, host swap-in/decompression telemetry, cautious coincidence wording, Open Activity Monitor / Unload Model / Keep Running actions, every-page balloon, 10/10 classifier+lifecycle tests) — and the REAL end-to-end row is done on a 128 GB M5 Max:

- Clean 4M-aligned load (73 GB resident), baseline 6,815 MB swap. 48 GB every-page balloon inflated against the resident model: swap grew to 11,645 MB (+4.8 GB, macOS expanded the swap pool) — banner went RED live with the real model name, Running phase, measured 4.8 GB, and all three actions, no simulated tag.
- Slowdown the banner predicts, measured: clean turn TTFT 0.75 s / 39.1 tok/s; under pressure TTFT 8.76 s / 5.3 tok/s (7x). sqlite even threw a disk I/O error during the thrash.
- Negative path: with the balloon inflated BEFORE the load, the existing RAM admission gate refused the 74 GB load outright ('Not enough free memory') — refusal, not thrash, and no stale baseline left behind.
- Recovery: balloon released, swap receded 11.6 -> 8.4 GB in 60 s; the banner correctly holds while current growth remains above the exit threshold (three-sample hysteresis, peak re-baselines on recovery).

Held open for design iteration on copy/tints/icons per the PR body.

PR #2508, opened 2026-08-28
evals(community): OsaurusAI/gemma-4-12B-it-MXFP8 on Apple M5 Max

jax-0n-git:
Crowdsourced model-compatibility run (see COMMUNITY_EVALS.md).

One contribution file; no other changes.

Environment: chip=Apple M5 Max · RAM 128GB · osVersion=26.6.2 · commit=f4b6ffa · judge=self-judge · catalogHash=2e766b2b52f4cf37 · contributor=jax-0n-git

### DIFF
--- a/reports/community/apple-m5-max-OsaurusAI-gemma-4-12B-it-MXFP8-20260827.json
+++ b/reports/community/apple-m5-max-OsaurusAI-gemma-4-12B-it-MXFP8-20260827.json
@@ -1,0 +1,162 @@
+{
+  "domains" : [
+    "agent_loop",
+    "apple_script",
+    "cache_proof",
+    "capability_claims",
+    "computer_use_loop",
+    "default_agent",
+    "http_api",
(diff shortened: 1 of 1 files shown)


jjang-ai:
Cross-family audit: this Gemma-4 12B report is not a normal 53/90 quality result. AgentLoop contains 140 errored rows and only 1 scored row; decode/TTFT are absent and the 57.7 MB peak means the model path was not resident, not that Gemma used 57.7 MB. The aggregate omits error taxonomy/transcripts, so no runtime/parser cause can be honestly assigned. It predates #2505 and #2528. PR #2546 adds exit/cache aggregation, but errored-row blocker/error taxonomy may also need preservation after the exact sidecar shape is inspected. Required follow-up: current-head rerun that first proves model load and then tool-result continuation, gen-config parity, multi-turn, cache, tok/s, TTFT, and real footprint.

PR #2509, opened 2026-08-28
Single, surfaced greedy-while-MTP coercion site (corrective for #2506)

jjang-ai:
## What
Review of #2506 found the greedy-while-MTP sampler coercion existed in TWO places: the long-standing site in MLXBatchAdapter and a redundant upstream copy #2506 added in ModelRuntime. Follow-up review then found the "single site" was still double-resolved: `effectiveGenerationSettings` already produced the coerced values that build `mlxParams` and the API readout, while a second `applyMTPGreedyIfNeeded` pass (and its log) ran dead code downstream.

This PR ends with exactly ONE resolution: `effectiveGenerationSettings` coerces greedy when native MTP is active and carries two flags — `mtpGreedyEnforced` and `samplerWasChanged` — that drive the run parameters, the API diagnostics, the log line, and the UI disclosure. Nothing can drift apart because nothing is computed twice.

## Changes
- Remove ModelRuntime.enforcingMTPGreedy and its call site; restore the four GenerationParameters sampler fields to lets.
- Delete the dead `applyMTPGreedyIfNeeded` helper; the surfaced log now keys off `effective.samplerWasChanged` from the single resolution.
- `last_effective_generation` gains machine-readable `mtp_greedy_enforced` and `sampler_was_changed`.
- The draft-strategy description becomes `native_mtp:dN·greedy-when-active` (Live Activity, cachedModelSummaries, load-plan log) — "when-active" because the marker describes the contract, not a per-request event.
- The Speculative Depth selector (Off/Auto/1/2/3) discloses the contract in a footnote: Auto activates only from measured tuning, 1–3 activate that depth directly, and while active the model decodes greedily (temperature 0). Localized for de/ko/ru/zh-Hans/zh-Hant.
- MTPGreedyCoercionTests rewritten against the REAL resolver, starting from Qwen 3.8 Flash Next bundle defaults (temp 1.0 / top-p 0.95 / top-k 20): native MTP resolves 0/1/0/0 with both flags true; AR and DFlash 2 preserve the bundle sampler; an already-greedy resolution reports enforced but not changed.

## Policy note (explicit)
Greedy-while-MTP itself is retained deliberately: it is the product owner's stated contract (MTP active ⇒ temp 0 for that model+session, everything else untouched) — native MTP's output-equivalence guarantee is only defined under greedy decoding. If the contract changes, the single resolution site is the one place to change it.

## PROOF
- MTPGreedyCoercionTests: 4/4 green through the real `effectiveGenerationSettings` pipeline (bundle defaults in, coerced/preserved values + flags out).
- i18n gate green locally (`scripts/i18n/check.sh`): catalog diff is additions-only (+34 lines, 1 key, 5 locales).
- Release app build green on this branch.
- Behavior unchanged from merged #2506: live legs on the isolated root showed manual-d2 activation fingerprints twice, Auto d1 on 2L with cross-model byte-identical greedy output, and mode-off sampling per bundle generation_config at 39.3 tok/s.
- Harness byte-exact frozen-prompt proof stands: 2L greedy MTP output identical to greedy AR at d1/d2/d3 (RunBench, fixed prompt, no clock, seed 42).
- Live app leg for THIS head (Off/Auto/1/2/3 + last_effective_generation showing mtp_greedy_enforced) to be posted before merge.


jjang-ai:
Review follow-ups pushed: the coercion is factored into a pure, unit-tested helper (`applyMTPGreedyIfNeeded`) — 3 tests pin the contract (MTP-active ⇒ effective 0/1/0/0; non-MTP preserves the bundle generation_config sampler incl. Qwen 3.8's 1.0/0.95/20; already-greedy reports no coercion, so the surfaced log fires only on real changes). The strategy marker is now `greedy-when-active` since the loaded strategy is load-state and tiny-prompt/cold-warmup requests can fall back to AR. The pre-activation explanation beside the Speculative Depth selector is queued with the design pass that owns that popover. In-process HTTP-harness proof of /v1/models + effective-settings readout is tracked as the qualification-lane follow-up (no shell curl).

jjang-ai:
Live proof for head 7036e6e78, dev-built app on the isolated test root, JANG_2L (Qwen 3.8 Flash Next), turns driven through the real chat UI:

**Off** — bundle generation_config governs. `last_effective_generation`: temperature 1.0, top_p 0.95, top_k 20 (exactly the bundle's shipped sampler), `draft_strategy: null`, `mtp_greedy_enforced: false`, `sampler_was_changed: false`. 245-token answer.

**Auto** — launches from the measured tuning stamp (best_depth=1). Runtime: `draft_strategy = native_mtp:d1·greedy-when-active`, `native_mtp_depth = 1`, reason "uses vmlx_mtp_tuning.json best_depth=1". Effective settings: 0/1/0/0 with `mtp_greedy_enforced: true`, `sampler_was_changed: true`.

**Manual depth 2** — `force_on` + `explicitDepth: 2`. Runtime: `native_mtp:d2·greedy-when-active`, depth 2, reason "User-enforced manual depth 2: tensor-evidence activation without measured tuning; greedy sampling enforced for this model+session." Effective settings: 0/1/0/0, both flags true. Picker screenshot shows the Speculative Depth row with "2" selected — matching the runtime telemetry — and the new disclosure footnote rendered beneath the segments.

Also caught by the engine's fail-closed validation during this proof: the 2L tuning stamp had been written with `quantization_bits: 2` (named after the bundle, not its actual 6-bit trunk) and Auto correctly refused to launch with reason "vmlx_mtp_tuning.json quantization_bits=2 does not match bundle quantization_bits=6" until the stamp was corrected. The gate works.

Issue #2510, opened 2026-08-28
Starved or killed generation leaves an eternal 'Thinking' turn with no bounded failure state


jjang-ai:
## What happened
While validating the swap-pressure warning under deliberately extreme memory pressure (incompressible balloons + a 65 GiB Flash Next 2L cold load on a 128 GB Mac), two chat turns ended in states the UI never resolves. **Both rows were correctly stamped `terminal_stop_reason = cancelled` in the chat DB** — the persistence layer records the outcome. The defect is presentation:

1. **Unbounded live spinner**: a starved generation shows "Thinking" with a live Stop pill indefinitely — no elapsed-time surfacing, no advisory, no bounded resolution until something external ends the run.
2. **Invisible cancelled rendering**: a cancelled turn with no visible answer renders as an empty assistant bubble (or bare thinking block) — indistinguishable from in-flight or quietly complete, despite the recorded `cancelled` marker.

## Why it matters
- The admission path already surfaces a bounded, visible error when a load is refused up front. An *admitted* run that starves has no equivalent bound.
- The recorded terminal state exists but is not rendered, so history and triage cannot tell interrupted from complete by looking at the transcript.

## Status
- Rendering half: PR #2516 renders the recorded cancelled state ("Interrupted — stopped before a response was produced.", localized) after whatever streamed.
- Spinner half: open. Deliberately NOT an auto-kill watchdog — cancelling work the user asked for on a timer would be an invented limit. Needs an advisory design (elapsed-time surfacing, or a "system under memory pressure" note tied to the swap monitor).

## Repro context
Isolated dev root, dev build; pressure via scripts/swap-balloon.sh (post-fix incompressible fill). Ordinary usage is unaffected — this needs genuine memory exhaustion, which is exactly the regime the swap warning ships for.

jjang-ai:
**Correction** (verified against the chat DB): the interrupted rows DO carry a terminal marker — both dangling assistant rows have `terminal_stop_reason = cancelled` (with `completed_at` NULL and empty content; one had 341 chars of thinking). The original claim that no failure marker is recorded was wrong.

The defect this issue tracks is therefore narrower and UI-side:
1. **Indefinite in-flight state**: a starved generation shows "Thinking" with a Stop pill forever — no watchdog, no bounded failure surfacing, no partial-output resolution, until something external kills the turn.
[...]

PR #2516, opened 2026-08-28
Render the recorded cancelled state instead of an invisible empty turn (#2510)

jjang-ai:
Fixes the rendering half of #2510: a completed assistant turn stamped `terminal_stop_reason=cancelled` with no visible answer produced **no block at all** (or only its thinking block), so the recorded interruption was invisible — indistinguishable from in-flight or quietly complete.

- Cancelled, non-streaming, no-visible-content turns now render *"Interrupted — stopped before a response was produced."* after whatever did stream (the thinking block stays).
- Streaming turns never show the line; unmarked empty turns keep the historical no-block behavior; billed empty turns keep the billing notice.
- 4 new ContentBlockDisplayTests; full suite 17/17.

The issue's other half — bounding the eternal in-flight "Thinking" state — is deliberately out of scope: an automatic kill would be an invented limit that cancels work the user asked for; an advisory treatment needs a design pass. #2510 stays open for that half.

Live screenshot of tonight's real cancelled rows rendering the notice to be posted before merge.


jjang-ai:
Live GUI proof (same dev Release build, local merge `171914c54` including this PR's head `37320b410`), both required cases driven through the real chat UI with the visible Stop control:

- **Cancelled before TTFT**: sent "Write a 300 word essay about deep sea vents.", pressed Stop before any output. DB row: `terminal_stop_reason=cancelled`, content 0, thinking 0. Transcript renders **"Interrupted — stopped before a response was produced."** where previously nothing rendered. Artifact `cancel-EARLY.png` sha256 `8aaa47dc44350723…` (shows two such rows, both visible).
- **Cancelled after reasoning, before visible answer**: sent "Compare tidal and wind energy…", let reasoning stream, pressed Stop. DB row: `cancelled`, content 0, thinking 93 chars. Transcript renders the Thought block (streamed reasoning preserved) followed by the Interrupted line. `cancel-THINK.png` sha256 `24270444feb98355…`.
- Notice string is localized (5 locales) as of `37320b410`; ContentBlockDisplayTests 17/17 at this head.

Note for the cancellation-consistency lane (not this PR): in the after-reasoning leg, the post-cancel turn's streamed reasoning referenced the *previous* cancelled prompt and decode ran at 1.3 tok/s — logged as evidence for the Stop/continue cache-consistency investigation.

jjang-ai:
Correction to the proof comment above: the "1.3 tok/s" side-note is RETRACTED as evidence — after the legs ran, ~68 GB of orphaned memory-balloon test processes were discovered still resident from earlier swap-pressure testing, which fully explains the slow decode. The cancellation-rendering proof itself is unaffected (block rendering from DB rows is memory-state-independent; both cancelled rows and their Interrupted rendering stand). The "post-cancel turn reasoned about the previous cancelled prompt" observation is downgraded to a lead requiring a clean-environment repro.

PR #2560, opened 2026-08-30
Remove hidden HTTP reasoning-budget coercion

jjang-ai:
## Problem

A finite OpenAI-compatible `max_tokens` silently armed `requestedReasoningBudgetTokens` for every HTTP request. The client did not request this. The runtime could mask/force reasoning-close behavior and it made native MTP ineligible because drafted tokens bypass per-token reasoning-budget processors.

Live Qwen3.8 Flash-Next evidence on the current Auto-D3 repin exposed the contradiction:

- host request: `draftStrategy=native_mtp`, fallback `none`
- authoritative completion: `decodePath=plain mtp=off`
- warm throughput: 32.9904 tok/s

This guard violates the bundle-driven generation contract and the no-hidden-coercion policy.

## Fix

Remove automatic HTTP reasoning-budget derivation and its coercion-specific tests. `max_tokens` remains only the output ceiling the client requested. No sampler, reasoning, prompt, or output behavior is invented.

An explicit reasoning-budget API can be added separately if clients need it; vMLX's explicit field remains available.

## Tests

Head below:

- focused source invariant: 1/1 PASS
- complete `RuntimePolicySourceTests`: 110/110 PASS

## Live gate

The exact Release served and visible-Chat after rows are running together with vMLX #368. This PR will not be merged until completion telemetry reports actual native MTP rather than the frozen plain-AR before result.


jjang-ai:
Current-head gate for `2683556fd54f55740238ede504b177c16c6c80a7`:

- Source trace: the HTTP finite `max_tokens` value is no longer reused as an implicit reasoning-token budget. Explicit reasoning budget remains the only budget source.
- Focused local evidence on the exact head: HTTP invariant 1/1 and RuntimePolicySourceTests 110/110.
- Integrated live evidence: the exact changed `MLXBatchAdapter.swift` bytes (`sha256 eb4ce...`) were used by the Release app proof that resolved Qwen3.8 Flash-Next to `native_mtp:d3`, completed the exact 1..200 sequence normally, and completed a real tool-call -> tool-result -> continued model turn in the visible chat UI. This demonstrates the request reaches native MTP without the removed hidden reasoning cap. The allocator-footprint policy is intentionally separate in #2561.
- Current CI rerun: 8/8 checks pass, including `test-core` (30m31s).

PR #2568, opened 2026-08-31
Preserve cache persistence after tool dispatch

jjang-ai:
## Summary
- stream the parsed tool envelope to the native UI immediately
- hold the executable `ServiceToolInvocation` until the upstream vMLX generation stream reaches terminal completion
- keep draining through the GPU synchronization and cache-persistence boundary before the next tool-loop inference can begin
- cancel upstream work only when the downstream consumer explicitly cancels
- add a regression test proving the envelope is visible before drain while tool execution waits for terminal completion

## Reproduced regression
A Raptor 8B-A1B JANG_6M agent loop initially completed small tool steps in about 1-2 seconds, then large 10-12 KB file/tool results grew later steps through 9.25s, 12.16s, 15.01s, and 72.25s. The public bridge delivered the executable tool error before vMLX completed the terminal GPU drain and `storeCacheAfterGeneration`, allowing the following inference to race the reusable checkpoint and restore an older boundary.

## Ordering after this change
`tool envelope parsed -> native UI receives tool metadata -> upstream terminal GPU/cache drain -> executable tool dispatch -> next inference`

This guard is independent of the cache topology. It prevents early dispatch for standard KV, native MTP, rotating/hybrid KV, SSM/linear-attention companion state, and disk/prefix tiers. Backend store refusal or companion-restore rejection still requires separate trace validation.

## Required proof before merge
- focused bridge regression test
- patched Release app build
- traced manual multi-turn tool loop showing an advancing cache boundary and bounded later-step prefill
- cache-store budget remains allowed under RAM-safety settings
- MTP enabled/disabled and alternate depth
- reasoning and non-reasoning
- VL/media tool path
- parser boundary plus prefix/suffix match behavior
- growing multi-tool/file payload regression added to OsaurusEvals after live proof

Release remains held until live proof.

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "1a2f4c17b45c57e86fa28c5a1ef2b41c4da94ffd"
+        "revision" : "a0c5b801050863c79ec6189bd6bf77de4c194bb3"
       }
     },
     {
--- a/Packages/OsaurusCore/Folder/FolderTools.swift
+++ b/Packages/OsaurusCore/Folder/FolderTools.swift
@@ -2697,8 +2697,10 @@ struct FileSearchTool: OsaurusTool {
 
     let name = "file_search"
     let description =
-        "Search files in the working directory. With `target=\"content\"` (default) it finds text by "
+        "Search plain-text files in the working directory. With `target=\"content\"` (default) it finds text by "
         + "case-insensitive substring match, returning matching lines with file paths and line numbers. "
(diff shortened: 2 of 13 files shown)


jjang-ai:
Live integration evidence with pinned vMLX a0c5b801050863c79ec6189bd6bf77de4c194bb3:

- Exact Release app built and launched from this PR.
- Eric manually confirmed the late multi-tool prefill meter now resumes from the expected recent boundary instead of restarting at an old low prefix.
- Real Raptor 8B A1B JANG_6M XLSX tool loop restored 29,450/30,254 tokens (804 new) and then 30,245/30,594 (349 new).
- The trace shows terminal generation drain/store completing before each subsequent Osaurus tool execution row.
- 7 tool executions, 9 terminal drains, and zero cache key-divergence events were observed.
- Store-budget telemetry was based on phys_footprint; tested rows had sufficient headroom. Low-headroom refusal/fallback remains unproven live.

Current merge gate: integration CI must finish. This evidence covers the growing tool-loop cache regression, not every broader model-family/release matrix item.

jjang-ai:
Stopped at request. This PR is intentionally closed and must not be merged or released.

PR #2572, opened 2026-08-31
Stop the memory-safety profile default from clamping the MLX allocator pool

jjang-ai:
## Problem

Safe Auto and Strict resolve their allocator cap to a fixed 128 MiB, and the session holder folded that profile default into every model load, capping `MLX.Memory.cacheLimit` at 128 MiB. Safe Auto is the shipping default, so effectively every user runs a 128 MiB freed-buffer pool.

Past ~13k tokens of context a single KV-sized decode intermediate no longer fits, buffer recycling silently stops, and every decoded token pays fresh zero-filled page allocation. On an idle machine the overhead is masked; under memory pressure (other resident models, per-tool-step KV cache stores) it collapses deep-context decode and delta-prefill for every family — the "every model becomes unusably slow after a few tool calls" report.

## Evidence — live A-B-A-B in one continuous chat session

Allocator flipped in-process with the value read back at every flip, 60 GB incompressible ballast holding pressure constant, interleaved small-context requests as a stability control (~127 tok/s throughout). Raptor-v0.5-8B-A1B, engine decode-only tok/s:

| Cap | Pressure | Deep-context decode |
|---|---|---|
| 128 MiB | none | 73-77 @ 8-13k, ~50 @ 18k |
| 128 MiB | 60 GB | 55 @ 8k → 37-40 @ 11.7k → 28 @ 12.3k |
| 8 GiB (flip) | 60 GB | 51.2 @ 13.2k |
| 128 MiB (flip back) | 60 GB | **18.9 @ 17.2k, 18.6 @ 19.2k** |
| 8 GiB (flip again) | 60 GB | **34.1 @ 19.7k** (first turn, one second after the flip) |

## Fix

`SessionHolder.allocatorCacheLimitBytes` records only a user-typed Memory Safety allocator override; with none set, the existing weight-scaled dynamic limit (`min(max(weights/4, 1 GiB), RAM/8, 8 GiB)`) governs. Explicit user overrides clamp exactly as before. `trimFreedBufferCacheUnderMemoryPressure` remains the idle backstop; the DSV4/native-MTP request-scoped ceiling is unchanged.

## Tests
Profile default reaches the holder as nil (no clamp); explicit user value clamps; `UInt64.max` saturates.

## Not in this PR
The hybrid compiled-decode gate change was measured live (comment below) and needs an overflow-bin sizing companion before it ships; it returns as its own PR.

### DIFF
--- a/Packages/OsaurusCore/Services/ModelRuntime.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime.swift
@@ -2466,6 +2466,25 @@ public actor ModelRuntime {
         return min(dynamicLimit, max(0, configuredLimit))
     }
 
+    /// The only allocator clamp a session holder may carry is one the user
+    /// explicitly typed into Memory Safety. The memory-safety *profile
+    /// defaults* (Safe Auto and Strict both resolve the allocator cap to a
+    /// fixed 128 MiB) must not override the weight-scaled `mlxCacheLimit()`
+    /// dynamic limit: past roughly 13k tokens of context a 128 MiB
+    /// freed-buffer pool cannot hold even one KV-sized decode intermediate,
+    /// so recycling silently stops and every decoded token pays fresh
+    /// zero-filled page allocation. Measured on Raptor 8B-A1B (same build,
+    /// same machine, single-variable A/B): decode 14.4 tok/s at 17.7k
+    /// context under the profile cap vs 47.7-52 tok/s at 14k with a working
+    /// pool; per-tool-step delta-prefill 1.3-3.0 s vs 0.3-0.5 s. The
+    /// dynamic limit stays bounded (weights/4, capped at RAM/8 and 8 GiB),
+    /// so the Qwen3.8 RSS-creep protection above is preserved.
+    nonisolated static func explicitAllocatorCacheLimitBytes(
+        customAllocatorCacheBytes: UInt64?
+    ) -> Int? {
+        customAllocatorCacheBytes.map { Int(min(UInt64(Int.max), $0)) }
+    }
+
     /// Decode-time ceiling. Native MTP / affine DSV4 may reuse large
     /// intermediates while a request is active, but the allocator pool must
     /// never be allowed to consume the entire load-admission budget. That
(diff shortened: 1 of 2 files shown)


jjang-ai:
test-core triage: the run had two components — GitHub-runner CoreData/AddressBook XPC noise (pre-existing, unrelated), and one real assertion: the MiniMax effective-settings test still expected the old app-side denial. Updated: the readout field now reflects the app-side request; the engine's typed gate refuses the divergent families (DSV4/Hy3/Laguna/MiniMax) at generation time. Note the field's semantics are now 'requested' — surfacing engine-actual engagement in the readout is a follow-up.

jjang-ai:
Scoping this PR down to the allocator fix only, based on the live A/B just completed on Qwen3.8-27B-JANG_4D (quiet machine, same allocator both legs, identical tool-chain prompts, engine decode-only tok/s):

| context | eager | compiled (current bin sizing) |
|---|---|---|
| ~2k | 31.3 | 19-24 |
| ~3-5k | 24.7-29.1 | 17.7-19.8 |
| ~9-10k | 17.7-20.4 | 17.5-20.4 |
| ~14.5k | 18.8 | 17.7-18.8 |

Compiled decode under the app's sizing (no compiledMaxCacheLength, chat maxTokens 16384 → overflow bin ~prompt+16k wide from the first token) is 25-40% slower at shallow context and only reaches parity at depth; Raptor 8B-A1B measured neutral at depth (38-43 vs 42-52) with fully coherent output. The engine's in-tree 16.6→24.6 tok/s win was measured with a small compiled bin, and the bin waste erases it at these settings.

The hybrid compiled-gate change therefore needs a bin-sizing companion before it ships; it will return as its own PR. What remains here is the memory-safety allocator fix, which is fully proven (A-B-A-B table in the description) and unaffected by this scoping.

jjang-ai:
## SSD prefix-cache store/reuse proof (same live sessions as the allocator A/B)

The disk (L2) tier — the only cross-request reuse tier with paged KV off — was exercised end-to-end throughout today's runs, engine trace lines verbatim:

**Reuse (restore + delta-only prefill), growing tool-loop conversation:**
```
[vmlx][cache/fetch] HIT disk boundary=14965 remaining=1827 ssm=36 tokens=16792   (Raptor: restore 14,965 of 16,792; prefill only the 1,827-token delta)
[vmlx][cache/fetch] HIT disk boundary=10499 remaining=4043 ssm=96 tokens=14542   (Qwen3.8-27B)
[vmlx][cache/fetch] HIT disk boundary=14541 remaining=114  ssm=96 tokens=14655   (114-token delta at 14.7k)
[vmlx][cache/fetch] HIT disk boundary=14654 remaining=1510 ssm=96 tokens=16164
```
Every tool step in every chain restored the prior boundary from SSD and prefilled only the delta — with this PR's allocator fix, restore + delta-prefill per ~134-token step measured **0.3-0.5 s** (was 1.3-3.0 s under the 128 MiB cap). Warm TTFT after model load: 0.79 s (Kanana 30B).

**Store (post-generation boundary persist to SSD):**
```
[vmlx][cache/store-phase] count=19654 eval=0.033s save=0.189s   (~1.2 GB record for a 19.7k boundary)
[vmlx][cache/store-phase] count=3083  eval=0.005s save=0.093s
```
Stores are synchronous but cheap on this hardware (0.1-0.3 s/record), and stored records round-trip: the restores above consumed records written by earlier steps of the same and prior sessions, across app restarts.

**Sustained speed over context with this fix** (Raptor, quiet machine, engine decode-only): 141 tok/s at ~0.5k → 91.8 @ 5k → 61.7 @ 9.2k → 47.7-52 @ 14k → ~50 sustained through 26k (user-driven session) — a smooth bandwidth-bound curve with no cliff, versus the 128 MiB profile cap's collapse to 10-18 tok/s under pressure documented in the description.

jjang-ai:
test-core triage for the final run: single recorded issue — a ChatSession.sessionId nil expectation in the chat-session persistence suite, untouched by this PR (which changes SessionHolder's allocator field, one helper, and its tests). The same workflow concludes failure on main itself: the two most recent main runs (11cd0c90e — current HEAD — and 91c418f65) are both red. Pre-existing, not introduced here. All other suites green.

PR #2588, opened 2026-09-01
Evals: cache write/reuse scored per tool call + sustained-decode collapse gate

jjang-ai:
Adds the eval lanes for the tool-loop, allocator-collapse, and compiled-decode regression classes — cache telemetry was scored per turn; agentic workloads regress per tool call.

## What's new

**Scripted tool loop** (`CacheProofEvaluator.toolResultFollowUps`): each payload is delivered as a tool-call continuation — assistant `tool_calls` + `role:"tool"` result, the exact request shape the agent loop produces between tool executions — with the probe tool's schema on every request so the chat template renders a consistent tool contract (what the assistant-continuation LCP boundary layer keys on). Model-initiated calls end a turn cleanly (`ServiceToolInvocations`/`ServiceToolInvocation` caught as expected terminals inside the scripted loop only). Deterministic: doesn't depend on the model choosing to call tools.

**New gates** on `cacheProof` expectations:
- `maxDiskL2StoresDelta` — SSD write-volume ceiling for low-RAM hosts; a rung-ladder regression (2× ~400MB exact+seed records per tool call, as measured live before vmlx#394) fails it.
- `minFinalDecodeTpsRatio` / `minFinalDecodeTps` — sustained-speed gate; pins the 2026-08-31 allocator-cap collapse shape (50 → ~10 tok/s as context grows) that every reuse floor passes. Per-turn decode tok/s now recorded in turn metrics.

## Cases (all PASSING live on Qwen3.8-27B-JANG_4D before this PR)

| case | gates | live result |
|------|-------|-------------|
| `tool-loop-per-call-restore` | disk-L2 hits ≥3 across 3 tool continuations, TTFT bounded per continuation | PASS — hits +3, all TTFT < 3s |
| `tool-loop-write-volume` | stores ≤12 across 4 requests + footprint growth cap | PASS — 9 stores, footprint bounded |
| `sustained-decode-deep-context` | final/first decode ratio ≥0.60 over 6 growing turns | PASS — ratio 1.00 (19.4 → 19.5 tok/s) |

Composes with `OSAURUS_EVALS_SIM_RAM_GB` for the memory-pressure variant of the sustained-decode case. Schema contracts pinned by `ToolLoopCacheProofCaseTests` (3/3). Off-CI cases (need a local model), same as the rest of the CacheProof suite.

### DIFF
--- a/Packages/OsaurusCore/Services/Context/CacheProofEvaluator.swift
+++ b/Packages/OsaurusCore/Services/Context/CacheProofEvaluator.swift
@@ -64,6 +64,10 @@ public struct CacheProofTurnMetrics: Sendable, Codable {
     /// Every typed progress frame in stream order. Optional so transcripts
     /// recorded before this field was introduced remain decodable.
     public let prefillProgressEvents: [CacheProofProgressEvent]?
+    /// This turn's decode rate from the stats sentinel. Optional so older
+    /// recorded transcripts remain decodable. Sustained-speed gates compare
+    /// the final turn's value against the first turn's.
+    public let decodeTokensPerSecond: Double?
 
     public init(
         turnNumber: Int,
@@ -78,7 +82,8 @@ public struct CacheProofTurnMetrics: Sendable, Codable {
         unclosedReasoning: Bool,
         visibleCharacterCount: Int,
         reasoningCharacterCount: Int,
-        prefillProgressEvents: [CacheProofProgressEvent]? = nil
+        prefillProgressEvents: [CacheProofProgressEvent]? = nil,
+        decodeTokensPerSecond: Double? = nil
     ) {
         self.turnNumber = turnNumber
         self.sessionNumber = sessionNumber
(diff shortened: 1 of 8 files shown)


jjang-ai:
Merged-proof record — all three cases executed live on Qwen3.8-27B-JANG_4D through the real ChatEngine streaming path before this merge (eval CLI, `OSU_MODELS_DIR=~/models`, reports in session artifacts): `tool-loop-per-call-restore` PASS (disk-L2 hits +3 across 3 scripted tool continuations, all TTFT < 3s, model-initiated eval_probe calls terminated turns cleanly), `tool-loop-write-volume` PASS (9 stores across 4 requests vs ceiling 12; footprint bounded), `sustained-decode-deep-context` PASS (6 growing turns, decode 19.4 → 19.5 tok/s, ratio 1.00 vs floor 0.60). Kit contract tests 3/3; catalog manifest regenerated via the manifest test's own canonical output. All required checks green.

PR #2586, opened 2026-09-01
Mark internal utility generations as auxiliary cache intent (repin vmlx 86d23df8)

jjang-ai:
Companion to vmlx-swift#391. Every `CoreModelService` one-shot is an internal utility (chat title, follow-up suggestions, memory distillation, transcript cleanup) whose prompt embeds per-turn content and is never resumed — yet each persisted its full prompt boundary. Measured live on Qwen3.8-27B: **six ~150MB hybrid KV records per chat turn** for title/suggestion prompts alone.

## Change

- Repin vmlx-swift 65ba6986 → 86d23df8 (adds `CachePromptIntent.auxiliary` + `LMInput.withCachePromptIntent`; nothing else changed upstream). Pin bumped in Package.swift, all three lockfiles, and both pin-consistency test expectations.
- `GenerationParameters.auxiliaryCacheIntent` (rides alongside `loadIntent` so it survives the ChatEngine → MLXService → ModelRuntime path).
- `CoreModelService` sets it on every one-shot — single choke point covering all four utility callers.
- `MLXBatchAdapter.prepareInput` applies `.auxiliary` to the prepared input; warmup prefill keeps its dedicated `.reusablePrefixWarmup` intent (explicitly excluded).

User chat requests are untouched — the flag defaults to false everywhere else.

## Proof (live, this exact tree, release build resolving 86d23df8)

- store-phase records for one chat turn: **12 → 6**; every 144–239-token title/suggestion boundary record eliminated.
- Auto-title and all follow-up suggestions still generated normally (suggestion step `promptTokens=194 genTokens=51` at 31 tok/s); main reply 29.6 tok/s.
- Remaining stores are exactly the useful set: warm-up 41/48, main-turn 3116/3117, post-gen 3007, warm-boundary 2969.
- Combined with vmlx#390 (already pinned), per-turn cache writes on a hybrid drop from ~5.9GB to ~1.6GB.

PR #2589, opened 2026-09-01
Repin vmlx-swift to de0c26d3 (one canonical boundary record per tool-call cycle)

jjang-ai:
Repins vmlx-swift 86d23df8 → de0c26d3, picking up vmlx-swift#394: the native-MTP store path gains the canonical-hybrid-boundary gate the solo/Batch/DFlash engines already had. Effect on hybrid tool loops (measured live): **one reusable boundary record per tool-call cycle instead of an exact+seed pair** (2 × ~100ms, ~400MB synchronous writes → 1), with every tool continuation still restoring at `promptMs=10`.

Together with the already-pinned vmlx#390/#391 and osaurus#2586, per-tool-cycle cache writes on a MambaCache hybrid are ~4× smaller than three days ago — the low-RAM SSD-oscillation shape this train targeted. The new osaurus#2588 `tool-loop-write-volume` eval lane gates against regression.

Pin bumped in Package.swift, all three lockfiles, and both pin-consistency test expectations.

## Verification (release build resolving exactly de0c26d3 per workspace-state.json)

Qwen3.8-27B chat turn: native MTP engaged (`STEP-MTP depth=3 active=2 avgCommitted=2.48`), 30.3 tok/s decode, and **exactly one** `store-phase` record for the turn (`count=1899`, the N-1 continuation boundary).

### DIFF
--- a/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/App/osaurus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -374,7 +374,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/osaurus-ai/vmlx-swift",
       "state" : {
-        "revision" : "86d23df8aaa91e568ca3f154db999f0768e81b08"
+        "revision" : "de0c26d3fe47ed556824914a5b84d28aaa938ae2"
       }
     },
     {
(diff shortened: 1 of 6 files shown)


jjang-ai:
test-core: another 180s time-limit flake (same runner-pool degradation documented across six runs on #2586 today — seven distinct environmental failures, zero repeats, zero in changed surface; this PR is a pin bump + comment + test-expectation hashes only). Rerunning.

PR #2592, opened 2026-09-01
Delegated runs resolve their own model's reasoning, not the parent's toggle

jjang-ai:
User-reported: spawn enforcement did not follow the reasoning configured for delegated models.

Root cause: `SubagentScope.enableThinking` carries the **parent** model's composer Thinking choice, and all five subagent kinds forwarded it verbatim into the delegated request — where an explicit `enable_thinking` overrides the delegated model's own persisted options in ChatEngine. A Thinking-off parent forced a Thinking-on child into no-think, and vice versa.

## Fix

`SubagentScope.enableThinking(forDelegatedModel:)` — the parent's choice survives only when the delegated model IS the parent's model (the documented reconstructed-step contract). Any other child gets `nil`, so ChatEngine's existing agent-options path (`agentModelOptionsProvider` → the child's own `ModelOptionsStore` entry → `AgentReasoningPolicy`) resolves the child's own configuration — the same resolution the visible picker shows for that model. All five kinds updated.

## Proof

- `SubagentScopeReasoningTests` 4/4: same-model inheritance (incl. case-insensitive), different-model independence, nil parent/child, unset choice.
- Live end-to-end (fixed build): Qwen3.8-27B parent → `spawn_batch` model-targeted `JANGQ-AI/Ling-3.0-tiny-JANG_6M` child → child resolved its **own** agent-context reasoning (3-token direct answer, no thinking; `STEP-STATS promptTokens=40 genTokens=3`) and reported the correct result (391) through the real spawn → model-resolution → ChatEngine chain.

Note from the same audit: the spawn allow-list + availability checks are coherent (both key on canonical picker ids); a hand-authored display-name entry fails availability with an error that echoes the invalid configured value as guidance — minor UX, not filed as a defect.

### DIFF
--- a/Packages/OsaurusCore/Subagent/Kinds/AppleScriptKind.swift
+++ b/Packages/OsaurusCore/Subagent/Kinds/AppleScriptKind.swift
@@ -300,7 +300,7 @@ final class AppleScriptKind: SubagentKind, @unchecked Sendable {
             dictionaryContext: knowledge.dictionary,
             recipeContext: knowledge.recipes,
             literals: literals,
-            enableThinking: scope.enableThinking
+            enableThinking: scope.enableThinking(forDelegatedModel: resolved.name)
         )
         return try Self.mapOutcome(result, model: resolved.name, mode: mode)
     }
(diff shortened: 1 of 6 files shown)


jjang-ai:
test-core: 180-second case time-limit again (the runner-pool degradation tallied across #2586's six runs). This PR's surface is SubagentScope + five call sites + its own new test suite. Rerunning.

PR #2646, opened 2026-09-05
Folder and knowledge listings: paged file_search/list_knowledge with totals, truncated-tree prompt line, ungrounded-claim advisory

jjang-ai:
## Problem

Discord reports on 0.24.7: a markdown notes repo attached as a folder chip or indexed as a project knowledge base is summarised as "50 files" (or 10 of 50) by Raptor 0.5, while Gemma 4 E4B sometimes reports it correctly. Live attribution on the 0.24.7 dev build with a 350-file fixture (custom agent, folder chip verified in the accessibility tree before every prompt):

| leg | tool path the model chose | answer |
|---|---|---|
| Raptor 0.5, cache on | `file_search(pattern:"*")` → 50 entries, `truncated:false`, no total | "50 files" |
| Gemma 4 E4B, cache on | `file_read(".")` (truncated listing) → `file_search("*.md")` → 50 | "**50** matching files" |
| Raptor, cache off | `shell_run` `find … \| wc -l` | "350 files" |
| Raptor, cache on, repeat | `file_read(".")` (`truncated` honest) | "146 files … the listing was truncated" |
| Raptor, temperature 0, cache on vs off | byte-identical trajectories: `file_search(pattern:"")` rejected as `invalid_args` twice → breaker notice, no answer | — |

So it is neither a model property (Gemma reports 50 too) nor the KV cache (temperature-0 trajectories are identical with the cache on and off). It is the tool flow:

- `file_search` pages at `max_results` (default 50) and reports `truncated:false` with no `total` / `next_offset`, so a model faithfully reports the page as the whole folder.
- `list_knowledge` has the same 50-row cap with no paging, and collection names in `collection` only match the exact display name.
- The prompt's folder tree stops at 300 files / 8 KB without saying so.
- After a failed listing, small models fabricate a summary instead of reporting the failure.
- Files-mode `file_search` with an empty pattern (how Raptor enumerates at temperature 0) was an `invalid_args` retry loop.

## Change

- `file_search`: `offset` argument; files mode reports `total`, `offset`, `next_offset` and a paging warning; content mode pages with `offset` and carries the same note; an empty/whitespace files-mode pattern means `*`; schema text tells the model how to enumerate a folder.
- `list_knowledge`: `limit` default 100 / max 500 (numeric strings coerced), `offset` paging with `total` / `next_offset` via `countDocuments`; collection matching accepts aliases, punctuation and case differences, and the sole granted collection.
- Folder prompt: `treeShownFiles` / `treeTotalFiles` / `treeTruncated` on `FolderContext`; the system prompt states "N of M files shown" and points at `file_search` paging (all three framings).
- `GroundedKnowledgeClaimCheck`: when `list_knowledge` failed and no knowledge read succeeded, a final answer that claims knowledge contents is regenerated once with a notice, then passes through honestly (bounded).
- Callers of the collection-resolution result updated for the new `note`; `.none` renamed `.unmatched`.

## Tests

`FolderListingPagingTests` (350-file fixture, deliberately not a multiple of the page size: tree line, disjoint pages covering all 350, string offsets, overrun, empty-pattern enumeration, content paging), `KnowledgeToolsTests`, `KnowledgeDatabaseTests`, `GroundedKnowledgeClaimCheckTests`, `KnowledgeClaimLoopDriverTests`, `AgentTaskStateTests`, `FolderContextRenderTests`, `UnifiedFileToolParamsTests`, `FolderSecretScopeTests`, `HarnessStabilityFixesTests` — 211 tests, all green locally (Xcode toolchain).

# PROOF:
- Live GUI runs (linker-signed dev build, harness FOLDER phase, driven by AX only): the table above — Gemma 4 E4B and Raptor both reported "50" from a paged `file_search` before the change; with paging totals the answer names the total and the truncation honestly; Raptor at temperature 0 gave byte-identical trajectories with the KV cache on and off, so the defect is the tool contract, not the cache.
- Unit: `FolderListingPagingTests` (350-file fixture, deliberately not a multiple of the page size — tree line, disjoint pages covering all 350, string offsets, overrun, empty-pattern enumeration, content paging), `KnowledgeToolsTests`, `KnowledgeDatabaseTests` (`listOffsetPastInt32IsClampedNotCrashing`: an offset past Int32.max no longer traps in the SQLite bind), `SandboxBridgePagingTests`.
- CI: all eight checks green at cd428d8b (test-core's first run failed with `Failed to create NSXPCConnection`, the same runner failure main's run 34008130163 at ae942a150 shows; the rerun passed).

### DIFF
--- a/Packages/OsaurusCore/Folder/FolderContext.swift
+++ b/Packages/OsaurusCore/Folder/FolderContext.swift
@@ -45,6 +45,19 @@ public struct FolderContext: Sendable {
     /// so the UI can hint at installable/relevant plugins for the folder.
     public let detectedFileExtensions: Set<String>
 
+    /// Files the rendered `tree` accounts for (listed by name or folded into
+    /// an extension-group summary) versus the visible files under the root.
+    /// The tree walk stops at `FileTreeOptions.maxFiles` (300), so a folder
+    /// of 350 notes renders 300 and a footer — and a model that counts or
+    /// summarises from the tree alone silently reports 300 (observed live,
+    /// 0.24.7: Raptor 0.5 with the Folder chip). `0`/`0` means unknown
+    /// (contexts built by hand); see `treeTruncated`.
+    public let treeShownFiles: Int
+    public let treeTotalFiles: Int
+
+    /// True when the tree omits files the folder actually holds.
+    public var treeTruncated: Bool { treeTotalFiles > treeShownFiles }
+
     public init(
         rootPath: URL,
         projectType: ProjectType,
@@ -53,7 +66,9 @@ public struct FolderContext: Sendable {
         gitStatus: String?,
         isGitRepo: Bool,
         contextFiles: String? = nil,
-        detectedFileExtensions: Set<String> = []
+        detectedFileExtensions: Set<String> = [],
+        treeShownFiles: Int = 0,
+        treeTotalFiles: Int = 0
     ) {
         self.rootPath = rootPath
         self.projectType = projectType
@@ -63,6 +78,8 @@ public struct FolderContext: Sendable {
         self.isGitRepo = isGitRepo
         self.contextFiles = contextFiles
         self.detectedFileExtensions = detectedFileExtensions
+        self.treeShownFiles = treeShownFiles
+        self.treeTotalFiles = treeTotalFiles
     }
 }
 
--- a/Packages/OsaurusCore/Folder/FolderContextService.swift
+++ b/Packages/OsaurusCore/Folder/FolderContextService.swift
@@ -57,13 +57,14 @@ public final class FolderContextService {
         // The tree walk, manifest/context-file reads and extension scan are all
         // synchronous filesystem I/O. Run them off the main actor so picking a
         // large folder can't hang the UI.
-        let (projectType, tree, manifest, isGitRepo, contextFiles, detectedExtensions) =
+        let (projectType, tree, manifest, isGitRepo, contextFiles, detectedExtensions, shownFiles, totalFiles) =
             await Task.detached(priority: .userInitiated) { [self] in
                 let projectType = detectProjectType(url)
                 let options = FileTreeOptions(
                     ignorePatterns: projectType.ignorePatterns
                 )
-                let tree = buildFileTree(url, options: options)
+                let built = buildFileTree(url, options: options)
+                let tree = built.tree
                 let manifest = readManifest(url, projectType: projectType)
                 let isGitRepo = checkIsGitRepo(url)
                 let contextFiles = readContextFiles(url)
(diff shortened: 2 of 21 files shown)


RaajeevChandran:
1. Files-mode file_search now walks the whole tree on every call. collectFileMatches used to stop after maxResults hits and now continues to the 20,000-entry visit budget so it can report an exact total. That's paid by every model on every files-mode search, including a targeted "find config.json". It's metadata-only and bounded so it's probably cheap, but can you post a quick timing on a large repo? Alternatively, only keep counting once the page is full, since a page that didn't fill already has an exact total.
  2. The generic alias list is fairly broad. docs, project, vault, and library are plausible real collection names. Exact match wins first so it's safe today, but the precedence is a hidden rule. A test that pins "exact name beats alias" would keep a future change from quietly breaking it.
  3. list_knowledge default limit doubled to 100. That's roughly 8 KB per call for any model that lists a big collection without a limit. Fine by me, just flagging it as a context-size change for all models rather than only the small ones.

jjang-ai:
Thanks — all three addressed.

**1. Walk cost.** Measured on this checkout (120,831 files outside .build/.git; walk is capped by the 20,000-entry visit budget), best of 3, files mode:

| call | before (main) | after (this PR) | match_count / total |
|---|---|---|---|
| `config.json` (targeted, page never fills) | full walk | 258 ms | 14 / 14 |
| `Package.swift` (targeted, 82 hits) | stops at 50 | 262 ms | 50 / 82 |
| `*.swift` | stops at 50 | 261 ms | 50 / 5333 |
| `*.swift`, max_results 500 | stops at 500 | 266 ms | 500 / 5333 |
| `*` | stops at 50 | 260 ms | 50 / 16520 (budget-truncated, flagged) |

The targeted case is unchanged: on main a page that never fills already walked to the end, so "find config.json" costs exactly what it did before. The added cost only exists when the page fills, and it is bounded by the visit budget: about 260 ms worst case on a 120k-file repo, metadata-only (`isRegularFile`/`isDirectory`, no reads). Counting once the page is full is what the code already does (entries stop being built; only `matched` increments) — the walk itself is the cost, and it is the price of an exact `total` the model can page against instead of reporting the page as the whole folder. I kept it.

**2. Alias precedence.** `exactNameWinsOverAlias` already pinned "Knowledge"/"Runbooks"; extended it to loop over the broad aliases you named (`Docs`, `Project`, `Vault`, `Library`): a granted collection with that exact name resolves to itself, and the same word with no such grant still resolves to "all granted". Reordering rules 1 and 2 now fails the suite.

**3. `list_knowledge` default 100.** Agreed it applies to every model. The old 50 with no paging is what produced the "10 of 50" summaries, and the schema now reports `total`/`next_offset` so a model that wants less can pass `limit`. Left at 100; happy to drop to 50-with-paging if you would rather keep the per-call size unchanged.

PR #2041, opened 2026-07-15
Fix Bonsai structured charts and hybrid memory safety

jjang-ai:
## Summary

Fixes the Bonsai 27B 1-bit JANG web-chart loop and the adjacent first-call/chart-memory failures reproduced in the real Osaurus UI.

The observed failures were not content-delta streaming failures. Content and reasoning deltas continued normally. The broken boundary was structured tool output:

- Bonsai nested a function-call `arguments` object one level too deep, sometimes as JSON text.
- The long UUID-like structured-data reference was mutated while emitting tool JSON.
- `web_search` exposed discovery snippets but did not provide a reliable transition to direct retrieval.
- The chart result re-prefilled hundreds of rendered points into the next model round.
- Osaurus discarded the resolved Safe Auto allocator cap after load and retained up to 50 materialized hybrid/SSM companion states.

## Changes

- Add a bounded discovery-to-retrieval transition and direct-URL support for `search_and_extract`.
- Preserve raw CSV/TSV/JSON and hand large structured payloads tool-to-tool through compact, exact, session-scoped `webdata:` references.
- Render referenced/nested JSON and narrowly recover clearly headerless numeric CSV/TSV.
- Safely unwrap a nested/stringified `arguments` object only when it is not a declared field and contains declared schema keys.
- Keep the full chart marker for the UI card while returning a compact structural result to the model.
- Honor the exact load-time allocator cache cap and bound live hybrid/SSM companion entries by the user-visible Memory Safety mode.
- Add focused and adjacent regression coverage for all of the above.

## Live app verification

Verified through Computer Use in an isolated, ad-hoc-signed Release build of the real app (`com.dinoki.osaurus.bonsai2041proof`), without replacing or reading the production app's bundle/preferences.

Model: `/Users/eric/models/dealign.ai/Bonsai-27b-1bit-JANG-CRACK`

Visible settings checked in the app:

- Tools, Memory, Charts, and Web Search enabled; Thinking toggled on and off.
- Prefix cache on; GPU paged KV and legacy disk cache off; SSM rederive on.
- Memory Safety = Safe Auto, resolved allocator cap 128 MiB, max concurrency 1, KV limit 65,536.

Observed rows:

- Direct CSV -> `search_and_extract` -> one `render_chart`: inline line chart rendered with no retry/loop, 506 -> 500 rows, TTFT 0.97 s, 44.0 tok/s.
[...]

PR #2647, opened 2026-09-05
Tool flow: no replay of transient knowledge replies, paging signals survive compression, manual mode holds across surfaces, capped results say so

jjang-ai:
## Problem

Follow-up to #2646 from the 2026-09-04 tool-flow audit (live repro on the 0.24.7 dev build, git archaeology per item). The common shape: the model's view of tools or results drifts from what the user configured or what the tool actually found, and nothing tells the model.

| # | Defect | Since | What a user saw |
|---|---|---|---|
| 1 | `search_knowledge` / `list_knowledge` "still indexing — retry in a moment" is a **success** envelope, and the read-like replay added by #2632 (0.24.7) serves an identical retry verbatim until a knowledge write. | 0.24.7 (the only defect created in the 0.24.4→0.24.7 window) | Retry never re-runs the tool; the finished index is never seen; "the collection is empty". Unit repro: the held result came back byte-identical. |
| 2 | `ContextBudgetManager` compresses a `kind:"search"` page to "[Compressed: 50 file match(es)]", dropping `truncated`, warnings, and #2646's `total`/`next_offset`; content searches lose their "(Results truncated at N.)" trailer. | #1313 (2026-05-31) | Long chats restate the page size as the whole folder even after #2646. |
| 3 | `growEnabledToolNames` appends every newly registered dynamic tool to **every** agent's manual list with no mode check. | #966 (2026-04-28) | Install a plugin or connect an MCP server → all its tools are silently enabled on agents the user set to "only what I ticked". |
| 4 | `duplicateRecord` drops `toolSelectionMode`, `manualToolNames`, `toolsEnabled`, `memoryEnabled`, `settings`. | fields added in #771/#1087/#1344, copy list never widened | Duplicate a manual agent → the copy runs AUTO with the whole registry and default grants. |
| 5 | The grant strips (`!isManual` block) are skipped in manual mode, but built-ins are not tickable in manual, so the grant toggles were the only off switch. | #1318 (2026-06-02) | Web Search OFF still exposes `web_search` / `search_and_extract` in a manual agent. |
| 6 | Settings → Orchestrator stores a manual tool list that `resolveTools` never reads. | #1271 (2026-05-28), born unwired | Ticking a plugin tool for the Orchestrator does nothing. |
| 7 | `capabilities` search prints "Found N capability(ies)" for a top-5/5/3 per query. | #1042 | "There are 5 email tools" when a provider exposes 12. |
| 8 | `search_and_extract` keeps the first 5 direct URLs with no warning. | #2041 | "Pages 6 and 7 returned nothing" for URLs never fetched. |

## Change

1. `KnowledgeToolScope.stillIndexingEnvelope` — a retryable `unavailable` failure (not a held error kind), used by both sites. A real success is still deduped.
2. Compressed search lines carry " — page of TOTAL total, next_offset N; not the full set" and "(truncated; …)"; content-search compression keeps the truncation / paging trailer.
3. `AgentManager.grownManualToolNames(current:mode:live:)` — pure, returns nil for `.manual`; used for the Orchestrator config and every agent.
4. `duplicateRecord` copies the capability configuration (plus `pluginInstructions`, `claudeCode`, avatar/voice).
5. Grant strips run in both modes; `additionalToolNames` ∪ ticked `manualToolNames` stay exempt as explicit picks.
6. Default-agent branch unions the ticked manual names into its allowlist when manual.
7. `CapabilitiesDiscoverTool.searchResultHeader` states the top-k and that it is not an inventory.
8. `SearchAndExtractTool.boundedDirectURLs` returns kept + dropped; the envelope carries `dropped_urls` and a warning.

Not in this PR (design call, noted in the audit): plugin `dispatch` requested tools and the plugin-host manual branch still ride the explicit-pick exemption.

## Tests

`ToolFlowTrueFixTests` (8: replay, compression, manual growth, duplicate, manual grant strip, Orchestrator manual picks, header, dropped URLs). Also green locally: `AgentTaskStateTests`, `HarnessStabilityFixesTests`, `KnowledgeToolsTests`, `SystemPromptComposerToolResolutionTests`, `Capabilities*Tests`, `WebSearchToolTests`, `SandboxDefaultEnablementTests`, `DefaultAgentConfigurationStoreTests`, `ContextBudgetPreviewTests`. Item 2's `total`/`next_offset` keys are produced by #2646; this PR reads them when present and is independent of merge order.
## Addendum

- Stacked on #2646 (base = `fix/knowledge-list-limit-and-grounding`); merge #2646 first, GitHub retargets this PR to main. No file overlap with #2630.
- Second commit: `shouldRecordAsLastEffectiveGeneration` also excludes auxiliary generations — live, the Live Activity readout showed the follow-up-suggestion sampler (`temp 0.4 · max 256`) for a chat turn that ran the bundle's 0.7 / 16384.
- A 'stop repeating this call' replay notice was tried and dropped: live at temperature 0 the model read it and kept issuing the identical call anyway, so it was an inert guard, not a fix.

## Live proof (dev build from this stack, Raptor v0.5 8B-A1B JANG_6M, 350-file folder, custom agent, folder chip verified in the accessibility tree before each prompt)

| run | sampling | tool calls | answer | loop? |
|---|---|---|---|---|
| 1 | bundle default 0.7 / 0.95 / 20 / rep 1.05 | file_read → file_search(files, 500) → 350/350 | "Exact file count: 350 files"; second turn listed the folder via shell | no (5 calls) |
| 2 | same | file_search(files) page 1 (50, total 350, next_offset 50) → file_search(files, 350) → 350/350 | "The folder contains 350 files in total"; second turn listed all 350 names (2,528 tokens) | no (2 calls) |
| 3 | user-forced temperature 0 (greedy) | file_search(max 100) ×15, identical, offset never added | none before the attempt cap | yes — model-level: at greedy Raptor drops the `offset` the result told it to pass and repeats deterministically; not reproduced at the default profile |

Settings wiring: `generation.maxTokens = 512` → readout `max 512`; cleared → `max 16384`. Tool-step readout = the bundle profile exactly (no hidden greedy, no dropped penalty).

### DIFF
--- a/Packages/OsaurusCore/Managers/AgentManager.swift
+++ b/Packages/OsaurusCore/Managers/AgentManager.swift
@@ -408,7 +408,22 @@ public final class AgentManager: ObservableObject {
             isBuiltIn: false,
             createdAt: now,
             updatedAt: now,
-            autonomousExec: agent.autonomousExec
+            autonomousExec: agent.autonomousExec,
+            claudeCode: agent.claudeCode,
+            pluginInstructions: agent.pluginInstructions,
+            // Capability configuration is part of what "duplicate" means: a
+            // MANUAL agent's copy used to come back AUTO with the whole
+            // registry, and every grant (memory, tools, knowledge, web…)
+            // reset to the defaults. The fields were added after the copy
+            // list was written and never joined it.
+            toolSelectionMode: agent.toolSelectionMode,
+            manualToolNames: agent.manualToolNames,
+            toolsEnabled: agent.toolsEnabled,
+            memoryEnabled: agent.memoryEnabled,
+            avatar: agent.avatar,
+            autoSpeak: agent.autoSpeak,
+            ttsVoice: agent.ttsVoice,
+            settings: agent.settings
         )
     }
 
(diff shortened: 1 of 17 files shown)


jjang-ai:
Added (from the build #13 matched runs, verified against the captured prompts): `osaurus_inspect` advertised the settings document sections in its scope enum and its validator rejection, then refused them at execution ("not an inspect scope") — a self-contradicting contract that produced a 29-call identical loop. A section name now reads the section; `osaurus_inspect`/`osaurus_help` join the deterministic-error dedupe. Engine replay of the exact looping prompt (fresh prefill vs warm restore, 5 seeds) was token-identical for 4/5 seeds, so the KV/recurrent cache is exonerated for that loop.

PR #2658, opened 2026-09-06
Reopened sessions: a window opened from a history row loads the stored transcript instead of erasing it on the next send

jjang-ai:
## What changed

"Open in New Window" on a sidebar/history row passed the row's metadata session (turns: []) to the new window. Three consequences, all observed live on 2026-09-06 with a 50-turn research session:

1. the new window rendered an empty transcript;
2. the next send reached the model with no history (prompt: system + the new message only), so the model truthfully answered that it had no record of the earlier work;
3. the incremental save deleted every stored turn the window never had — the session on disk went from 50 turns to 2.

`ChatWindowState.init` now hydrates a metadata-only session from disk the same way the in-place open already does (`ChatSessionStore.load(id:)`), falling back to the given data only for a brand-new session that is not on disk.

## Tests
`ReopenedSessionHydrationTests.windowCreatedFromAMetadataRowLoadsTheStoredTurns` — fails on main (`turns.count → 0`), passes with the change; window-state, folder-isolation, history-database and queued-send suites green (69/69).

## PROOF
Live before/after with the harness (reopen a stored session from History, send one follow-up, check the prompt's message count and the turn count on disk) follows in a comment once the current harness run finishes.

### DIFF
--- a/Packages/OsaurusCore/Managers/Chat/ChatWindowState.swift
+++ b/Packages/OsaurusCore/Managers/Chat/ChatWindowState.swift
@@ -256,7 +256,18 @@ final class ChatWindowState: ObservableObject {
         self.session.agentId = agentId
         self.session.applyInitialModelSelection()
         if let data = sessionData {
-            self.session.load(from: data)
+            // The sidebar and the history panel hold METADATA rows
+            // (`ChatSessionStore.loadAll` → `loadAllMetadata`, turns: []).
+            // "Open in New Window" handed that row straight here, so the
+            // new window showed an empty transcript, the next send went to
+            // the model with no history, and the incremental save then
+            // deleted every stored turn the window never had (2026-09-06:
+            // a 50-turn research session reopened from History became 2
+            // turns). Hydrate from disk exactly as the in-place open does
+            // (`loadSession(_:)` below); fall back to the given data only
+            // when the row is not on disk (a brand-new, unsaved session).
+            let resolved = data.turns.isEmpty ? (ChatSessionStore.load(id: data.id) ?? data) : data
+            self.session.load(from: resolved)
         }
         self.session.onSessionChanged = { [weak self] in
             self?.refreshSessionsDebounced()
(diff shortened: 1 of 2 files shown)


jjang-ai:
## PROOF (live visual harness, AX by pid; Ornith 1.5 9B JANG_4D, Web Researcher; engine 45bcb1de)

Recipe (harness phase RESUMER3): restore a stored 50-turn research session, open it from the History panel with "Open in New Window", send one follow-up, then read the follow-up's prompt dump (message count, tool responses) and the session's turn count on disk.

| Build | Head | Step | Prompt to the model | Turns on disk after | Model's answer |
|---|---|---|---|---|---|
| 0a29171c (before) | 87751954 | reopen + "Did you write the markdown file?" | 3 messages, 0 tool responses (session had 50 turns) | 2 | "I did not write either — I have no record of creating them." |
| 0a29171c (before, control) | 87751954 | reopen the same session again + "What was the first source you extracted…?" | 4 messages | 2 | "this is the first message, so there's no prior source extraction" |
| e3a72dae (after) | c2243cf77 | reopen the stored 50-turn session (runs 145738, 151428, 151946, 152411) | the reopened window's send-invariant warm-up request carried the transcript: 29 messages, 3 user turns, 23 tool responses | 50 of 50 original rows still present after the open (id, seq, role, content prefix identical) | follow-up NOT sent: the harness could not put text into this window's composer (AXValue set and pid-scoped keystrokes both left it empty; six attempts per run) — an automation failure, marked INVALID, not evidence about the app |

Files under the attached folder were unchanged in every step (sha256 before = after).

Unit: `ReopenedSessionHydrationTests.windowCreatedFromAMetadataRowLoadsTheStoredTurns` — reopen from a metadata row, append a follow-up pair, save, reload: on main the window had 0 turns and the save left only the pair; with the change the 4 stored turns keep their ids, order and content and the 2 new ones follow (6 rows). Window-state, folder-isolation, history-database and queued-send suites 69/69.

Not proven live: the reopen → send → save step itself (composer automation in a reopened long transcript). Proven live: hydration (the transcript reaches the model again) and no deletion on reopen; proven by unit test: the save after a follow-up keeps every original turn.

jjang-ai:
Evidence wording, stated plainly: live evidence covers reopen (transcript reaches the model again; 50 of 50 stored rows retained after the open). The reopen → follow-up → save step is covered by the regression test only; live follow-up submission in a reopened long transcript remains blocked by the harness's composer automation (four invalid runs), which is being handled separately.

PR #2660, opened 2026-09-07
Lazy chat model loading: selecting a model records the choice; the first Send loads and prefills the real request

jjang-ai:
## What changed

Proactive chat warm-up (#1897 and follow-ups) loaded the selected local model and ran a hidden one-token prefill on window focus, model pick, prompt-shape change, run completion, idle-unload recovery and freed-slot rewarm; DSV4 got an extra pre-send prefix request; and selecting a model evicted other models and preloaded the new one even with warm-up off (the disabled-warm-up fallback).

- `ChatWarmupController` is now a residency observer only. Every former scheduling entry point keeps its name (send handshake, window manager and Stop paths need no rewiring) and schedules nothing. Selection re-evaluates the residency dot — no eviction, no preload. The hidden generation producer, the DSV4 pre-send warm-up, the selection preload/eviction fallback and the post-run transcript rewarm are removed.
- The first Send loads the model and prefills the real request through the ordinary path (`ModelRuntime.loadContainer` on demand); the runtime's own residency policy decides eviction at load time. Native-MTP initialisation at load is untouched and therefore deferred to that first authorised load.
- "Automatically Warm Models on Load" is gone from Settings. The stored value still decodes; absent, true and false authorise nothing.
- Model chip: gray = not loaded, green = actually resident; no yellow state, no "warming up" tooltip. Send-time load/prefill progress stays in the normal progress UI.
- Kept intact: SSD/prefix caching on real requests, cancellation/drain, load admission, leases, history hydration (#2658), and the pure history renderers compaction reuses.

## Tests
`ChatWarmupControllerTests` rewritten for the lazy contract (no entry point schedules work for legacy warm-up values absent/true/false; selection records only and never evicts; focus, idle unload and freed slot refresh the dot and schedule nothing; revision gating and name matching; reset/Stop/shutdown leave nothing pending). `ChatSessionStopTests`: a Send after a selection dispatches synchronously with exactly one regular request and no warm-up request. Source assertions updated to the new contract. 406/407 locally across the loop, warm-up, stop, residency, source-policy, compaction, engine, sampler, hydration and queued-send suites; the one failure (`ContextBudgetPreviewTests`, section ids `enabledManifest`/`skillsGovern` present on this Mac) is being classified against an untouched main worktree before this PR is called green.

## PROOF
Live proof to follow in a comment: zero model work before the first Send through focus/history/picker/thinking/idle, then cold and warm sends, Stop and retry, and a relaunch on the same root with SSD telemetry, TTFT, prefill, decode rate and physical footprint — Ornith 9B (native-MTP negative) and a native-MTP-positive bundle; no DSV4 bundle is on this Mac (that row stays PARTIAL).

### DIFF
--- a/Packages/OsaurusCore/Models/Chat/ChatConfiguration.swift
+++ b/Packages/OsaurusCore/Models/Chat/ChatConfiguration.swift
@@ -117,10 +117,11 @@ public struct ChatConfiguration: Codable, Equatable, Sendable {
     /// When true, Osaurus will monitor the clipboard for new text content to offer as context.
     public var enableClipboardMonitoring: Bool
 
-    // MARK: - Model Warm-Up
-    /// When true, local chat sessions proactively load the selected model and
-    /// prefill the static prompt prefix (system + tools + history) before the
-    /// user sends, so the first response pays less time-to-first-token cost.
+    // MARK: - Model Warm-Up (retired)
+    /// No longer has any effect: chat model loading is lazy (selecting a
+    /// model records the choice; the first Send loads it and prefills the
+    /// real request). Retained only so existing configuration files keep
+    /// decoding; absent, `true` and `false` all authorize nothing.
     public var warmModelsOnLoad: Bool
 
     // MARK: - Auto-Generated Chat Titles
--- a/Packages/OsaurusCore/Resources/Localizable.xcstrings
+++ b/Packages/OsaurusCore/Resources/Localizable.xcstrings
@@ -1906,6 +1906,34 @@
         }
       }
     },
+    "Prefilling context %@ (%lld/%lld tokens)": {
+      "localizations": {
+        "de": {
+          "stringUnit": {
+            "state": "needs_review",
+            "value": "Kontext wird vorbereitet: %@ (%lld/%lld Token)"
+          }
+        },
+        "ko": {
+          "stringUnit": {
+            "state": "needs_review",
+            "value": "컨텍스트 프리필 %@ (%lld/%lld 토큰)"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "needs_review",
+            "value": "Предзаполнение контекста %@ (%lld/%lld токенов)"
+          }
+        },
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "needs_review",
+            "value": "正在预填充上下文 %@（%lld/%lld 个 token）"
+          }
+        }
+      }
+    },
     "Downloading, %lld%%": {
       "localizations": {
         "de": {
--- a/Packages/OsaurusCore/Services/Chat/ChatSessionWarmup.swift
+++ b/Packages/OsaurusCore/Services/Chat/ChatSessionWarmup.swift
@@ -2,155 +2,32 @@
 //  ChatSessionWarmup.swift
 //  osaurus
 //
-//  Warm-up payload assembly and model-switch helpers for `ChatSession`.
+//  `ChatSession`'s hooks into the residency observer, plus the pure renderers
+//  of the session's committed history. Chat model loading is lazy: nothing
+//  here loads, evicts or prefills a model — the first Send does that through
+//  the ordinary request path (`ModelRuntime.loadContainer` on demand, with
+//  the runtime's own residency policy at load time).
 //
 
 import Foundation
 
 extension ChatSession: ChatWarmupSessionContext {
-    func makeWarmupEngine() -> ChatEngineProtocol {
-        chatEngineFactory(source.inferenceSource)
+    /// Window focus / background work finished: refresh the residency-backed
+    /// chip dot from the runtime. No load is scheduled.
+    func notifySessionBecameActive() {
+        warmupController.handleSessionBecameActive(session: self)
     }
 
-    func makeWarmupPayload() async -> ChatWarmupPayload? {
-        guard let model = selectedModel, !model.isEmpty else { return nil }
-        guard selectedModelIsLocal, !isRemoteAgentTarget else { return nil }
-        guard !isImageGenerationModel(model) else { return nil }
-
-        let effectiveAgentId = agentId ?? Agent.defaultId
-        let executionMode = await prepareChatExecutionMode(agentId: effectiveAgentId)
-
-        // The plugin catalog is part of the static system/tool prefix. Wait
-        // for its one-time launch snapshot before warming, otherwise the same
-        // installed plugins can render a different prefix across process
-        // restarts and bypass an otherwise valid disk-L2 cache entry.
-        await PluginManager.shared.ensurePromptCatalogReady()
-
-        let liveToolMode = AgentManager.shared.effectiveToolSelectionMode(for: effectiveAgentId)
-        let liveFingerprint = SessionToolState.fingerprint(
-            executionMode: executionMode,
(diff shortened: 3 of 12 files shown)


jjang-ai:
## PROOF — live visual harness, ad-hoc build from this branch

**Build:** `fix/lazy-chat-load` @ 24d294649 (engine pin 45bcb1de, adhoc linker-signed, binary sha256 `1a1afe16c8641743`). The two later commits on the branch (2b4f882fd) change only the evals test file. Harness: the running app driven by pid over AX (never foregrounded, never curled), app log scoped to the pid, every turn gated on the persisted DB row. Phase `LAZYLOAD`: UI readiness (no warm wait), then the former warm-up triggers — window hide/show ×3, History panel open/close, model picker open → other model → back, Thinking on/off, 30 s idle — with a zero-model-work assertion from the pid-scoped app log (`loadContainer: begin`, `BatchAdapter submit`, `warmupPrefill`, `generateEventStream: start`) and zero `[Load]` lines before the first Send; then cold Send, warm follow-up, Stop mid-generation + retry, relaunch on the same root, 20 s idle assertion, reopen the session and Send with SSD telemetry. TTFT is the persisted `turns.time_to_first_token` (stamped after the load; the load is reported separately as load→submit from oslog timestamps), prefill/decode from the runtime's `[perf] mlxStats` of the real request, footprint from the runtime's `cache/store-budget active=` (phys_footprint).

### No loading before Send
| row | launch: after hide/show, History, picker, Thinking, 30 s idle | relaunch: after 20 s idle |
|---|---|---|
| Ornith-1.5-9B-JANG_4D (native-MTP negative) | model-work=0, load-lines=0 → **OK** (3 runs: full sweep 194422 and 204538, two bisect runs) | **OK** (all runs) |
| Qwen3.8-Flash-Next-JANG_4M (native-MTP positive, 96 GB) | model-work=0, load-lines=0 → **OK** (runs 200326, 203055) | **OK** (both runs) |

Chip: gray with tooltip "Model not loaded — your next message loads it first" before the first Send; green after. Picker "other model" pick and back: no load, no eviction (chip stays gray, rss 1.13 GB).

### Cold / warm / Stop / relaunch (real requests, actual-request prefill)
Ornith 9B, run 194422 (full sweep):
| leg | load→submit | prompt | prefill | decode | TTFT (post-load) | cache |
|---|---|---|---|---|---|---|
| R1 cold | **4.3 s** (SSD) | 3277 | 315 ms (10.4k tok/s) | 198 tok @ 31.1 tok/s | 1.13 s | HIT disk (system prefix rows from the same root) |
| R2 warm | 0 (container cache hit) | 3625 | 298 ms (12.2k) | 147 @ 34.1 | 1.15 s | HIT disk boundary=3270 |
| R3 Stop | submit 4051 tok, cancelled — `terminal_stop_reason=cancelled`, Stop button gone, no perf line | | | | | |
| R4 retry | 0 | 4322 | 259 ms (16.6k) | "43" | 1.14 s | HIT disk |
| relaunch R5 (reopened session) | **0.9 s** (page cache) | 3974 | 453 ms (8.8k) | 56 @ 54.1 | 0.85 s | **HIT disk boundary=3130** — SSD rows survived the relaunch |
Real weight loads: exactly 1 per launch (cache-hit re-entries excluded). Footprint 3.3–3.4 GB.

Flash-Next JANG_4M, run 203055 (94 GB free at start):
| leg | load→submit | prompt | prefill | decode | TTFT (post-load) | cache |
|---|---|---|---|---|---|---|
| R1 cold | **20.9 s** (page cache) + native-MTP init 2.6 s | 3277 | 145 ms (22.5k) | 158 @ 15.8 | 1.87 s | HIT disk |
| R2 warm | 0 | 3585 | 250 ms (14.3k) | 233 @ 13.0 | 12.8 s (queued behind the post-R1 title/suggestion helpers) | HIT disk |
| R3 Stop | cancelled | | | | | |
| R4 retry | 0 | 4122 | 401 ms (10.3k) | "43" | 5.9 s | HIT disk boundary=3578 |
| relaunch R5 | **18.9 s** | 3774 | 374 ms (10.1k) | 97 @ 26.4 | 1.88 s | **HIT disk boundary=3130** |
Footprint 100.8–101.9 GB (96 GB bundle materialised bf16). Native-MTP initialisation is untouched and now runs on the Send path right after the load: `native MTP cold warmup: first request … uses AR before enabling native MTP` → two 48-token probes (max 2 = AR bootstrap, max 8 = verifier warmup, 2.6 s) and then the user request decodes with MTP (`[NativeMTP] depth=3 …`).

### Cold first-message latency (measured, not promised)
The first message of a session now pays the load that selection used to pay: Ornith 9B **+4.3 s** from SSD / +0.9 s from page cache; Flash-Next JANG_4M **+20.9–35.7 s** (page cache / cold SSD) plus 2.6 s native-MTP init. Send-time load/prefill shows in the normal progress UI (`WarmupProgressHub` phases unchanged). Warm follow-ups are unchanged: prefill 250–300 ms on 3.6k-token prompts with SSD cache hits in both families.

### Controls and things that did NOT go as planned
- Flash-Next run 200326 (immediately after the 10-minute Xcode build; swap 2.6 GB used, compressor active) decoded at 6.0 → 1.1 tok/s at the same 101 GB footprint, and the R4/R5 sends never dispatched behind ~30 s helper generations. Run 203055 on the same build with 94 GB free: 13–26 tok/s. Memory state is the demonstrated variable. No main-build control on Flash-Next was run today, so "no decode regression vs main on Flash-Next" is **unverified**. The PR does not change the load or decode path (`ModelLoadIntent` only changes eviction policy; the `[Load]` lines are identical) — it changes when the load starts.
- Ornith first-launch decode is bimodal (≈33 vs ≈54 tok/s per request) across the full-sweep runs; a bisect of the sweep (hide+idle / History+picker+Thinking / none, one variable per run) does not reproduce it deterministically and every relaunch leg decoded at 54 tok/s. Eric's own instance issued no requests in the window. **UNEXPLAINED, intermittent; not tied to this PR by any evidence.**
- One sweep-off run (203843) got a stray "etc" Send 5 s after launch that the harness never issued (keystrokes reached the hidden instance's composer) — that window is INVALID and the harness now fails closed on an unplanned Send (user-turn count at T0 vs assertion time).
- DSV4: no bundle on this Mac → **PARTIAL** (the removed DSV4 pre-send request is covered by the source assertions and unit tests only).

### CI
24d294649: test-core green after cataloguing the tooltip key (`Prefilling context %@ (%lld/%lld tokens)`, percent formatted with `.formatted(.percent)`); test-evals failed on `SelectedChatWarmupLifecycleTests` still driving the removed warm-up API → rewritten to the lazy contract (grey after eviction, no replacement load, nothing ever warming; 3/3 locally). Head 2b4f882fd.

### Unverified list
1. Decode rate vs main on Flash-Next (no main-build control today; memory-state confound demonstrated).
2. Ornith per-request bimodal decode (33 vs 54 tok/s) — intermittent, cause unknown.
3. DSV4 row (no bundle).
4. Reopened-window composer AX input (human check still pending from #2658).

Evidence: `~/vmlx-private-evidence/mtp-swift-2026-09-04/logs/LAZYLOAD_*` (.out, .run, .app.oslog, .launch1.app.oslog, .sqlite per run) and ISSUES.md entries 20:50 / 21:05.

PR #2667, opened 2026-09-07
Swap-pressure banner: Unload Model through the guarded unload lifecycle, refusal reported, emulated target and button accessibility fixed

jjang-ai:
## Problem (CODEX-MERGED-LAZYLOAD-UNLOAD-AUDIT-2026-09-07, item 2)
The swap-warning banner's **Unload Model** called `ModelRuntime.shared.unload(name:)` directly and discarded the Bool. That skipped `MLXService.unloadRuntimeModel` — the guarded path the cache inspector uses: `ChatWindowManager.prepareSessionsForExplicitModelUnload` (Stop lifecycle for every session on the model, so no successful-run follow-up reloads it) plus the 5 s lease-drain timeout — and a refused unload looked like success.

## Change
- Unload Model → `MLXService.shared.unloadRuntimeModel(named:)`; single-flight ("Unloading…", control disabled) while the runtime drains; a refusal shows the inspector's "Couldn't unload … because it is still in use" text in the banner.
- Emulated banners (`OSAURUS_SWAP_EMULATE`, labelled "(simulated)") name a placeholder model; the button targeted that placeholder and unloaded nothing (first live press). An emulated banner now targets the chat's selected model; a real episode keeps targeting the model it measured.
- Accessibility: the banner container's `accessibilityLabel` shadowed all three buttons (assistive tech heard the swap sentence for Unload Model / Keep Running / Activity Monitor). The container now keeps its children and each button exposes its own title.
- Source test in `RuntimePolicySourceTests` pins the guarded call, the refusal report, single-flight, the emulated target and the per-button labels.

## PROOF
Live proof in a comment (pid-driven AX harness, banner raised by the labelled emulation, everything else real): idle unload, unload during decode, unload during a cold load, no reload across hide/show/focus/idle, one coherent reload on the next Send — Ornith 9B and Qwen3.8-Flash-Next (96 GB), footprint and chip dot receipts.

### DIFF
--- a/Packages/OsaurusCore/Tests/Service/RuntimePolicySourceTests.swift
+++ b/Packages/OsaurusCore/Tests/Service/RuntimePolicySourceTests.swift
@@ -3914,4 +3914,44 @@ struct RuntimePolicySourceTests {
             "Blocked bundles must not advertise selectable Auto or explicit-depth chips."
         )
     }
+    @Test("Swap-pressure banner unloads through the guarded lifecycle and reports a refusal")
+    func swapBannerUnloadUsesGuardedLifecycle() throws {
+        let floatingInput = try Self.source("Views/Chat/FloatingInputCard.swift")
+        let bannerStart = try #require(floatingInput.range(of: "private func swapPressureBanner("))
+        let bannerEnd = try #require(
+            floatingInput.range(
+                of: "private func swapPrimaryButton(",
+                range: bannerStart.upperBound ..< floatingInput.endIndex
+            )
+        )
+        let banner = String(floatingInput[bannerStart.lowerBound ..< bannerEnd.lowerBound])
+
+        #expect(
+            banner.contains("await MLXService.shared.unloadRuntimeModel(named: target)"),
+            "The banner's Unload Model must take the same guarded path as the cache inspector: Stop-lifecycle preparation of every session on the model, then the timed runtime unload"
+        )
+        #expect(
+            !banner.contains("ModelRuntime.shared.unload(name:"),
+            "The banner must not call the runtime unload directly (that skipped session preparation and the lease-drain timeout, and dropped the result)"
+        )
+        #expect(
+            banner.contains("if !didUnload {") && banner.contains("swapUnloadFailure = String("),
+            "A refused unload (model still in use after the lease-drain timeout) must be reported in the banner instead of being discarded"
+        )
+        #expect(
+            banner.contains("guard let target, !swapUnloadInFlight else { return }")
+                && banner.contains(".disabled(swapUnloadInFlight)"),
+            "The Unload control must be single-flight while the runtime drains leases"
+        )
+        #expect(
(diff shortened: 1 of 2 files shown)


jjang-ai:
## PROOF — live UNLOAD harness phase

Harness: the running app driven by pid over AX, never activated; app log streamed by process name from before launch and filtered by pid; the swap banner is raised by `OSAURUS_SWAP_EMULATE=elevated` (the app labels it "(simulated)") — the resident model, the runtime unload, the footprint (`footprint -p`, phys_footprint), the chip dot (window capture + saturated-pixel classifier) and the transcript are real. Unload Model is pressed by its AX frame; disk cache and model files are untouched.

### Build 990024f5 = exactly this PR's sources (09d70c513 on 32a7fc57a), Ornith-1.5-9B-JANG_4D, run 030113, 0 invalid cells
| cell | receipt |
|---|---|
| a11y | AX tree: `AXButton\|\|Unload Model`, `AXButton\|\|Keep Running`, `AXButton\|\|Activity Monitor`; the banner group carries the swap sentence (before: all three buttons carried the sentence) |
| R1 Send | one real load, footprint 898 MB → 1936 MB (rss-side), dot **green** |
| U1 Unload while idle | footprint → **974 MB**, dot **gray** |
| hide/show ×2 + focus + 20 s idle | model-work delta **0** — no foreground/idle reload |
| R2 Send | real-loads delta **1**, coherent answer, footprint 3584 MB, dot green |
| U2 Unload during decode | run ended through the Stop lifecycle (the model had called the `todo` tool mid-count; Stop button gone), footprint → **1028 MB**, dot **gray** |
| R3 Send | reloads exactly once, answers "42" |

### Same code minus the accessibility container line (build 219e22be), earlier today
- Ornith 022322: U1 3260 → 984 MB, gray, no reload across hide/show/focus/idle, R2 reload once; U2 during decode → `assistant:cancelled`, 1023 MB, gray, R3 "42". 0 invalid cells.
- **Qwen3.8-Flash-Next-JANG_4M (96 GB) 022606, 0 invalid cells:** U1 **75 GB → 1.8 GB**, gray, no reload; R2 reload once (82 s gate incl. load), coherent; U2 during decode → cancelled, 94 GB → 2.1 GB, R3 "42"; **U3 Unload during the cold load** (+6 s: load task started, not loaded, no submit) → the load is abandoned, footprint 1.65 GB, app alive, the transcript keeps the user row without a cancelled marker (explicit-unload path), R4 reloads once → "72".

### What the first live press found (fixed in this PR)
Build 22526515 (guarded path only): pressing Unload Model did nothing — the emulated state's placeholder "Simulated Model" was the target. Keep Running on the same build dismissed the banner, so AXPress reached the buttons; the target was wrong.

### Not covered
- A refused unload (model still leased after the 5 s drain) was not produced live; its report path is source-pinned only.
- Real (non-emulated) swap pressure was not induced on this Mac.

PR #2663, opened 2026-09-07
Lazy chat loading follow-up: residency-only chip tooltip, stale warm-up comment, activation semantics

jjang-ai:
Follow-up to #2660 for the review items on head 24d294649 (CODEX-PR2660-REVIEW): no behaviour change to loading, generation or residency.

- Model chip tooltip states residency only ("Model loaded — ready to respond" / "Model not loaded — your next message loads it first"). The previous "Loading model… / Prefilling context N%" branch observed `WarmupProgressHub`, which only ever receives phases from the suppressed-progress (hidden warm-up) path that #2660 removed; ordinary Send-time load/prefill progress goes to `InferenceProgressManager` and is shown by the normal progress UI. The dead branch, its hub observation and the catalogued key it needed are removed.
- Drops the stranded "model chip dot reflects warm-up state (yellow/green)" comment in `FloatingInputCard`.
- `ChatWarmupController` header documents the one retained non-observational side effect: window activation reads `ModelRuntime.chatActivationResidencySnapshot`, which cancels a PENDING idle-eviction decision for the selected model (pre-existing behaviour; loads nothing; an eviction in flight is not reversed, no replacement load is scheduled).

## PROOF
Live proof in a comment: the repaired LAZYLOAD harness (pid-filtered app log from before launch, logger-health check, genuine model pick/back and Thinking assertions, effective-reasoning ground truth from `turns.thinking`, request-matched cache/perf/TTFT, tool-continuation cell, cold-load Stop + retry on a third launch, chip dot colour captures, fail-closed INVALID cells) on Ornith-1.5-9B-JANG_4D and Qwen3.8-Flash-Next-JANG_4M with the build made from these exact sources.

PR #2668, opened 2026-09-07
Chat: a run cancelled during its cold load must not roll back the retry that already owns the session

jjang-ai:
## Problem (CODEX-MERGED-LAZYLOAD-UNLOAD-AUDIT-2026-09-07, item 3 — the lost-rows retry)
A run stopped while its model was still loading observes `CancellationError` only when the abandoned load throws, seconds later. The run task's catch branch told a user Stop apart from a privacy-review cancel by the **session** flag `stopRequested`, which the next `send()` resets. A retry sent inside that window therefore made the stale task take the "cancelled before any delta → restore draft" path against the retry's transcript: `handleCancelledBeforeFirstDelta()` removed the retry's assistant placeholder and user turn and put the stopped prompt back in the composer, while the retry's stream kept generating into a detached turn. Engine answered, nothing persisted.

Reproduced deterministically with the focused `COLDSTOP` harness on Qwen3.8-Flash-Next (20 s cold load): retry gap 1 s → lost, 3 s → lost, 8 s / 20 s → persisted. This is the unexplained cell from #2663's proof.

## Change
`if stopRequested || !isRunActive(runId)` — only the run that still owns the session may roll the transcript back; a stale task keeps its hands off the newer run's turns. One line plus the comment.

## Tests
`ChatSessionStopTests.stop_duringColdLoad_thenRetry_lateCancellationKeepsRetryTurns`: engine whose first request blocks like a cold load and then throws `CancellationError` late, second request answers "30". Red on the old code with exactly the live symptoms (users == ["Count to 400"], answer "", terminal "cancelled", composer holding the retry text); green with the fix. ChatSessionStopTests + RuntimePolicySourceTests: 123/123.

## PROOF
Live proof in a comment: `COLDSTOP` gap 1 s and 3 s on Flash-Next with the build made from exactly these sources — retry persisted (user turn + `stop` answer rows), composer clean.

### DIFF
--- a/Packages/OsaurusCore/Tests/Chat/ChatSessionStopTests.swift
+++ b/Packages/OsaurusCore/Tests/Chat/ChatSessionStopTests.swift
@@ -148,6 +148,49 @@ struct ChatSessionStopTests {
         )
     }
 
+
+    /// A run stopped during its cold load observes cancellation only when the
+    /// abandoned load finally throws — seconds later. If a retry was sent in
+    /// the meantime it has reset `stopRequested`, and the stale task used to
+    /// take the "cancelled before any delta → restore draft" branch against
+    /// the RETRY's transcript: its user turn and assistant placeholder were
+    /// rolled back while its stream kept generating into a detached turn, so
+    /// the engine answered and nothing was persisted (Flash-Next cold-load
+    /// Stop + retry within ~10 s, 2026-09-06/07). The stale task must leave
+    /// the newer run's turns alone.
+    @Test
+    func stop_duringColdLoad_thenRetry_lateCancellationKeepsRetryTurns() async throws {
+        try await ChatHistoryTestStorage.run {
+            let session = ChatSession()
+            let engine = LateCancellingThenAnsweringEngine()
+            session.chatEngineFactory = { _ in engine }
+
+            session.send("Count to 400")
+            try await waitUntilAsync(timeout: Self.asyncTimeout) { await engine.firstRequestStarted }
+            session.stop()
+            #expect(session.isStreaming == false)
+
+            // Retry while the first task is still winding down (its "load"
+            // has not thrown yet).
+            session.send("What is 5 times 6?")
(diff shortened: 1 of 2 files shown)


jjang-ai:
## PROOF — focused `COLDSTOP` cells on the build made from exactly these sources

Build `fd7e884fe0025ca6` = 768626f3b on 32a7fc57a (receipt dirty = exactly `ChatView.swift` + `ChatSessionStopTests.swift`). Qwen3.8-Flash-Next-JANG_4M (≈20 s cold load), pid-driven AX harness, DB gate on the persisted turn.

| retry gap after Stop (+6 s into the load, load task started / not loaded / no submit) | before (build 219e22be) | this build |
|---|---|---|
| 1 s | **lost**: user-turns unchanged, engine answered 2 tokens, transcript ended on the cancelled marker | **persisted**: user turn + `assistant:stop` "30", gate OK |
| 3 s | **lost** (same shape) | **persisted**, "30" |
| 8 s / 20 s | persisted | (not rerun) |

Same engine-side picture in both builds (one real load, MTP probes, a 3428-token submit, 2 generated tokens) — the difference is only whether the retry's turns survive the stale task's late cancellation.

Observed, not changed by this PR: after a Stop that lands before the first delta, the stopped prompt is put back into the composer (the `savedDraftOnCancel` restore); the retry's own text was sent and is not in the composer.

Unit: `stop_duringColdLoad_thenRetry_lateCancellationKeepsRetryTurns` red on the old code with the live symptoms (retry user turn gone, empty `cancelled` answer, draft restored), green with the fix; `ChatSessionStopTests` + `RuntimePolicySourceTests` 123/123.

Artifacts: `~/vmlx-private-evidence/mtp-swift-2026-09-04/logs/COLDSTOP_Qwen3.8-Flash-Next-JANG_4M_03*` (this build) and `_025*`/`_024*` (before); ISSUES.md 2026-09-07 02:55 item 3.

Issue #2604, opened 2026-09-02
model files (*.safetensors, config.json) are altered and rewritten


auer-kk:
### Checks

- [x] I have searched existing issues and discussions
- [x] I can reproduce this with the latest `main` or release

### Describe the bug

When running a model like "Qwen3.5-4B-MLX-4bit" a tempfile (".model.safetensors.vmlx-heal-UID.tmp") is created. Then the model.safetensors file is altered and rewritten. The config.json is altered as well and some values get converted.
Examples:
```
"gradient_clipping": 10000000000.0 => "gradient_clipping": 10000000000,
 "rms_norm_eps" : 9.9999999999999995e-07, =>  "rms_norm_eps": 1e-06,

```
This might lead to problems with storage and compatibilty with other engines and verifying the local directory against the original hf-repo is no longer possible. If this behaviour is intentional it should be clearly stated somewehere. 

### Provider type

Not applicable

### Osaurus version / commit

0.24.3

### macOS version

15.7.9

### Apple Silicon chip

M2 16GB

### Logs

```shell

```

jjang-ai:
Fixed in vmlx-swift **#412** (merged): the safetensors storage healer is now **opt-in and off by default**, so Osaurus never rewrites your original `*.safetensors` in place. When a bundle's tensors are misaligned for mmap, the loader falls back to MLX's aligned-copy path (reads correctly, touches nothing on disk) instead of realigning the file. Healing only happens if you explicitly set `MLXPRESS_HEAL_SAFETENSORS=1` / `JANGPRESS_HEAL_SAFETENSORS=1`.

To be clear on scope: `config.json` was never rewritten by the healer — it only ever touched `.safetensors` shards — and the `.vmlx-heal-*.tmp` file you saw was the atomic-replace staging file for that in-place realignment, which no longer runs by default.

This ships to the app in the next vmlx repin; I'll close this out once that lands so it's actually in your build, not just in the dependency.

jjang-ai:
Update after investigating both halves of this report:

**1. `.safetensors` rewrite + `.vmlx-heal-UID.tmp` — this is ours, and it's fixed.**
[...]

PR #2670, opened 2026-09-08
Model load: never rewrite a bundle's config.json for the Gemma-4 audio stamp (#2633)

### LINKED ISSUE #2633: model file (config.json) is altered and rewritten
### Checks

### Describe the bug

As I stated in #2604 the config.json might get altered and invalidated by Osaurus. 

### Steps to reproduce

1. Download "OsaurusAI/gemma-4-E4B-it-4bit" and verify the local directory against the hf repo.
2. Load/use the model with Osaurus
3. config.json is altered and rewritten. The conversion of floats to ints makes the config unusable for other engines.

### Provider type

Local model (MLX)

### Model name

OsaurusAI/gemma-4-E4B-it-4bit (and others, only gemma-4.x?)

### Osaurus version / commit

0.24.6

### macOS version

15.7.9

### Apple Silicon chip

M2 16GB

### Xcode version

_No response_

### Logs

```shell

```

### Screenshots

_No response_

jjang-ai:
Fixes #2633 (and the config.json half of #2604 for Gemma-4 bundles).

## Writer
`ModelRuntime.patchGemma4JangAudioMultimodalIfNeeded` (added in #1506) parsed `config.json` with `JSONSerialization`, added `quantization.multimodal`, and re-serialized the whole file. Foundation prints an integral double without its `.0`, so `10000000000.0` became `10000000000`, `50.0` became `50`, keys were re-sorted, and the bundle no longer matched the Hub — the exact change the reporter saw. The healer tests from #412/#416 were right about the healer; this function was the writer they did not cover.

Its "JANG/affine" gate also matched the ordinary mlx_lm `quantization.mode == "affine"` stamp. `OsaurusAI/gemma-4-E4B-it-4bit` from the report satisfies all four conditions (`model_type: gemma4`, `mode: affine`, no `multimodal`, `embed_audio.embedding_projection` in the index), so every plain 4-bit Gemma-4 audio conversion was rewritten on first load.

## Scope
This preflight no longer rewrites config.json. That prevents future rewrites; a config.json an earlier build already rewrote is not repaired (re-download it from the source). As of this change no Swift code reads the key: osaurus does not (main, and the June commit that added the patch), and vmlx-swift's quantization override decoder (`BaseConfiguration`) skips a string value as scalar metadata at both the June engine pin and the current pin. The jang converter stamps the flag at conversion time. Load or audio equivalence with the old behaviour is not claimed here (see Not verified).

## Change
- The preflight becomes read-only: `noteGemma4AudioBundleMissingMultimodalStamp(at:name:)` keeps the same gate, logs a missing stamp, returns whether one is missing, and never writes.
- Regression `Tests/Model/Gemma4AudioConfigRewriteTests.swift`: reporter-shaped bundle (mlx_lm affine + audio index), JANG-stamped variant, stamped and vision-only controls; asserts `config.json` bytes are identical after the preflight.

## Evidence (supervised local runs, `swift test --filter Gemma4AudioConfigRewriteTests`)
- Before (main 7e109ade6 + the test): 2/3 fail — reporter config 270 B → 371 B, `10000000000.0` and `50.0` lost, `multimodal` injected; JANG variant rewritten; controls untouched.
- After: 3/3 pass — bytes identical in both unstamped shapes; detector true for them, false for stamped and vision-only.

## Not verified
- A live load of a Gemma-4 audio bundle through the app (no gemma-4 bundle is present here and model loads are not part of this lane). The preflight call site is unchanged apart from the function name.
- Audio and load behaviour on JANG Gemma-4 audio bundles that lack the converter stamp, with versus without the stamp: not re-measured.
- The Python engine carries an equivalent in-place patch (`vmlx_engine/utils/jang_loader.py`); tracked separately.

### DIFF
--- a/Packages/OsaurusCore/Services/ModelRuntime.swift
+++ b/Packages/OsaurusCore/Services/ModelRuntime.swift
@@ -4143,14 +4143,10 @@ public actor ModelRuntime {
         // the whole process — taking osaurus with it. Caught here so the user
         // gets a clear error and the server stays up.
         try await Self.ensureJANGTQSidecar(at: localURL, modelId: id, name: name)
-        // One-time, idempotent: Gemma-4 JANG (affine) audio bundles shipped
-        // without the `quantization.multimodal` fp16-passthrough flag that the
-        // mxfp4/mxfp8 QAT bundles carry. Without it the audio embeddings fuse at
-        // the affine-quantized precision and the model degenerates on audio
-        // input (a spurious `thought` channel stub / dropped answer). Adding the
-        // flag — exactly the value the mxfp bundles use — restores clean audio
-        // and leaves vision unaffected (verified live on 12B JANG_4M).
-        Self.patchGemma4JangAudioMultimodalIfNeeded(at: localURL, name: name)
+        // Read-only: report a Gemma-4 audio bundle that lacks the
+        // `quantization.multimodal` stamp. This preflight no longer rewrites
+        // config.json (osaurus#2633); see the function.
+        Self.noteGemma4AudioBundleMissingMultimodalStamp(at: localURL, name: name)
         // Compute the incoming bundle's on-disk weight size for *every* policy.
         // Previously only `manualMultiModel` did this (strict left it 0), which
         // meant the resident-RAM accounting — and the pre-load feasibility gate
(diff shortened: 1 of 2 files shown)


jjang-ai:
## PROOF — exact-head integrated app check

Head `c8f0709a1256685b43dc08df39dc79479f08914f`, clean proof checkout; all eight exact-head checks SUCCESS, re-read before merge. Engine resolved to `45bcb1de740d979a2322570a06edd457ce3d0697`. Post-sign development binary SHA256 `d27eb1d71ecaa17913e56a95a96c980555bca90ef130417f844d4005efe2e1d0`.

This extends the earlier 3/3 config-fixture regression with the previously missing live text-load proof. Isolated optimized Release app (incremental compilation, no batch mode, two Xcode tasks), unique app identity/private history; installed app and original bundle untouched. It is not a shipping-build speed comparison.

Real Gemma E2B QAT metadata clone, `quantization.multimodal` removed only from the clone, original weights symlinked. First Send reached `noteGemma4AudioBundleMissingMultimodalStamp`: app.oslog line14267 at19:47:34 explicitly reports config left untouched. Clone config SHA256 remained `d172b9623bbe6cfd06d7d5a747ab3d9a3314d1f8f1fbfc76189b97988b2a4f30` before/after two turns and unload. Original config remained `173c3e49dfffb2b7c83de803c3bf46d506706935346d392e208d716db3e75d15`.

Two real UI turns acknowledged and then correctly recalled Cedar/Thursday, each13tokens, natural completion. UI rates58.1tok/s then90.0tok/s; TTFT3.63s (+3.5sload) then0.30s. Thinking Off, bundle sampling unchanged. Gray/unloaded before Send. Additional lifecycle observation: a labeled simulated swap warning's real Unload button left resident_models empty, gray chip and about1.05GB tracked footprint. This was not induced swap, minimized-window proof, or all-lifecycle qualification.

Live supervised run exit0, peak3.43GB, swap flat1.17GB, sampled pressure normal, cleanup group/tracked/watchdog0/0/0. Build peak8.21GB. No build/model overlap.

Artifacts: `/Users/eric/vmlx-private-evidence/mtp-swift-2026-09-04/proofs/pr2670-live.03C6fO/RESULT.md`, `turn2.png`, `turn2.ax.txt`, `app.oslog`, `unload-after.health.json`; logs `SWIFTTEST_PR2670Live__194547.*` and `SWIFTTEST_PR2670AppIncremental2__192452.*`.

Scope remains prevention of future config rewrites. No repair of previously rewritten files, no audio-equivalence claim, no every-family/cache/tool qualification. The PR body's earlier statement that no live bundle was available is superseded only by this named text-load row; audio parity remains unverified.

PR #2828, opened 2026-09-19
Add Serply as a bundled search provider

googio:
## Summary

Adds **Serply** as a ninth bundled API search provider. Serply is a Google SERP
API whose free tier needs no credit card, which puts it next to Tavily as one of
the few keyed options a user can try without entering payment details.

It is pure data. The definition uses the existing declarative format, so there
are no executor changes and no privileged built-in path: web and news come off
the same `https://api.serply.io/v1/search` endpoint, with the vertical selected
by Google's own `tbm` parameter rather than a separate path.

Posting this after tpae welcomed it on Discord.

## Changes

- [x] Behavior change
- [ ] Refactor / chore
- [x] Tests
- [ ] Docs

The UI change is the one a new catalog entry always produces: another card in
Settings > Search > Search Providers, rendered by the existing preset view from
`summary`, `pricingNote`, `instructions` and `signupURL`. No view code changed.
See Screenshots for why there is no image.

## Test Plan

**Model used:** none. This is a declarative data change; no model was involved.

**What I could not run, stated up front.** I do not have a Mac, so
`xcodebuild test` and SwiftLint did not run here and I have left the "verified
build on macOS" checkbox unchecked. I am relying on CI for the macOS run. To
compensate I verified in layers.

*Layer 1: the definition is exercised by the suite's own fixtures.* Three tests
added in `DeclarativeSearchBackendTests.swift`, following the file's convention
of one build test plus one parse test per bundled definition:

- `serplyClampsNumAndSendsKeyInHeader` - GET, `X-Api-Key` header, `num` clamped
  to 10, 0-based `start`, `tbs=qdr:m`, no body.
- `serplyNewsSelectsTheVerticalWithTBM` - same URL plus `tbm=nws`, and `tbs`
  omitted when no recency filter was asked for.
- `serplyResponseReadsTheNestedSourceDomain` - the `metadata.display_url` dot
  path, and that `publishedDate` stays nil.

`serply` was also added to the documented-provider list in
`catalogProvidesTheDocumentedProviders`, so the existing catalog invariants
(must require a key, must support web, ids must be unique) now cover it.

*Layer 2: the request the definition builds, run against the live API.* I
transcribed `buildRequest` for this definition and issued the exact URLs it
produces:

```text
web            GET .../v1/search?q=swift%206%20concurrency&num=10&start=0
               -> HTTP 200, 10 hits
web + recency  GET .../v1/search?q=...&num=10&start=0&tbs=qdr:m
               -> HTTP 200, 10 hits
web page 2     GET .../v1/search?q=...&num=5&start=5
               -> HTTP 200, 5 hits, zero overlap with page 1
news           GET .../v1/search?q=...&num=10&start=0&tbm=nws
               -> HTTP 200, 10 hits

every hit populated title + url + snippet + sourceDomain in all four cases
num=25 requested -> the definition sends num=10 (clamp), API returns 9
wrong key        -> HTTP 401
```

Two things that run settled, and why each is in the definition rather than
assumed:

- **`start` is 0-based.** It is Google's own parameter, so unlike Google CSE's
  1-based `start` it maps to `{{offset}}`, not `{{start}}`. I checked: `start=1`
  returns a page shifted by one, silently dropping the top hit. `start=0` is
  identical to omitting it.
- **A page tops out at 10 hits** however high `num` goes, hence `clampMax: 10`.

**`publishedDate` is deliberately unmapped.** There is no dedicated date field.
`metadata.attributes` holds a date on web hits (`"2025/04/24"`) but a relative
age on news ones (`"3 weeks ago"`), so mapping it would be right about half the
time. Better nil than sometimes wrong. `sourceDomain` comes from
`metadata.display_url`, which is a real domain on web hits and the publication
name on news ones, matching what Serper's news mapping already puts there.

**Localization.** The two new `L()` strings are in
`Localizable.xcstrings` for all five locales, matching the house pattern for
every other provider string: zh-Hans and zh-Hant translated, de/ko/ru seeded
`needs_review` with the English value. The other three strings
(`"Free: 2,500 queries to start"`, `"Copy your API key from the dashboard."`,
`"Paste it below."`) reuse existing keys verbatim, so they need no new entries.
`"Free: 2,500 queries to start"` is Serply's actual free allowance, not a
copy/paste from Serper.

## Screenshots

None, and I would rather say so than hand you an image I made up. Building the
app needs macOS and Xcode, which I do not have. The entry renders through the
unchanged preset card view, so it looks like the Serper card with Serply's own
name, summary, pricing line and three connect steps.

## Size

6 files, +203/-7. Two thirds of that is not logic:

- `Localizable.xcstrings` is +68/-0, entirely the two new string blocks in five
  locales each. It is a large generated file; the change is purely additive and
  local, and `json.load` round-trips it.
- `DeclarativeSearchBackendTests.swift` is +61.

The provider itself is +65 in `SearchProviderCatalog.swift`, and the remaining
three files are one-line wirings: the `apiProviders` list, the
`settings.providers` keyword list (AGENTS.md asks every settings change to land
in that catalog), and the header comment that enumerates the bundled catalog.

Happy to drop the localization file into a follow-up if you would rather review
the provider on its own.

## Checklist

On docs: there is nothing provider-specific to update. The provider list lives
in the catalog itself, `docs/examples/osaurus-config.sample.{yaml,json}` uses
Tavily as its single worked example and is not an inventory, and I updated the
one place that does enumerate the bundled catalog in prose, the
`SearchProviderConfiguration.swift` header comment.

The last box is unchecked on purpose: see the Test Plan.

Disclosure: I work with Serply. Happy to adjust scope, naming, or drop this
entirely if it isn't a direction you want for the project.


tpae:
I didn't welcome this.

